PR #1, opened 2016-05-29
Provide a more flexible dag output.

johanneskoester:
This PR makes the dag subcommand more flexible. It allows to pipe out the dag, and supports multiple formats. So far, gml and dot are implemented, but anything supported by networkx can be added.


bgruening:
Nice addition!

johanneskoester:
Thanks!

PR #3, opened 2016-07-16
implement blacklists

daler:
Mostly working, but I can't quite get the nosetests to see the generator correctly. @johanneskoester would you take a look?

Testing quick start, in an interactive `bioconda-build` Docker container:
* clone and install this repo via `python setup.py develop`
* copy the example config over to the bioconda-recipes dir
* add some recipes that should be built and a subset of those to `r-blacklist`
* from bioconda-recipes, run
```bash
bioconda-utils build . example_config.yaml --verbose
```


johanneskoester:
I will have a look. I will also do a bit of refactoring on this branch today.

johanneskoester:
Ok, I have changed a few things (speedup and logic). Further, I have adapted our code to the latest conda-build. Sorry for hijacking your branch for this.

johanneskoester:
Can you test this on your side and let me know if it works? Btw., bioconda-utils should now be invoked from within the recipes directory.

daler:
running as-is, I get:
```
Error: Failed to render jinja template in /bioconda-recipes/recipes/boltons/meta.yaml:
'dict object' has no attribute 'GIT_DESCRIBE_TAG'
```
After temporarily removing that recipe and bumping the build number on another recipe to trigger a build, the recipe builds successfully but then at the end I get:
 
```
  File "/anaconda/lib/python3.5/site-packages/nose/suite.py", line 76, in run
    test(result)
TypeError: 'bool' object is not callable
```
After changing

```python
yield build(recipe, env, verbose, testonly, force)
```
to

```python
yield build, recipe, env, verbose, testonly, force
```
as per the nose docs that say test generators should yield a tuple where the first item is a callable, then I get:
```
  File "/anaconda/lib/python3.5/site-packages/nose/suite.py", line 76, in run
    test(result)
TypeError: 'tuple' object is not callable
```
and nothing gets built. This seems like LazySuite and nose.main.run are not behaving the way the nose docs say they should.  Is this working for you? I'm running my tests in the latest biocona/bioconda-builder container.

daler:
I added some commits to actually use the blacklist when building the dag; nose tests issue seems to be last remaining problem. The Blacklist class is currently unused, not sure how you wanted work with that so I left it as-is.

johanneskoester:
My fault with that blacklist class. I forgot to remove it. The different nose behavior is weird. I will test again on my machine, it might have changed depending on the version.

johanneskoester:
I have removed all the ugly nose hacking. Now it should work (according to my tests). Can you confirm?

daler:
Yep, this works for me. There are some small things I'd like to fix later, like be clear about where -- in build_recipes or get_dag -- the blacklist filter should be applied, whether the filter should be applied at the *package* level or the *recipe* level (i.e, filter all versions or just a specific version), and to be more verbose in some cases for testing. But I'll submit another PR later when I have time to tinker.

It seems to be working fine for our use-case of preventing *all* R packages from building upon bumping the env_matrix.yml.

johanneskoester:
Good, thanks!

johanneskoester:
The next thing I plan to do before I activate bioconda-utils for the recipes repo is to integrate container based linux builds. This will allow empty subdags to fail early.

PR #4, opened 2016-07-20
Allow building inside docker container via the flag --docker.

johanneskoester:
@bioconda/core, please check this out and post your opinion.
This PR integrates docker building into bioconda-utils itself. This has two advantages:

1. With our SUBDAG setup in travis, we have to launch the build process for each subdags. Previously, even empty subdags have triggered the docker image pull. Now, we can decide to pull the image only if the subdag is not empty. This should provide a nice speedup.
2. It becomes easier to manually build packages within the docker image. All you need to do is to add the flag --docker. Thanks to the way the image is build, it can share the `conda-bld` directory with the host, so that the package will be available on the host system.

# Important note:
I decided to switch to the docker image `continuumio/conda_builder_linux` from @msarahan. As far as I know, this image is used to build the official default channel. My hope is that this is the first step towards everbody using a common build environment. This also means that we don't have to maintain our own image anymore.

### DIFF
--- a/bioconda_utils/utils.py
+++ b/bioconda_utils/utils.py
@@ -5,6 +5,7 @@ import glob
 import subprocess as sp
 import argparse
 import sys
+import shutil
 from collections import defaultdict, Iterable
 from itertools import product, chain
 import networkx as nx
@@ -158,7 +159,7 @@ def get_dag(recipes, blacklist=None):
     return dag, name2recipe
 
 
-def get_recipes(repository, package="*"):
+def get_recipes(recipe_folder, package="*"):
     """
     Generator of recipes.
 
@@ -166,8 +167,8 @@ def get_recipes(repository, package="*"):
 
     Parameters
     ----------
-    repository : str
-        Top-level dir of the repository
+    recipe_folder : str
+        Top-level dir of the recipes
 
     package : str or iterable
         Pattern or patterns to restrict the results.
@@ -175,7 +176,7 @@ def get_recipes(repository, package="*"):
     if isinstance(package, str):
         package = [package]
     for p in package:
-        path = os.path.join(repository, p)
+        path = os.path.join(recipe_folder, p)
         yield from map(os.path.dirname,
                        glob.glob(os.path.join(path, "meta.yaml")))
         yield from map(os.path.dirname,
@@ -225,7 +226,13 @@ def filter_recipes(recipes, env_matrix):
         exit(1)
 
 
-def build(recipe, env, verbose=False, testonly=False, force=False):
+def build(recipe,
+          recipe_folder,
+          env,
+          verbose=False,
+          testonly=False,
+          force=False,
+          docker=None):
     """
     Build a single recipe for a single env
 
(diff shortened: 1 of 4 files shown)


msarahan:
If this works for you, great!  I think the longer-term build is going to be a gcc conda package (either current 5.x series or 6.x series).  I'll encourage people to run this in a docker image that supports their minimum GLIBC target.  For conda-forge, this will be CentOS 6.  I will plan on updating the docker image you reference here - it will no longer contain a custom gcc, but instead will contain miniconda, perhaps with a gcc package pre-installed.

bgruening:
I'm all in favour of this! @johanneskoester great work!

chapmanb:
Great work. I only wanted to plug keeping support for CentOS 5 but otherwise happy to have better coordination with conda-forge

bioconda_utils/utils.py
```diff
@@ -251,35 +258,66 @@ def build(recipe, env, verbose=False, testonly=False, force=False):
...
+        build_args += ["--no-anaconda-upload"]
+    if not force:
+        build_args += ["--skip-existing"]
+    if docker is not None:
+        conda_build_folder = os.path.join(os.path.dirname(os.path.dirname(shutil.which("conda"))), "conda-bld")
+        recipe = os.path.relpath(recipe, recipe_folder)
+        env = dict(env)
+        env["ABI"] = 4
+        container = docker.create_container(
+            image="continuumio/conda_builder_linux:latest",
```
bgruening: Should we pin here a version?

daler:
This is a great idea. This PR doesn't appear to index the build dirs correctly, though maybe later conda versions do that automatically? I created a new PR #5 (to avoid cluttering this one) that addresses the indexing issue.

kyleabeauchamp:
On a related note, do we have a good way to keep track of what version of the build environment was used during creation of a given uploaded package?

bioconda_utils/utils.py
kyleabeauchamp: Seems reasonable to pin, and also to somehow propagate that version to the package build metadata somehow so we can keep track.

johanneskoester: You are right. I will pin the current version.

johanneskoester:
@msarahan glad to hear we all want to use a packaged gcc. In fact, we do that already in bioconda. Hence, the gcc in this container normally won't be used by us. If you update the image in the described way (removing the system gcc and also any libs/headers that are already available as conda packages), that would be perfect for our use case. The decision of conda-forge against CentOS 5 is really a problem for us. I think it will likely mean that we can never depend on their channel without risking breakage for some of our users.

johanneskoester:
@kyleabeauchamp one could think about adding this to the build string. Maybe there is also the possibility for conda to do this automatically (@msarahan)?

kyleabeauchamp:
Regarding Centos 5: I for one am willing to sacrifice backwards compatibility in order to get modern GPU support.

bgruening:
For me it is also more important to be compatible with conda-forge than maintaining CentOS5 compatibility.

@kyleabeauchamp why do you think the build-env should be part of an package? The build-environment is just an artefact, admitted an important one, but should ideally not matter once we agreed on a standard. I don't think this should be part of the build-string. Instead, can we integrate this automatically in the info field of meta.yaml during the build?

johanneskoester:
Technically, it might be possible to use a CentOS 6 docker image only in cases where we need it (e.g. GPU support).

johanneskoester:
I would just like to mention that it can be really hard to convince a sysadmin of, say, a large cluster environment to update the OS, especially if your argument is: "I want to use Conda/Bioconda, which is actually a way to bypass your control over the packages that are installed".

bgruening:
@johanneskoester so true and I need to take this fight here as well, but this does not outweigh the great opportunity to join forces with conda-forge, imho. If we can convince conda-forge to stick to CentOS-5 even better, but they had some good arguments for CentOS-6 and there was a survey ...

bgruening:
thanks @daler 
Ready for merging?

daler:
I think so, but it's up to @johanneskoester. And next will be enabling it for bioconda-recipes proper. And *then* we can finally get to updating R/bioconductor recipes!

kyleabeauchamp:
Regarding tracking the build env, i just wanted some way to see how things
were built.  Preferably programmatically.  Not urgent though

johanneskoester:
Thanks for the useful comments! Regarding the other raised questions/plans/concerns. I guess it would be great if those interested in further pursuing them would create issues for further discussion.

PR #5, opened 2016-07-20
Docker with indexing

daler:
#4 didn't index the build dirs, so in a single `bioconda-utils` invocation, later recipes wouldn't see earlier-built recipes. I couldn't figure out a good way of sending the both the build command and the index command to the same container, so I split out the container-building and container-running into separate subfunctions.

This now indexes the build dir(s) correctly with or without `--docker` mode. Note that both linux-64 and osx-64 dirs are indexed even in the docker for consistency. Hopefully we can use the non-docker version for building on osx.

(also added some minor tweaks for log output)

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -11,6 +11,7 @@ import argh
 from argh import arg
 import networkx as nx
 from networkx.drawing.nx_pydot import write_dot
+from docker import Client as DockerClient
 
 from . import utils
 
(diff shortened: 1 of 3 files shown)


johanneskoester:
I tested it without indexing, and my impression was that this is done automatically in the current conda-build. Did you try without manual indexing?

daler:
You're right, it's done automatically now.

PR #8, opened 2016-08-02
Make bioconda-utils work with the Continuum Docker image

bgruening:


bgruening:
@johanneskoester I tried to create a new user but this does not work, because the Docker user (docker --user) does not exist and is not in sudoers, so are not allowed to create users.

johanneskoester:
I just can't believe that we are trying to solve a new problem here. There must be a simple solution for this in Docker.

bgruening:
There is, it's called user-namespaces but this needs to be configured at the daemon level, afaik. Nothing we want to deal with I guess and nothing we want to enforce to our users.

johanneskoester:
Can you remind me of the exact problem we want to solve here? Is it that we can't read/write the mounted volume inside the docker container, or does the problem occur when reading the new files outside of the docker container, after the build has finished?

bgruening:
The latter I think.
Next thing I wanted to try is to remove the `--user` flag and start the container as root. In the startup script create the user and change to this user before anything else. But I expect some problems and I have seen @daler trying something similar.
So getting this in and working on it as a separate PR would be nice.

johanneskoester:
So, if it is the latter, can't we just pass $USER as an environment variable to the docker container, and chown to that user **within the container** after the build process has finished?

bgruening:
You can not chown a file for a user that does not exist. So you at least need to create this user beforehand. Also it can happen that you have inconsistent permissions. I think we can try a few scenarios from here on.
@daler was already passing the $USER into the startup script.

johanneskoester:
If you just pass user id and group id (not the names) to chown, you can change it, even if they don't exist. In case of names, this won't work, because chown needs to map them to the ids. However, here, we can just use the ids.

bgruening:
Yes, we can probably do this - leaving the problem with inconsistent permissions. So I'm more in favour of creating this user, currently.
The question is more do you want us to try it before merging or can this wait until the weekend.

johanneskoester:
This can wait. Can you explain what kind of inconsistent permissions could occur?

PR #12, opened 2016-08-03
Add the ability to specify labels

johanneskoester:
This allows to have dev packages in a channel that are ignored by conda install/update (unless enforced).


kyleabeauchamp:
Seems reasonable to me as long as we trust that conda install / update will consistently ignore `dev` packages.

johanneskoester:
conda will only consider the default label (main) by default. Everything else has to be specified at the command line. Hence, yes, dev will be always ignored.

daler:
Closing for now; we've made a lot of progress since this was originally opened.

PR #14, opened 2016-09-04
Update README.md

kyleabeauchamp:
I discovered a couple of tricky bits:

1.  Need python 3 presumably (see stacktrace below)
2.  You have to pre-install conda-build, because otherwise pip will try to do it automatically and choke.

```
    yield from map(os.path.dirname,
             ^
SyntaxError: invalid syntax
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ Utilities for building and managing
 
 Installation:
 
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
Closing in favor of an issue

PR #18, opened 2016-09-14
use docker for building linux recipes

daler:
The new `docker_utils.py` module now supports building a package in an isolated container and copies over the built package to the host's conda-bld dir. The host's conda-bld dir is mounted in the container, but is only used as a local channel rather than the container's conda-bld dir (doing so caused all sorts of permission and dangling symlink issues). In the end it's not that complicated, but I only worked it out after trying and discarding a bunch of other options. The `docker_utils.py` module documentation has all the details.

This setup seems to work well, and remains quite flexible for the kinds of things we need to do like install custom versions of conda on top of the condaforge/linux-anvil image. Most of the complexity is in the build script, which is actually built fresh and mounted each time, so there's a lot of flexibility there. 

I've also started a test suite that includes a convenient way of building temporary sets of recipes for test-cases. I'll continue to work on testing (including connecting to travis-ci) but wanted to run this by you guys @johanneskoester and @bgruening now to get some feedback.


bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,325 @@
...
+                fout.write(open(pkg_resources.resource_filename(
+                    'bioconda_utils', 'bioconda_utils-requirements.txt')).read())
+
+        with open(os.path.join(build_dir, "Dockerfile"), 'w') as fout:
+            fout.write(dedent(
+                """\
+                FROM {self.image}
+                COPY requirements.txt /tmp/requirements.txt
+                RUN /opt/conda/bin/conda install --file /tmp/requirements.txt
+                """.format(**locals())))
```
johanneskoester: Are you building a new container in order to avoid redundancy? Otherwise, wouldn't it be possible to just prepend those commands to the actual conda build command?

bioconda_utils/utils.py
```diff
@@ -289,72 +295,41 @@ def build(recipe,
...
-        status = docker.wait(container=cid)
-        logger.debug('Docker status: %s', status)
-        if status != 0:
-            logger.error(docker.logs(container=cid, stdout=True, stderr=True).decode())
-            return False
-        logger.info('Successfully built %s', recipe)
-        return True
-    else:
-        try:
+    CONDA_BUILD_CMD = ['conda', 'build', '--quiet', '--override-channels'] + channel_args + build_args
```
johanneskoester: Superfluous line above? But the override-channels is needed to ensure that the docker container does not leak channels into our build environment. E.g., with condaforge container, we would always get the conda-forge channel even if it is not in our config file.

bioconda_utils/docker_utils.py
daler: Yep, could mount `requirements.txt` at container run time and the requirements installed before every conda-build. The reason for having a build step was twofold: first, the conda install would only have to happen once per PR rather than once per recipe (hopefully cutting down on log clutter and a minimal time savings); and second, I was guessing we'll want at least some minimal modifications to whatever image we're using and doing it this way means there is a clear place for making changes to the image that lives alongside the rest of the build logic.

bioconda_utils/utils.py
daler: Thanks, this was leftover from testing.

`--override-channels` and other args to conda-build are passed in at container run-time in the `build_recipe` method.

bioconda_utils/docker_utils.py
johanneskoester: Makes sense, thank you!

bioconda_utils/utils.py
johanneskoester: Ok!

johanneskoester:
Sorry for the most likely heavy conflicts with master. I had to change a lot while debugging bioconda-utils together with conda-build 2.0.

daler:
No worries, thanks for working all that out.

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,325 @@
...
+    res = os.path.dirname(
+        os.path.dirname(
+            sp.check_output(
+                ['conda', 'build', recipe, '--output'],
+                universal_newlines=True
+            ).splitlines()[0]
+        )
+    )
+    # TODO: when we start running conda-build 2.0+, will need a `conda purge`
+    # call to clean up temp recipes
```
bgruening: I guess we need this now :)

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,325 @@
...
+        usr = pwd.getpwuid(uid)
+        self.user_info = dict(
+            uid=uid,
+            gid=usr.pw_gid,
+            groupname=grp.getgrgid(usr.pw_gid).gr_name,
+            username=usr.pw_name)
+        self.container_recipe = container_recipe
+        self.container_staging = container_staging
+
+        # Don't import until here to allow module import if docker not
```
bgruening: `is not`

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,325 @@
...
+                """\
+                FROM {self.image}
+                COPY requirements.txt /tmp/requirements.txt
+                RUN /opt/conda/bin/conda install --file /tmp/requirements.txt
+                """.format(**locals())))
+
+        logger.debug('Dockerfile:\n' + open(fout.name).read())
+
+        # TODO: need a better way of detecting build failure
+        response =self.docker.build(path=build_dir, rm=True, tag=self.tag)
```
bgruening: I think we could use `quit=False` or `decode=True`.

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,325 @@
...
+        self.container_staging.
+
+        Expects `HOST_USER` and `CONDA_BUILD_ARGS` to be passed in as
+        environmental variables.
+        """
+        build_dir = tempfile.mkdtemp()
+        with open(os.path.join(build_dir, 'build_script.bash'), 'w') as fout:
+            fout.write(dedent(
+                """
+                #!/bin/bash
```
bgruening: Can we put this on top of this file to and make it more obvious and easy to change.

bioconda_utils/example_config.yaml
```diff
@@ -1,10 +1,12 @@
-env_matrix: ../scripts/env_matrix.yml
+env_matrix: example_env_matrix.yml
+requirements: requirements.txt
 blacklists:
-  - r-blacklist
+<<<<<<< HEAD
```
bgruening: Failed merge

bioconda_utils/utils.py
```diff
@@ -306,58 +318,26 @@ def build(recipe,
...
-        logger.info('Successfully built %s', recipe)
-        return True
-    else:
-        try:
-            p = sp.run(["conda", "build", "--override-channels", recipe] + build_args + channel_args,
+    CONDA_BUILD_CMD = ['conda', 'build']
+
+    try:
+        if docker_builder is not None:
+
```
bgruening: To many empty lines.

test_env_matrix.yaml
```diff
@@ -0,0 +1,10 @@
+CONDA_PY:
```
bgruening: This file is redundant here, isn't it?

daler:
@bgruening thanks so much for the comments. I've addressed them all; the largest one was the visibility of the build script. I've moved it to a module-level template with explanatory comments. I figured the dockerfile template should be there too. They can both be overridden, which helps with testing.

I'd still like to get more test coverage before merging this, esp with respect to gcc and perl recipes.

bgruening:
@daler great work! Thanks for addressing them! Is the plan now to include on gcc/boost package into `test/test_case.yaml` and one for perl?

daler:
Exactly. Though possibly as separate `test_{gcc,perl}_cases.yaml` files to keep things isolated.

daler:
@bgruening we now have support for mulled-build testing in `bioconda_utils.pkg_test.test_package()`. Next up is integrating into CLI.

bioconda_utils/pkg_test.py
```diff
@@ -0,0 +1,104 @@
+import subprocess as sp
+import tempfile
+import tarfile
+import os
+import shlex
+
+from conda_build.metadata import MetaData
+
+'mulled-build build-and-test "gffutils" --test "python -c \'import gffutils\''
```
bgruening: Is this a comment?

bioconda_utils/pkg_test.py
```diff
@@ -0,0 +1,104 @@
...
+def get_tests(path):
+    "Extract tests from a built package"
+    tmp = tempfile.mkdtemp()
+    t = tarfile.open(path)
+    t.extractall(tmp)
+    input_dir = os.path.join(tmp, 'info', 'recipe')
+
+    tests = []
+    recipe_meta = MetaData(input_dir)
+
```
bgruening: Let me know if you need anything from this here upstream.

bgruening:
This looks awesome! We will have such a nice build and test system soon!
Let me know if you need anything in addition!

bioconda_utils/pkg_test.py
daler: Leftover from testing; will delete in future commit

bioconda_utils/pkg_test.py
daler: I didn't want to mess around with building a fake `args` object to use the existing mulled-build test extraction, so just wrote one here that takes a package. While it would be nice to just import from mulled_build, it might take some refactoring on your end which I'm not sure is worth it.

bgruening: Mh, maybe I don't understand the problem. But you depend on galaxy-lib, so you can simply import it, no? `from galaxy.tools.deps.mulled import foo`

daler: It's just that to use `galaxy.tools.deps.mulled.mulled_build.get_tests` I'd need to provide an `args` object with `.recipes` attribute, that's all. 

Also since the package testing in this module is intended to work on the level of individual tarballs, I figured I'd give it the tarball directly rather than an unpacked recipe dir. Those two things made it easier to just write a new function.

Is there another `get_tests` I'm missing from elsewhere though?

bioconda_utils/pkg_test.py
```diff
@@ -34,10 +29,8 @@ def get_tests(path):
         elif tests_imports and ('perl' in requirements or 'perl-threaded' in requirements):
             tests.append(' && '.join('''perl -e "use %s;"''' % imp for imp in tests_imports))
 
-        tests = ' && '.join(tests)
-        tests = tests.replace('$R ', 'Rscript ')
-    else:
-        pass
+    tests = ' && '.join(tests)
```
daler: @bgruening, upstream could use this fix for cases where there are no tests

bioconda_utils/pkg_test.py
bgruening: Ok, got it. I might look at it if I can improve it on our side but nothing urgent! Thanks.

bioconda_utils/pkg_test.py
bgruening: https://github.com/galaxyproject/galaxy-lib/pull/30
But I think we should hard fail if no test is provided.

bgruening: Thanks!

daler: True -- I only noticed it when building throwaway test cases.

daler:
### Recap
@bgruening @johanneskoester a recap on the recent changes:

- using the conda-build 2.0 API as much as possible. This needed that temp_env context manager to be able to pass things around via os.environ. 
- the actual conda-build commands are kept as subprocess calls (rather than use api) in order to keep the local system commands and docker container commands as similar as possible to help with debugging
- no longer using docker-py and instead build the subprocess calls. Seems to help with debugging, and one less dependency
- using mulled-build for testing built packages in a busybox container; removed the other "prototype" version of testing I had earlier
- better logging, testing, and docs all around

### Current problems
 Currently issues I'm working on (please tackle independently if you get a chance):

- I'm not sure how handle `numpy x.x` in dependencies. One way would be to change build string to `{{ CONDA_NPY}}_{{PKG_BUILDNUM}}` and remove the `numpy x.x` dependency, but this requires changing some recipes (bx-python, qiime, possibly others).

To see what I'm talking about, try this:
```bash
bioconda-utils build recipes config.yml --packages bx-python
```

- filtering is actually kind of slow across 1500+ recipes. It doesn't seem to be doing anything sophisticated, but I'm not sure how to speed it up further, e.g., 

```bash
bioconda-utils build recipes config.yaml --loglevel=info
```

bgruening:
> - filtering is actually kind of slow across 1500+ recipes. It doesn't seem to be doing anything sophisticated, but I'm not sure how to speed it up further, e.g., 

Most of the time (at least on github) we do not need to filter the entire set of repos isn't it? We could use TRAVIS_COMMIT_RANGE to get a list of all changed meta.yaml files. This obviously only works for normal recipe updates where no blacklist is involved but in this cases it could speed the testing up. We could detect blacklist changes as well and in this case trigger a full filter_recipes scan.

daler:
Good idea. Might also be able to do something with timestamps too (modification time on disk vs package upload time on anaconda) for local testing.

bgruening:
> Currently issues I'm working on (please tackle independently if you get a chance):
I'm not sure how handle numpy x.x in dependencies. One way would be to change build string to {{ CONDA_NPY}}_{{PKG_BUILDNUM}} and remove the numpy x.x dependency, but this requires changing some recipes (bx-python, qiime, possibly others).

I think the {{ CONDA_NPY}}_{{PKG_BUILDNUM}} approach is more what we are doing with other recipes I think, whereas the `x.x` approach is more that conda-forge is doing. I know that @johanneskoester once had a patch to enable the `x.x` functionality for arbitrary packages not sure what the status is was this. For the time being and because I would like to get this in as soon as possible ... hackathon is coming soon ... I would vote for {{ CONDA_NPY}}_{{PKG_BUILDNUM}} and change by-python and co.

daler:
Here's an issue and some PRs on the topic. 

- https://github.com/conda/conda-build/issues/1142
- https://github.com/conda/conda-build/pull/728
- https://github.com/conda/conda-build/pull/966

From my reading it seems like a lean towards jinja rather than the `x.x` syntax. So maybe changing those recipes to use the jinja variables would be the best way forward.

bgruening:
Ok let's do this. I will try as soon as I'm back.

daler:
@bgruening re: numpy x.x this is working in ebe880. So really we're down to either speeding up the filtering or deciding if we care about the time.

bgruening:
Imho we can care about optimization later on. You spend so much time and we will get so much better testing that we should get this out asap :-) 
I vote for optimizing later.
On Oct 16, 2016 4:23 PM, Ryan Dale <notifications@github.com> wrote:@bgruening re: numpy x.x this is working in ebe880. So really we're down to either speeding up the filtering or deciding if we care about the time.

daler:
Agreed. I'm currently working on using this along with https://github.com/bioconda/bioconda-recipes/pull/2595

README.md
```diff
@@ -18,3 +18,19 @@ See the help for the `bioconda-utils` command-line interface for details:
...
+- given a subdag of recipes to build, if one recipe fails then other recipes
+  that depend on it will be skipped (and noted in the log).
+- uses the API from conda 2.0+ and in general works with the latest conda-build
+  versions (instead of maintaining our own branch)
+- test suite for the build tools
+- documentation
+- each built package can now be tested in an isolated busybox container thanks
+  to `mulled-build` and `involucro`. This will catch issues where a recipe fails
+  to specify libs (e.g., libgcc) in the run dependencies.
+- the `mulled-build` testing also means we get per-package docker containers for free
```
bgruening: We need to implement `push` at some point to get the containers into biocontainers. I will do this after this PR is merged and working upstream.

conda-requirements.txt
```diff
@@ -0,0 +1,12 @@
+argh
```
johanneskoester: What is the purpose of this in contrast to the other requirements file above?

test/test_utils.py
```diff
@@ -0,0 +1,573 @@
...
+            package:
+              name: one
+              version: "0.1"
+            requirements:
+              run:
+                - python
+                - numpy x.x
+              build:
+                - python
+                - numpy x.x
```
johanneskoester: So, numpy x.x works now?

travis-setup.sh
```diff
@@ -0,0 +1,39 @@
...
+else
+    tag=MacOSX
+fi
+curl -O https://repo.continuum.io/miniconda/Miniconda3-latest-${tag}-x86_64.sh
+sudo bash Miniconda3-latest-${tag}-x86_64.sh -b -p /anaconda
+sudo chown -R $USER /anaconda
+export PATH=/anaconda/bin:$PATH
+
+# TODO: add remaining pip reqs to conda-forge
+conda install -y --file conda-requirements.txt --channel bioconda --channel conda-forge --channel anaconda
```
johanneskoester: Is the order of the channels also affecting their priorities here? If so, we should maybe have the same order as in bioconda-recipes. anaconda, r, conda-forge, bioconda.

johanneskoester: I'd vote for simply using an environment.yml file. We can specify the channels, and both conda and pip dependencies in one place.

travis-setup.sh
```diff
@@ -0,0 +1,39 @@
...
+
+python setup.py install
+
+pip install -r pip-test-requirements.txt
+pip install -r pip-requirements.txt
+
+mkdir -p /anaconda/conda-bld/osx-64 # workaround for bug in current conda
+mkdir -p /anaconda/conda-bld/linux-64 # workaround for bug in current conda
+conda index /anaconda/conda-bld/linux-64 /anaconda/conda-bld/osx-64
```
johanneskoester: Above three lines are most likely obsolete.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+os:
+  - linux
+language: python
+  - "3.5"
+sudo: required
+services:
+  - docker
+install: "bash travis-setup.sh"
```
johanneskoester: If travis-setup.sh is executable, you can omit the bash call.

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,402 @@
...
+
+        self._pull_image()
+        self._build_image()
+
+    def _pull_image(self):
+        """
+        Separate out the pull step to provide additional logging info
+        """
+        logger.info('Pulling docker image %s', self.image)
+        p = sp.run(['docker', 'pull', self.image], stdout=sp.PIPE,
```
johanneskoester: Just out of curiosity, because I am considering this for Snakemake: Did you experience problems with docker-py, or why did you decide to use plain system calls instead?

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,402 @@
...
+                   stderr=sp.STDOUT, check=True, universal_newlines=True)
+        logger.debug('stdout: %s', p.stdout)
+        logger.debug('stderr: %s', p.stderr)
```
johanneskoester: If STDERR is piped into STDOUT (which happens via `stderr=sp.STDOUT)`, there is no need to log it separately.

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,402 @@
...
+            cmd,
+            stdout=sp.PIPE,
+            stderr=sp.STDOUT,
+            check=True,
+            universal_newlines=True)
+
+        if p.returncode != 0:
+            raise ValueError(
+                p.stdout,
+                p.stderr)
```
johanneskoester: Again, stderr is superfluous, because of the redirecting above.

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,402 @@
...
+            '/bin/bash', '/opt/build_script.bash',
+        ]
+
+        logger.debug('cmd: %s', cmd)
+        p = sp.run(cmd, stdout=sp.PIPE, stderr=sp.STDOUT, universal_newlines=True, check=True,)
+        if p.returncode != 0:
+            raise DockerCalledProcessError(
+                cmd=cmd,
+                stdout=p.stdout,
+                stderr=p.stderr,
```
johanneskoester: Redirection of stderr again.

bioconda_utils/docker_utils.py
```diff
@@ -0,0 +1,402 @@
...
+            '-v', '{0}:/opt/build_script.bash'.format(build_script),
+            '-v', '{0}:{1}'.format(self.pkg_dir, self.container_staging),
+            '-v', '{0}:{1}'.format(recipe_dir, self.container_recipe),
+        ] + env_list + [
+            self.tag,
+            '/bin/bash', '/opt/build_script.bash',
+        ]
+
+        logger.debug('cmd: %s', cmd)
+        p = sp.run(cmd, stdout=sp.PIPE, stderr=sp.STDOUT, universal_newlines=True, check=True,)
```
johanneskoester: If `check=True`, `sp.run` will already check the exit code and fail with a CalledProcessError. Hence, I would think that DockerCalledProcessError will never be raised here. We can just remove it entirely.

bioconda_utils/example_config.yaml
```diff
@@ -1,10 +1,11 @@
...
 upload_channel: bioconda
 channels:
-  - anaconda
-  - r
-  - bioconda
-#setup:
-#  - pip install git+https://github.com/bioconda/conda-build.git@reintroduce-skip-existing
+    - r
+    - anaconda
+    - bioconda
```
johanneskoester: Let's use the same channel order as in bioconda-recipes.

bioconda_utils/build.py
```diff
@@ -0,0 +1,351 @@
...
+        Channels to include via the `--channel` argument to conda-build
+
+    docker_builder : docker_utils.RecipeBuilder object
+        Use this docker builder to build the recipe, copying over the built
+        recipe to the host's conda-bld directory.
+    """
+    env = dict(env)
+    logger.info(
+        "BIOCONDA BUILD START %s, env: %s",
+        recipe, ';'.join(['='.join(map(str, i)) for i in sorted(env.items())])
```
daler: Each yielded env from an EnvMatrix is a list of tuples like `[('CONDA_BOOST', '1.60'), ('CONDA_PY', '2.7'), ('CONDA_NPY', '110')]`. I think it's easier to keep them as tuples so they can be used to update os.environ.

That said, this log entry might be simplified by using `utils.built_package_path(recipe, env)` and printing that package+version+buildstring instead of the full env. Preferences?

bioconda_utils/build.py
```diff
@@ -0,0 +1,351 @@
...
+                        except sp.CalledProcessError as e:
+                            print(e.stdout.decode(), file=sys.stderr)
+                            if b"already exists" in e.stdout:
+                                # ignore error assuming that it is caused by
+                                # existing package
+                                pass
+                            else:
+                                raise e
+                    else:
+                        logger.error(
```
daler: Will do.

Also note that this code path is totally untested at the moment. I'd like to set up a bioconda-utils channel with full read/write anaconda client access. Once the encrypted ANACONDA_TOKEN is available in the .travis.yaml, we can test package uploading (and then clean up when done). Without such a test channel, we would have to wait for merge into bioconda-recipes master to learn that the uploading fails.

bioconda_utils/cli.py
```diff
@@ -35,36 +37,62 @@
 @arg('--force',
      help='Force building the recipe even if it already exists in '
      'the bioconda channel')
-@arg('--docker', nargs='?', metavar='URL', const='unix://var/run/docker.sock',
-     help='Build packages in docker container (continuumio/conda_builder_linux).')
+@arg('--docker', action='store_true',
+     help='Build packages in docker container.')
 @arg('--loglevel', help="Set logging level (debug, info, warning, error, critical)")
+@arg('--mulled-test', action='store_true', help="Run a mulled-build test on the built package")
+@arg('--build_script_template', help='''Filename to optionally replace build
```
daler: It actually works the same (at least with the `@arg` decorator). The benefit of the multiline is easier editing without having to manually re-wrap.

conda-requirements.txt
daler: In travis-setup.sh, I'm just trying to use conda to quickly install things that are available, and then add pip stuff later, i.e.,

```
conda install -y --file conda-requirements.txt --channel bioconda --channel conda-forge --channel anaconda
python setup.py install
pip install -r pip-test-requirements.txt
pip install -r pip-requirements.txt
```

I think I remember seeing a way of installing pip packages with conda so the separate files are not necessary but haven't played around with that.

test/test_utils.py
daler: Yep! Turns out the conda-build recipe rendering really wants stuff in os.environ rather than passed as context dicts for jinja templating. I think the x.x is from pre-jinja days, so it was especially sensitive to having things in os.environ. 

The trick was to use `utils.temp_env` to dump an env dict into os.environ, and now it works.

travis-setup.sh
daler: I had tried taking them out but got failures, that may have been pre-conda-build-2.0 though. Will try again.

conda-requirements.txt
daler: (ah right, environment.yml as you describe below can handle pip + conda)

kyleabeauchamp:
Ive seen bugs using env update.  Imho env create is more robust. Ymmv

daler:
@johanneskoester thanks so much for the review. I've made most the changes with the exception of environment.yml and the indexing steps in travis-setup.sh.

bioconda_utils/build.py
johanneskoester: Sorry, I should have remembered that myself, since it is my own implementation ;-). Still, I think you can get rig of the `map(str, i)` then. A join on the tuple should do.

bioconda_utils/cli.py
johanneskoester: Ok, fine!

travis-setup.sh
johanneskoester: I don't think so. But you can install conda-build into the root env, and then instantiate environment.yml. However, this would also mean having two things. Let's keep it like this and try to add the pip packages to conda-forge at some point.

bioconda_utils/build.py
daler: I had tried that, but it turns out conda-build now expects integer python versions (27, 34, 35). For example the new env_matrix.yml over in bioconda-recipes (https://github.com/bioconda/bioconda-recipes/pull/2595) uses integers now. For flexibility in bioconda-utils I'm allowing int or str in CONDA_PY, so the extra map() make the `'='.join` happy. I could just dump the tuple straight to the log message, but it seemed like all the quotes, commas, and parentheses clutter things up.

johanneskoester:
@daler, feel free to merge and use it in bioconda-recipes. Thanks again for the great work!

PR #32, opened 2016-11-12
Container test and push improvements

bgruening:
This PR fixes the container tests and implements a feature for future automatic pushes to the biocontainers registry.

ping @daler


daler:
@bgruening this will be great. I'm guessing that the `push_containers` script is intended to be called in the recipes' `travis-run.sh`, and that it will need an encrypted token?

bgruening:
@daler yes that's the plan! Hopefully it works out ;)

PR #34, opened 2016-11-16
mulled-build exit checks

daler:
xrefs:

- https://github.com/bioconda/bioconda-recipes/pull/2973
- https://github.com/bioconda/bioconda-recipes/issues/2957

and probably others I can't find at the moment. Several fixes here: the mulled-build failure detection (based string in stderr) was happening *after* the logger reported success, which was confusing. Second, by returning False right away, we didn't have a chance to log the errors within `build()`. And third, the logger level for reporting those mulled-build errors was set to debug rather than error.

ping @bgruening, @mvdbeek, @martin-mann

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -125,6 +125,7 @@ def build(recipe,
 
     res = pkg_test.test_package(pkg_path)
 
+    # TODO remove the second clause once new galaxy-lib has been released.
     if (res.returncode == 0) and (res.stdout.find('Unexpected exit code') == -1):
         logger.info("BIOCONDA TEST SUCCESS %s, %s", recipe, utils.envstr(env))
         logger.debug('STDOUT:\n%s', res.stdout)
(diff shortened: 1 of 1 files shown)


bioconda_utils/build.py
```diff
@@ -125,17 +125,15 @@ def build(recipe,
 
     res = pkg_test.test_package(pkg_path)
 
-    if res.returncode == 0:
+    if (res.returncode == 0) and (res.stdout.find('Unexpected exit code') == -1):
```
johanneskoester: So, there are cases where the mulled test fails, but the exitcode is 0? How come? Will that be fixed within mulled?

bgruening: It already is fixed in galaxy-lib just waiting for a new release.

bgruening:
Great. The error and debug level changes gmakes a lot of sense. :+1: from me.

bioconda_utils/build.py
mvdbeek: the loglevel change is still important, I'd say.

johanneskoester: Of course, I forgot about that :-)

bgruening: Hopefully next week.

johanneskoester:
Added to recipes repo in PR bioconda/bioconda-recipes#2981.

bgruening:
Thanks!

PR #38, opened 2016-11-23
better recipes discovery

daler:
https://github.com/bioconda/bioconda-recipes/issues/2915


daler:
@johanneskoester despite creating the keypairs without a passphrase, `ssh-add` is still asking for it when running on travis, so builds timeout.  Not sure what else to try. Could you give it a shot? I'm mostly following https://gist.github.com/domenic/ec8b0fc8ab45f39403dd, basically:

- generate keypair with no passphrase
- add public key to bioconda-utils settings
- run `travis encrypt-file` on the private key, keeping track of the unique ID it reports
- add encoded file to repo and change ENCRYPTED_LABEL in .travis.yml

daler:
@johanneskoester never mind -- I had left of the `-d` from `openssh aes-256-cbc` so it never got decrypted, and default is to ask for passphrase.

PR #46, opened 2016-12-21
Resolve symlinks in tempdir path

lparsons:
On OSX /var is a symlink to /private/var. This causes problems for
mounting in Docker unless the symlinks are resolved first. See:

* https://docs.docker.com/docker-for-mac/osxfs/#/namespaces
* https://forums.docker.com/t/getting-mounts-denied-on-previously-working-container-after-upgrading-to-v1-12-0-beta18-3-gec40b14/17546/8
* https://forums.docker.com/t/docker-for-mac-beta-does-not-handle-symlinks-when-mounting-volumes/10955/9

### DIFF
--- a/bioconda_utils/docker_utils.py
+++ b/bioconda_utils/docker_utils.py
@@ -388,7 +388,7 @@ class RecipeBuilder(object):
         self.conda_build_args = build_args
 
         # Write build script to tempfile
-        build_dir = tempfile.mkdtemp()
+        build_dir = os.path.realpath(tempfile.mkdtemp())
         with open(os.path.join(build_dir, 'build_script.bash'), 'w') as fout:
             fout.write(self.build_script_template.format(self=self))
         build_script = fout.name


daler:
Great, thanks for finding this. As you're quickly finding out, few of us have OSX development machines . . .

Is docker available in OSX VMs on travis-ci? If so, would you please add a test for this?

lparsons:
To be honest, @daler I don't really know how to go about adding a test, though I agree, it would be a good idea. Sort of seems to work for me (at least I get past this obvious issue). However, I'm still not able to get any builds to work locally. `AttributeError: 'FileNotFoundError' object has no attribute 'read'`. Could still be related to this mounting problem or not. Not enough info in the debug output for me to know. See gist: https://gist.github.com/lparsons/0547a36b15224c7b6dd13d2cd693528e

daler:
re: testing, we'd need to get a way for travis-ci to create the same error you were seeing, then apply this fix and make sure the same error goes away.

I feel like I've seen that FileNotFound error before, possibly related to docker container updates. But it looks like the correct version is being installed.

bgruening:
@daler I think we can merge this as it is, correct. I was not able to figure get a test running, but I can also not see how this patch make thinks worse.

Issue #47, opened 2016-12-21
Travis builds are skipped based on host os, not TRAVIS_OS_NAME


lparsons:
If a recipe specifies it should be skipped on OSX, and I run `TRAVIS_OS_NAME=linux ./simulate-travis.py --docker --packages quast` I expect that the test will be run on Linux. However, instead, no tests are run:
```
DEBUG bioconda_utils.utils:tobuild(486): BIOCONDA FILTER: not building quast-4.3-boost1.60_0.tar.bz2 because it defines skip for this env
DEBUG bioconda_utils.utils:tobuild(486): BIOCONDA FILTER: not building quast-4.3-boost1.60_0.tar.bz2 because it defines skip for this env
DEBUG bioconda_utils.utils:tobuild(486): BIOCONDA FILTER: not building quast-4.3-boost1.60_0.tar.bz2 because it defines skip for this env
```

If I remove the `# [not osx]` from `meta.yaml` and run the same command, the linux build is attempted.

daler:
My guess is that it's because conda is reading the `# [not osx]`, not bioconda-utils. Bioconda reports what conda tells it, and conda is probably checking something like `os.platform`.

I think the solution would be for bioconda-utils to check the combination of TRAVIS_OS_NAME and whatever mechanism conda uses, and make the decision to skip based on both.

daler:
@lparsons looking more closely at this . . . `conda_build.metadata.MetaData` object only reports *that* a recipe should be skipped given the current environment, not *why*. So from `bioconda-utils` we can't differentiate between a recipe being skipped due to OS or other things like py2 or py3.

So instead, the strategy in #48 is to fool `MetaData` by temporarily overriding `sys.platform`. Can you give #48 a try?

daler:
@lparsons closing for now; please reopen if #48 doesn't work for you.

PR #48, opened 2017-01-03
Address #47

daler:


daler:
Seemed a bit hacky, but hopefully it works!

Issue #51, opened 2017-01-04
recipes with no tests should automatically fail


daler:
xref: https://github.com/bioconda/bioconda-recipes/issues/124#issuecomment-270265869

daler:
closed via https://github.com/bioconda/bioconda-utils/pull/57

Issue #50, opened 2017-01-04
cross-reference conda-forge and bioconda channels


daler:
Need a good mechanism for detecting whether a package on bioconda has been replaced by one on conda-forge and can therefore be moved to a `recipes-deprecated` subdir.

xref https://github.com/bioconda/bioconda-recipes/issues/2796

daler:
#57 works at a package/version level to flag recipes as duplicated across channels. While we can dig deeper into dependencies and environments, it may be best to keep it at this level and rely on manually reading recipes to decide on whether to deprecate.

Issue #23, opened 2016-10-09
linters


daler:
Need linters for existing recipes to flag easily-identified issues.

For example, if gcc is a dependency, then llvm should be a dependency for osx and libgcc should be a run dependency.

I'm picturing this as a module of functions that each take a recipe and check for a discrete issue, returning a list of issues identified. Adding new functionality to the linter would be a matter of just adding a new function.

daler:
List of linter functions to add, please edit and add more:

- hash must exist
- license flagged if not in controlled vocabulary list of licenses

bgruening:
@daler here are a few ideas: https://github.com/bioconda/bioconda-recipes/issues/1860

daler:
thanks, was just looking for that issue :)

peterjc:
Another suggestion:

* Check valid formatting for dependency versions

e.g. preempt ``conda.exceptions.CondaRuntimeError: Runtime error: Invalid version spec: >=`` from ``libgd >= 2.2`` rather than ``libgd >=2.2`` on https://github.com/bioconda/bioconda-recipes/pull/4655

epruesse:
Closed as all but the "license flagged if not in controlled vocabulary list of licenses" lint now exist.

PR #57, opened 2017-01-13
Linting

daler:
This PR adds a linter for doing QC on bioconda recipes and a system for easily adding new kinds of QC checks.

There was previously some discussion about using the conda-forge linter, but after looking at that I realized we have different needs. So I went back to the drawing board. The output of this is a TSV that can filtered, summarized, tabluated, etc for QC issues. See [this gist](https://gist.github.com/daler/332499b79eeb7ac5df2bcd258f673000) for an example report run on all existing recipes, which was generated by:

```bash
bioconda-utils lint recipes config.yml --packages "*"  | awk '{if ($2!="already_in_bioconda") print $0}' 
```

After a few iterations, I think the architecture is now pretty simple. Adding new things to check is very straightforward -- see `lint_functions.py`.

There's some room for improvement on the output format. I'm not sure the `info` field, and the further expansion of it into extra columns, is needed. I had initially used that for listing which other channels a package is found in. But after adding more checks, that one ended up being the only check that needed extra info.

However I think this is already a useful tool for inspecting existing recipes, and can also be run on new recipes to check for issues. That can be done manually (maybe using `simulate-travis.py`), or automatically in the travis build or as a github bot.

Hopefully this will help reduce workload on checking and correcting incoming PRs. Actually, given Bjoern's activity in monitoring PRs maybe we should call the linter `bjoern-o-matic` or even `what-would-bjoern-say.py` . . .

cc @bioconda/core; xrefs:
- #51
- #50 
- #23 
- bioconda/bioconda-recipes#1860
- https://github.com/bioconda/bioconda-recipes/issues/3417
- and probably some others I'm missing.


johanneskoester:
Great work!! I'd vote for putting this into the travis builds (before building), and let them fail if something is wrong.

bgruening:
Yeah, awesome work. Fail early and fail often ;)
@daler I curious have you looked at the anaconda-linter or the conda-forge linting? Is this somehow compatible?

bioconda_utils/lint_functions.py
```diff
@@ -0,0 +1,207 @@
...
+        i for i in
+        [reqs.get('build'), reqs.get('run'), reqs.get('test')]
+        if i is not None
+    ]
+
+
+def uses_perl_threaded(recipe, meta, df):
+    if 'perl-threaded' in _get_deps(meta):
+        return {
+            'depends_on_perl_threaded': True,
```
bgruening: really up to date I see :)
Thanks a bunch!

bioconda_utils/lint_functions.py
```diff
@@ -0,0 +1,207 @@
...
+        return {
+            'depends_on_java-jdk': True,
+            'fix': 'use "openjdk" instead of "java-jdk"',
+
+               }
+
+def uses_setuptools(recipe, meta, df):
+    if 'setuptools' in _get_deps(meta):
+        return {
+            'depends_on_setuptools': True,
```
bgruening: Does this fail the linter? Do we have a concept of a warning?

bioconda_utils/lint_functions.py
```diff
@@ -0,0 +1,207 @@
...
+        src = meta['source']
+        if src is None:
+            return
+    except KeyError:
+        return
+
+
+    if not any(
+        (
+            'md5' in src,
```
bgruening: Do we want to force a strong hash?

johanneskoester: Sometimes, md5 is even stronger than sha256 in my opinion. This is the case if md5 is provided by upstream. I think it is always better to use the upstream hash than a hash that you have to compute on your own, because at that time you already might have downloaded a corrupted source.

daler:
@bgruening conda-forge has a web service here https://github.com/conda-forge/conda-forge-webservices/blob/master/conda_forge_webservices/linting.py

but the actual work is done here: https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/lint_recipe.py

However https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/lint_recipe.py#L42 is roughly compatible in that it takes a parsed meta and the recipe dir. So if we wanted to also use conda-forge's linting, a new linting function in our `lint_functions.py` would then be:
```python
import conda_smithy.lint_recipe
def condaforge(meta, recipe, df):
    results = lint_recipe.lintify(meta, recipe)
    return {'conda-forge says': results}
```

bioconda_utils/lint_functions.py
daler: `lint_functions.registry` actually used to be a dict of `{function: severity}` which made it into a column in the TSV. To me it seemed like clutter, but trivial to re-implement.

Right now, the linter is for information rather than pass/fail. I'm sort of treating everything as warnings, since even missing a source section can happen if you're looking at a metapackage. 

We can easily adjust as needed though.

daler:
@johanneskoester with respect to failing tests, I can add a mechanism for severity of problems, have the CLI pay attention to that, and `sys.exit(1)` if the the most severe problems are found. For example missing tests would be ERROR, using `git_url` would be WARNING, using setuptools would be WARNING or maybe INFO.

Or maybe it should be built into `bioconda_utils.build.build_recipes`, since that does detection of which packages need to be built. Though a common error is to forget to bump version or build number, and those cases won't be caught in `build_recipes`. So maybe we should keep the linting separate and check what recently changed in git rather than rely on what is detected as needing to be built.

bioconda_utils/lint_functions.py
bgruening: I know what you are referring too, yet md5 is a strong checksum anymore, so we should encourage our users to use sha256 if possible.

bgruening:
@daler I would do this before fitlering, just on the git diff of TRAVIS_COMMIT_RANGE. This can fail very fast for a few tests. We can think about enabling a few other linters during the build-step if this is needed but currently, also in respect to the build time and reviews I would say fast linting and failing in bad cases it needed.

bioconda_utils/lint_functions.py
johanneskoester: Indeed, I agree. As long as we don't enforce it, something like an advice should be fine.

johanneskoester:
@daler @bgruening I agree. Everything that changed in git should be linted. With a concept of warnings and failures, it would be good to have a separate status in the PR for the linting. Looks like this can be done "easily" with curl and the github REST API: https://developer.github.com/v3/repos/statuses/. The only question is what to put in there for `sha` (is that the git commit?).

daler:
@bgruening @johanneskoester I've been playing around with creating statuses. I can get it to work if I use a personal access token from my own account, but organizations don't have personal access tokens. Organizations can use OAuth2, so I was playing around with that. I was not successful because it requires setting up a server to receive the OAuth2 token.

(using PyGithub and personal access token `TOKEN`:)

```python
from github import Github
g = Github(TOKEN)
user = g.get_user('daler')
repo = user.get_repo('connect-test')
commit = repo.get_commit('1b6de0b14505b3e3c692a01917e7534744c1ac91')
commit.create_status(state='error', description='linter found a problem', context='linting')
```

The result is here: https://github.com/daler/connect-test/pull/3

I think the best option will be to have an OAuth2 `client_id`, `client_secret`, and `access_token` stored as encrypted env vars on travis. Getting `access_token` is the tricky part. As far as I can tell the quickest solution will be to temporarily spin up an AWS instance with a server to accept the access token response. Any easier suggestions?

bgruening:
@daler please have a look at https://github.com/conda-forge/conda-forge-webservices
They are using https://www.heroku.com which might be free for our usecase.

daler:
If you mean using heroku like conda-forge does, I was thinking that, but looks like the free tier "sleeps after 30 mins of activity". Not sure exactly what that means. I was hoping that by providing the credentials in the travis env, everything could be done there at build time.

johanneskoester:
I agree. I don't think we need heroku just to do the linting. This post http://stackoverflow.com/a/31160721/7070491 suggests to create a machine account or use a personal access token of one user. A machine account sounds like somebody would have to remember the credentials. I'd say I just create a personal access token with my account instead.

bioconda_utils/utils.py
```diff
@@ -639,3 +639,26 @@ def get_list(key):
 
     default_config.update(config)
     return default_config
+
+
+def modified_recipes(git_range, recipe_folder):
```
daler: @johanneskoester some changes here that might be useful, specifically making `--relative` be the provided recipe dir, and using `shell=True` which allows the wildcards to be expanded. I think the existing code in master looks for the literal filename "recipes/*/meta.yaml" without expanding the glob.

johanneskoester: No, the globbing actually works fine. I tested it extensively. I think the reason is that git expands it internally. Let's not use shell=True if it is not needed. I agree with `--relative`, though.

daler: It wasn't working for me locally without `shell=True`. But after your comment about git expanding the glob I upgraded git from 1.9.1 to 2.11 and now it works even with no `shell=True`. Something to consider if/when this is run in `simulate-travis.py` when folks may not necessarily have the latest git.

daler:
I built in the ability to push status updates to github, but only later realized that it won't work in pull requests. Pull requests from forks (the model we use) don't have access to travis' encrypted env vars, which means the GITHUB_TOKEN doesn't exist and the status update fails. I think it would work from master branch, but that defeats the purpose.

It would have been a nice touch to have a separate check and status for linting, but for now the linting is rolled into the travis build. If linting fails, everything fails.

johanneskoester:
OK, that's fine for me. As long as the linting is clearly visible in Travis this is not a problem at all.

bioconda_utils/utils.py
johanneskoester: I see. Maybe we should do a version check.

daler:
OK, the combination of this and https://github.com/bioconda/bioconda-recipes/pull/3933 seems to be working. Can you guys review when you get a chance please?

bioconda_utils/lint_functions.py
```diff
@@ -0,0 +1,186 @@
...
+    name = meta['package']['name']
+    version = meta['package']['version']
+
+    # Some recipes have an empty build section, so it'll be None and we can't
+    # do a chained get.
+    build_number = 0
+    build_section = meta.get('build')
+    if build_section:
+        build_number = int(build_section.get('number', 0))
+    return df[
```
johanneskoester: Could the DF have a hierarchical index with name, version and build number, or is this operation to infrequent to justify this?

bioconda_utils/lint_functions.py
```diff
@@ -0,0 +1,186 @@
...
+        return
+
+
+def _get_deps(meta):
+    reqs = meta.get('requirements')
+    if reqs is None:
+        return []
+    return [
+        i for i in
+        [reqs.get('build'), reqs.get('run'), reqs.get('test')]
```
johanneskoester: Does this only return the dependency name or also the version string? In case of the latter, the test below would not see `perl-threaded`, right?

bioconda_utils/lint_functions.py
johanneskoester: But the test fail if any lint is produced, right?

johanneskoester: Apart from that, I don't really understand this lint. If the package needs setuptools, why should it not be listed in the dependencies? I'd rather think the lint should complain if it is listed in the run dependencies.

bioconda_utils/linting.py
```diff
@@ -0,0 +1,227 @@
...
+            ['git', 'log', '--format=%B', '-n', '1',
+             os.environ['TRAVIS_PULL_REQUEST_SHA']]
+        )
+        txt = p.stdout
+
+    # For example the following text in the commit message will skip
+    # lint_functions.uses_setuptools:
+    #
+    # [ lint skip uses_setuptools ]
+    print('txt:', txt)
```
johanneskoester: I guess the line above was debugging output?

bioconda_utils/lint_functions.py
daler: True, it would be cleaner to do `df.loc[name, version, build_number]`, but only if all these keys exist. In the case where a build number was bumped and so doesn't exist in the df, you'd have to catch the KeyError and then return an empty dataframe. So I think I'd prefer to keep it flat for now, unless I'm missing something in pandas where it handles missing hierarchical keys gracefully.

johanneskoester: Ok, that's reasonable.

bioconda_utils/lint_functions.py
daler: In my tests this gets the dependencies as listed in the meta.yaml, so version string is not included and the `uses_perl_threaded` test works. In fact many of our existing recipes fail the test because they have perl-threaded (but I think the perl team already knows that)

bioconda_utils/lint_functions.py
daler: Right, currently tests fail if any lint is produced, but there is a mechanism for skipping them on a case-by-case basis. But good point here on setuptools as run dependency. I had added this because I remember Bjoern pointing out setuptools in meta.yamls, but I don't remember if this was in the run or build. I'll change it to build.

daler: (sorry, I mean I'll change it to *run*)

bioconda_utils/linting.py
daler: Oops, thanks.

daler:
@johanneskoester, @bgruening after some testing I realized the following issue with the lint check "already_in_bioconda":

It happens like this: I create a branch off of master. I update a recipe. Then I pull recent changes from master into my working branch (I generally use `git pull upstream master`). Since those changes were in master, the packages were since uploaded to the channel.  The problem is that `git diff master HEAD` in my now-updated branch shows those merged-in changes from master as different. So those recipes are linted, but then fail because they already exist in the channel.

Possible solutions:
- There's probably a git command for not committing the merge from master as a separate commit. We'd have to ensure everyone uses that method when merging from master.
- Maybe a better `git diff` command that would solve this?
- Disable the `already_in_bioconda` lint. Since the downstream parts of the build system (the filtering step) will not build packages that exist, the `already_in_bioconda` lint was just an extra way of explicitly calling this out up front and failing early.

Any other ideas?

bgruening:
I would say never ever work on master ;) Is this to hard? It would solve it I guess.
@daler `already_in bioconda` should be changed to `already_available` and should check `r,conda-forge,bioconda,default`. I agree that testing bioconda has not that much value, but checking conda-forge does. For example the recent `atlas` contribution, which has a different package in conda-forge.

daler:
@bgruening I actually don't work on master but I do pull changes from upstream master to make sure my branch is up-to-date. It's that pull from upstream master that causes the problem. Is there a better way of doing this?

And actually there's already a separate check for `in_other_channels` that does exactly what you're saying :).

bgruening:
Ha, perfect. Reviewing this beast is on the todo list for this evening :)
Does `git diff master...branch` works?

daler:
Ah! I didn't know about the triple-dot syntax. I think that's the trick. Let me do a little testing and then I'll change to that if it works

daler:
@bgruening that was it, thanks.

Sorry to ruin your evening on this :) Don't hesitate to ask about anything I can help explain. It's easiest to have both bioconda-utils and bioconda-recipes checked out to their respective `linting` branches. Also it's as-yet-undocumented, but the current mechanism for controlling linting and building in `simulate-travis.py` is with env vars BIOCONDA_UTILS_LINT_ARGS and BIOCONDA_UTILS_BUILD_ARGS, so like:

```bash
BIOCONDA_UTILS_LINT_ARGS="--cache .cache" ./simulate-travis.py
```

johanneskoester:
Very nice. Thanks for working this out! I can't wait to merge this :-)

daler:
Our cluster is down until Wed so I am considering merging soon and then remaining on standby to fix problems as they pop up. @bgruening if you haven't had a chance to review do you mind if I merge?

bgruening:
Actually, I did, but only half of all linters. So far everything worked ... so :+1: from my side!
This is a great improvement! Thanks a bunch!

daler:
OK here we go! My plan is to merge this first and wait for docs to build. Then switch over bioconda-recipes to use the linting, and finally create an issue documenting the changes to `bioconda/all`

PR #59, opened 2017-01-22
Control conda versions

daler:
This PR makes it easier to control conda and conda-build versions, via env vars or the CLI. This will let us add

```yaml
global:
  - BIOCONDA_CONDA_VERSION=4.2.15
  - BIOCONDA_CONDA_BUILD_VERSION=2.0.7
```
to `bioconda-recipes/.travis.yaml` so that there is an obvious place to set the versions.

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -50,7 +50,13 @@ logger = logging.getLogger(__name__)
      ignored.''')
 @arg('--conda-build-version',
      help='''Version of conda-build to use if building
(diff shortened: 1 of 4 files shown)


johanneskoester:
I think I'd prefer the docker container to use exactly the conda version that the host uses. The reason is that otherwise we might have to ensure the conda version in two places (host, for osx, and container, for Linux). Using the bioconda-utils requirements.txt sounds like a better option to me.

daler:
closed via #84

PR #71, opened 2017-02-24
Filter recipes by git range

johanneskoester:
This moves some of the functionality from PR bioconda/bioconda-recipes#3891 into bioconda-utils. The benefit is that `scripts/travis-run.sh` will be simpler again, and the functionality becomes available locally. In the long run, it can also deprecate `--quick`, since it is more general.

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -106,18 +106,19 @@ def build(recipe_folder,
     # handle git range
     if git_range:
         p = sp.run(['git', 'diff',
-                          '--exit-code',
                           '--relative=recipes',
                           '--name-only'] + git_range +
                          [os.path.join(recipe_folder, '*', 'meta.yaml'),
                           os.path.join(recipe_folder, '*', '*', 'meta.yaml')],
                          stdout=sp.PIPE,
-                         universal_newlines=True)
-        if p.returncode == 0:
+                         universal_newlines=True,
+                         check=True)
+        modified = p.stdout.strip().split('\n')
+        if not modified:
             logger.info('No recipe modified according to git, exiting.')
             exit(0)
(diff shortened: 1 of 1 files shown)


daler:
Nice!

bioconda_utils/cli.py
```diff
@@ -98,6 +103,24 @@ def build(recipe_folder,
     else:
         docker_builder = None
 
+    # handle git range
+    if git_range:
+        p = sp.run(['git', 'diff',
+                          '--relative=recipes',
+                          '--name-only'] + git_range +
+                         [os.path.join(recipe_folder, '*', 'meta.yaml'),
```
bgruening: @johanneskoester if we leave this out here and consider all files in the `run` function and filter later by `modified = [mod for mod in modified if mod.endswith('mata.yaml')]` this should be more stable I assume.

johanneskoester: Maybe, but also a bit less explicit. I don't think it makes much of a difference. I plan to anyway refactor this a bit in the next days.

bgruening: How can this be less explicit? It is way more explicit and you are way more flexible in your folder structure. Glob'ing is everything else than explicit.

johanneskoester: Well, the this version shows in what kind of pattern recipes are expected. Using all files and `endswith`, recipes can be everywhere (within the recipe folder). I find this more implicit. But this is rather an explanation what I had in mind, and why I don't really see the advantage of the other approach. Maybe best would be to have a common recipe discovery for both `--packages *` and `--git-range`. In any case, I think I'd rather wait on Ryans refactorings (see the linting PR), because he has also generalized parts of this code.

Issue #72, opened 2017-03-01
LOCKERROR: It looks like conda is already doing something.


sebastian-luna-valero:
Having this problem here:
https://travis-ci.org/bioconda/bioconda-recipes/jobs/206522192#L576
https://travis-ci.org/bioconda/bioconda-recipes/jobs/206536065#L480

I have reported to conda:
https://github.com/conda/conda/issues/3981

glormph:
I got this as well today, and was suspecting it had something to do with the S3 outage earlier.

glormph:
Error seems to have disappeared now, at least when restarting a build in my own repo.

sebastian-luna-valero:
Here is the feedback from the conda team:

> These specific locks have all been removed in conda 4.3. I think upgrading to 4.3 is the best solution at this point. Latest version is 4.3.13.

I am sure we must be using conda 4.2 for a reason in bioconda, is that right?

daler:
Automatically using the latest version has caused problems in the past, so we keep it pinned to avoid surprises. I'm not sure we've tested 4.3 yet.

If you'd like to help with testing, create a new bioconda-utils branch that updates the version here: 
https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/bioconda_utils-requirements.txt

and then in a new bioconda-recipes branch, point to your new bioconda-utils branch here:

https://github.com/bioconda/bioconda-recipes/blob/master/.travis.yml#L20

sebastian-luna-valero:
Sure, thanks @daler 

Testing here https://github.com/bioconda/bioconda-utils/pull/73

sebastian-luna-valero:
This issue can now be closed.

PR #73, opened 2017-03-02
 test updating pinned versions of conda and conda-build

sebastian-luna-valero:
As requested https://github.com/bioconda/bioconda-utils/issues/72


sebastian-luna-valero:
@daler tests are passing now. Please have a look.

I can see that you are adding a parameterised version of this PR here https://github.com/bioconda/bioconda-utils/pull/59 which I think is a great idea so just go ahead and close it if you want.

daler:
Thanks for working on this. As @johanneskoester pointed out in #59, it will be better to have what you use here. That is, configure the versions in requirements rather than as env vars. We will soon be merging in the linting functionality, so I'd prefer to wait until that goes live and we fix any outstanding issues before merging the conda version updates.

daler:
Closed via #84

PR #74, opened 2017-03-06
Hotfix

daler:
1. Addresses https://github.com/bioconda/bioconda-recipes/issues/4038#issuecomment-284535179.
2. Allows 1 or 2 args to --git-range, and in bioconda-recipes this will default to `--git-range HEAD` so that local tests no longer require a commit
3. Improve changed recipe detection: linting only cares about meta.yaml files but building cares about anything in any recipe dir.


nsoranzo:
Bump version in `bioconda_utils/version.py` ?

PR #77, opened 2017-03-08
use regex to parse out git version for OSX git

tomkinsc:
This adds a regex to parse out the git version in a way compatible with Apple git, since under macOS the git version is reported like this: `git version 2.10.1 (Apple Git-78)`.


daler:
Thanks!

PR #82, opened 2017-03-11
Blacklist handling

daler:
This PR allows `modified_recipes` to also detect recipes that were newly blacklisted in the provided git range. Otherwise a recipe would have be modified in some way before being detected.


johanneskoester:
Nice!

PR #94, opened 2017-04-21
Add test case that checks whether tokens leak into the build environment

johanneskoester:
This PR adds a test that checks whether GITHUB_TOKEN is leaked into the build script.


johanneskoester:
Actually I am suprised by this. I thought that I have discovered a leak [here](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/build.py#L118), but for some reason, GITHUB_TOKEN is not leaked into the build script. Am I missing something?

johanneskoester:
@daler, what do you think about this? Am I missing something?

daler:
@johanneskoester the `utils.temp_env` contextmanager temporarily resets `os.environ`. If you make this test build script into:

```bash
#!/bin/bash
env
exit 1
```
you'll see all the env vars that conda-build sets, but nothing from the original environment. 

That said it would be much clearer here to simply use `p = utils.run(cmd, env=env)`, no need for the temp env.

daler:
Hmm, I take that back, I think it is safer to temporarily reset `os.environ` just to be sure. But still clearer to use `p = utils.run(cmd, env=env)`.

daler:
Aaaaand now that I look closer, the context manager is not resetting but updating os.environ. So I'm back to being confused as well. Sorry for the noise.

johanneskoester:
Nice to see you had exactly the same thoughts as myself, even in the same order :-).

daler:
Hah! I'm in good company then!

After attempting to inject various env vars into build.sh, it appears that conda-build is using a whitelisted env. Looks like it's happening here: https://github.com/conda/conda-build/blob/master/conda_build/build.py#L966. However rendering the template still reads os.environ, so a meta.yaml like:

```yaml
version: {{ GITHUB_TOKEN }}
```
will still leak, and the token will show up in the built package's meta.yaml. Should we use a whitelist?

johanneskoester:
Either that or a blacklist. Blacklist can be really short: `ANACONDA_TOKEN`, `GITHUB_TOKEN`, `INVULCRO_AUTH`.

daler:
Sounds good. I'll do that, add some tests and then push to this PR.

johanneskoester:
Wait, I changed my mind, see above :-). What do you think? Whitelist or blacklist?

daler:
Good points. If the env matrix is storing a var in in plain text then it should be fine to send to conda-build. I'd rather err on the side of too paranoid and maintaining portability, so I'd vote whitelist.

johanneskoester:
I agree!

bgruening:
I just wanted to point out the `script_env` option: https://conda.io/docs/building/meta-yaml.html#use-environment-variables
But it seems this solution here takes already care of this! Good work!

daler:
Thanks @bgruening, didn't know about that

daler:
@bgruening @johanneskoester OK to merge?

johanneskoester:
Sorry for the delay, I thought I had submitted the review above...

bioconda_utils/build.py
```diff
@@ -63,10 +63,22 @@ def build(recipe,
         If mulled_test == True, the name of the mulled docker image will be
         appended to this list.
     """
-    env = dict(env)
+
+    # Clean provided env and exisiting os.environ to only allow whitelisted env
+    # vars
+    _env = {}
+    _env.update({k:str(v) for k, v in os.environ.items() if utils.allowed_env_var(k)})
+    _env.update({k:str(v) for k, v in dict(env).items() if utils.allowed_env_var(k)})
```
daler: It's true that `env` should be coming from the `env_matrix` file and that shouldn't contain any secrets. But it is technically possible to pass `env=os.environ` to this function, so I was trying to guard against that. I don't think this hurts anything as-is, but I'm fine either way.

johanneskoester: Agreed.

johanneskoester:
Feel free to merge!

PR #96, opened 2017-05-03
Upload packages immediately after build

johanneskoester:
This will help with bulk updates: bioconda-utils will try to build as many recipes
as possible until the timeout occurs. Since they are immediately uploaded (of course not in PRs), subsequent builds won't need to rebuilt them when they succeeded in a previous build.

Hence, bulk updates can be easily handled by fixing recipes in an iterative process, without blacklisting.
On the bioconda-recipes side, I intend to handle this via creating pull requests to special bulk/* branches,
such that we don't have failures in master.

In total, I have the hope that this will mostly solve our bulk update issues, such that we can finally proceed to Python 3.6 and R 3.4.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -1,5 +1,5 @@
 import subprocess as sp
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 import os
 import logging
 import networkx as nx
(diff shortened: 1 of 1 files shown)


johanneskoester:
I agree, but having a bulk branch seems cleaner to me. We don't pollute master with the bulk update. It won't look like there is something failing in general. And it is clear that the bulk update is not complete as long as a bulk branch has no checkmark.

Absolutely, bulk branches need to be protected, just like master. Further, PRs that update the env matrix but are not targetting a bulk branch have to fail with an error.

PR #99, opened 2017-05-04
Optionally build only the latest version of a recipe.

johanneskoester:


bioconda_utils/build.py
```diff
@@ -240,11 +241,15 @@ def build_recipes(
 
     logger.info('blacklist: %s', ', '.join(sorted(blacklist)))
 
+    get_recipes = utils.get_recipes
```
johanneskoester: It is handled in the if below.

bgruening: I mean this could be in an else, belonging to the if below.
```
if latest_only:
    get_recipes = utils.get_latest_recipes
else:
    get_recipes = utils.get_recipes
```

johanneskoester: Yes, that is right. I did it this way because I thought it better reflects that there is a default case, and a special case (and possibly more in the future).

epruesse:
@johanneskoester - What is the status of this? Is it still needed?

PR #111, opened 2017-05-16
Custom base image definition via meta.yaml.

johanneskoester:
This PR allows to define a custom base image for mulled via
```
extra:
  container:
    base: "<baseimage>"
```

### DIFF
--- a/bioconda_utils/pkg_test.py
+++ b/bioconda_utils/pkg_test.py
@@ -77,7 +77,8 @@ def get_image_name(path):
 def test_package(path,
                  name_override=None,
                  channels=["r", "defaults", "conda-forge"],
-                 mulled_args=""
+                 mulled_args="",
+                 base_image=None
     ):
     """
     Tests a built package in a minimal docker container.
@@ -98,6 +99,9 @@ def test_package(path,
         Mechanism for passing arguments to the mulled-build command. They will
         be split with shlex.split and passed to the mulled-build command. E.g.,
         mulled_args="--dry-run --involucro-path /opt/involucro"
+
+    base_image : None | str
+        Specify custom base image. Busybox is used in the default case.
 
     """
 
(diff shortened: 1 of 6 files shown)


johanneskoester:
We have to wait for bioconda/bioconda-recipes#4713.

johanneskoester:
@bgruening this is strange: `mulled-build` cannot be found anymore...

bgruening:
Installing it locally seems to work, it is also in the conda tarball ... :(
I try to look at it today.

bioconda_utils/pkg_test.py
```diff
@@ -133,4 +137,8 @@ def test_package(path,
     cmd += channel_args
     cmd += shlex.split(mulled_args)
     logger.debug('mulled-build command: %s' % cmd)
-    return utils.run(cmd)
+
+    env = {}
```
bgruening: You are overwriting here env which is passed to subprocess and this can result in mulled not being on pass, I guess.

johanneskoester: Good catch!

johanneskoester:
@bgruening this looks like a bug in the new mulled-build version: https://travis-ci.org/bioconda/bioconda-utils/jobs/233125954#L1231, am I right?

test/test_pkg_test.py
```diff
@@ -40,16 +58,24 @@ def _build_pkg():
...
-    Make sure calling mulled-build with the wrong arg fails correctly
+    Make sure calling mulled-build with the wrong arg fails correctly.
     """
-    built_package = _build_pkg()
+    built_package = _build_pkg(RECIPE_ONE)
     with pytest.raises(sp.CalledProcessError):
         res = pkg_test.test_package(built_package, mulled_args='--wrong-arg')
+
+
+def test_pkg_test_custom_base_image():
```
bgruening: @johanneskoester nearly working. OSX is still broken I guess we need to skip the tests on OSX with `@pytest.fixture(scope='module', params=PARAMS, ids=IDS)`. I just don't understand why this was passing before.

johanneskoester: Thanks! I have disabled mulled on osx. Let's see if it works now.

johanneskoester:
Yes, finally fixed! Thanks for your help @bgruening!

johanneskoester:
Now we should be able to fix the click-related recipes. Is it ok if I use the debian container image for them? Or did you find a better solution @bgruening?

PR #114, opened 2017-05-23
fix unnecessary multiple builds

daler:

PR #115, opened 2017-05-24
Remove check for the case of having multiple recipes for the same package name.

johanneskoester:
This is common, because we sometimes store multiple versions of the same recipe in subdirectories.
Hence, it should not fail.

The regression was introduced in PR #114. This change should not affect the main benefit of that PR. Since we still use a set, there cannot be multiple builds of the same recipe.


daler:
Ah, thank you for catching this!

It's getting to be time for another round of test-writing I think :)

PR #118, opened 2017-05-25
Improved cleanup

johanneskoester:
If disk space is less than 10 MB, all conda packages are cleaned up.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -21,7 +21,6 @@ def purge():
     free = utils.get_free_space()
     if free < 10:
         logger.info("CLEANING UP PACKAGE CACHE (free space: %iMB).", free)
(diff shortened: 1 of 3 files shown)


bgruening:
@johanneskoester just for your information, mulled is cleaning up after a build and removes everything in it's ./build folder. Everything else is a bug.

johanneskoester:
Ok, thanks. I thought, since I have seen these problems from time to time in travis, it is maybe safer to just have this additional mechanism in bioconda-utils. Now, even if internal cleanup fails, we will never have a problem.

johanneskoester:
Let me add that having done all the migration in the last weeks, including fixing a lot of packages, I am more than ever convinced that mulled is an amazing improvement in our QC.

PR #122, opened 2017-06-03
remove r channel from testing

bgruening:


bgruening:
@daler @johanneskoester this is also ready for review.

daler:
I think you got them all . . . what an annoying thing to grep for!

bgruening:
Sorry was getting food, what happend :) Is it working? Yes, please let's wait an other day. The branch does not need to be empty but at least the strange ICU and readline issues needs to go away.

bgruening:
You also had a similar branch? Btw. there is also `meta.build_number()` to return the build number.

johanneskoester:
Yes, I basically had exactly the same change as you. What I meant was: this branch (kill_r) should not be merged into master before bulk is empty, right? Otherwise, we will run into problems I guess?

bgruening:
Realistically, we need many more days for R. There are really broken recipes. At some point we put all on a blacklist and then we can merge this branch here. Work in a new bulk branch to get rid of the blacklist.

johanneskoester:
Ok, but no need for a new bulk branch. You can simply merge bulk into master at some point (together with the blacklist), and go on in the very same bulk branch (the name is hardcoded and it is protected and so on, so it shall not be deleted).

johanneskoester:
Mhm, after some more thoughts I think we cannot do that. When we blacklist R recipes and move to the new channel order in the master branch, all blacklisted recipes will be broken when people try to install them (e.g. because they are 3.3.1 and there is no r-base 3.3.1 in conda-forge). So, I think the bulk rebuild has to be complete before we merge the bulk branch and this PR into master.

bgruening:
The bulk branch is "complete" there are a few packages blacklisted, but this was true before.
However, I think we can not disable the r-channel completely. As we do not store old versions in git these old versions are not rebuild and this means if people don't activate the r-channel they will not be able to install older version.

What do you think about adding the r-channel as last channel in the priority.

johanneskoester:
Yes, we can do that for a while.
Great news! Thanks for working on this!!

daler:
Awesome work Bjoern!

bgruening:
@daler @johanneskoester ready when you are ready!

docs/source/contribute-a-recipe.rst
```diff
@@ -144,7 +144,6 @@ You should recommend to your users that they set up the same channel order as
 used by bioconda to ensure that all dependencies are correctly met, by doing
 the following::
 
-    conda config --add channels conda-forge
     conda config --add channels defaults
```
johanneskoester: Above line can be removed now.

johanneskoester:
Really nice!

PR #123, opened 2017-06-03
test latest conda-build version

bgruening:


nsoranzo:
The second error in the Travis build is due to this change: https://github.com/conda/conda-build/commit/2fcded8c67259ad3c2cd5b5f4acdc8241a1c9122#diff-a273b977a04f704063ec213a6230e203R68

Changing:
```
print(bldpkg_path(metadata, metadata.config))
```
to
```
print(bldpkg_path(metadata))
```
at line 467 of `test/test_utils.py` should fix it.

test/test_utils.py
```diff
@@ -464,7 +464,7 @@ def test_filter_recipes_extra_in_build_string():
     from conda_build.render import bldpkg_path
 
     metadata = MetaData(recipe, api.Config(**dict(CONDA_EXTRA='asdf')))
-    print(bldpkg_path(metadata, metadata.config))
+    print(bldpkg_path(metadata))
```
johanneskoester: Maybe it would be better to use API.get_output_file_path as we use it for filtering recipes.

bgruening: You mean `api.get_output_file_path(recipe, metadata.config)`

johanneskoester: Yes, I think so.

johanneskoester: Just to be sure that we capture the same problem as we see it when filtering.

johanneskoester:
Seems like it finally works!

bgruening:
Cool, let's see if this fixes the bulk branch.

johanneskoester:
Make sure that you merge master into bulk before starting a rebuild. Otherwise, it might rebuild the duplicate recipes again that we are now deleting from anaconda.

PR #127, opened 2017-06-07
Implement functionality to automatically remove duplicate packages.

johanneskoester:
This does not work due to an anaconda-client bug though :-): we run the correct command, but anaconda-client apparently creates a wrong REST API call.

Before this can be merged, PR #130 has to be merged.


bgruening:
This you mean with flags, indeed tried that as well and it did not work :(
So manual removal?

johanneskoester:
Yes. I will first implement noarch and then start from the bottom.

daler:
I think you need to explicitly provide the token -- we're already doing this to test uploading: https://github.com/bioconda/bioconda-utils/blob/master/test/test_utils.py#L166

johanneskoester:
Good idea. I will try.

johanneskoester:
Unfortunately, this does not help. It simply seems to use a deprecated URL.

daler:
Ugh, and we are already using the latest anaconda-client. That also means at some point we started accumulating test packages, which will be consuming some space. Prob not much, but non-zero.

daler:
https://anaconda.org/bioconda/repo/files?type=all&label=bioconda-utils-test does not have a big list though, so they are somehow getting deleted

bgruening:
@daler I did this in the morning we were running out of space and did a huge cleanup.

daler:
Ah, never mind, just checked the anaconda.org log and @bgruening did some cleanup a couple of hours ago.

daler:
Thanks! And I just checked gitter, should have checked that before posting here

daler:
According to https://anaconda.org/bioconda/repo/history?page=2&type=all&label=bioconda-utils-test, this seems to have been working as recently as yesterday. Maybe there's something funky going on with the servers today.

PR #128, opened 2017-06-07
[WIP] Handle noarch when on travis.

johanneskoester:
Recipes that have `build:noarch` defined will only be built on linux.
Before this can be merged, PR #130 has to be merged.


johanneskoester:
@bioconda/core from now on, we can use noarch in our recipes. When reviewing, please instruct people to use it whenever possible (in order to save space). It would be nice to have updated docs and maybe a lint that fires in case of python and no gcc dependency or metapackages. Unfortunately I don't have the time for that now.

PR #130, opened 2017-06-07
Parse metadata two times in order to ensure that build number is set …

johanneskoester:
…before determining output file path. Upgrade to conda-build 2.1.15.

@bgruening @daler this should fix our issues with the latest conda-build. I have tested this locally with real recipes. I suggest that we use this branch in bulk and see if it works. Once we are sure, we can merge into master.


daler:
Ah, so this was user-error when using the conda-build API in tests? Good news!

johanneskoester:
Kind of. It could of course be a bit more intuitive. How should one know that it returns an incorrect string in that case? Anyway, I am merging this branch into kill_r now, so that we can test it in the bulk branch.

PR #131, opened 2017-06-07
metadata parsing

bgruening:
@johanneskoester this seems to work for me with the lastest version,


bgruening:
At least on the abyys recipe. Will test more in the evening ...

johanneskoester:
I found the same, see PR #130

johanneskoester:
I also merged into kill_r in order to test on bulk.

PR #133, opened 2017-06-12
Docs update

daler:
@bgruening anything you want to add to biocontainers info?


bgruening:
Looks good to me! I wondering if `platform-depdendent compilation` is too confusing and `compilation` is enough. Essentially, if we `gcc` (or similar) is defined we don't want to have `noarch`, correct?

bgruening:
Thanks @daler!

PR #134, opened 2017-06-12
add linter for noarch

daler:
@johanneskoester @bgruening anything else you can think of to look for in the linter?

### DIFF
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -211,7 +211,7 @@ def should_be_noarch(recipe, meta, df):
         ('python' in deps) and
         not _has_preprocessing_selector(meta)
     ) and (
-        'noarch' not in meta['build']
+        'noarch' not in meta.get('build', {})
     ):
         return {
             'should_be_noarch': True,
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -1,5 +1,6 @@
 import os
 import glob
+import re
 
 
 def _subset_df(recipe, meta, df):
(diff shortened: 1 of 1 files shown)


bioconda_utils/lint_functions.py
```diff
@@ -185,6 +203,22 @@ def has_windows_bat_file(recipe, meta, df):
...
 
+
+def should_be_noarch(recipe, meta, df):
+    deps = _get_deps(meta)
+    if (
+        ('gcc' not in deps) and
+        ('python' in deps) and
+        not _has_preprocessing_selector(meta)
+    ) and (
+        'noarch' not in meta['build']
```
johanneskoester: Does above line also work if meta has no build section?

bioconda_utils/lint_functions.py
```diff
@@ -185,6 +203,22 @@ def has_windows_bat_file(recipe, meta, df):
             'fix': 'remove windows .bat files'
         }
 
+
+def should_be_noarch(recipe, meta, df):
+    deps = _get_deps(meta)
+    if (
+        ('gcc' not in deps) and
+        ('python' in deps) and
+        not _has_preprocessing_selector(meta)
```
johanneskoester: What is the point of the preprocessing selector? Is it just because we cannot be sure in that case?

bioconda_utils/lint_functions.py
daler: thanks, fixed

bioconda_utils/lint_functions.py
daler: Right. I was thinking if there was something strange enough in the recipe that it needs preprocessing selectors then it's not a good candidate for noarch.

johanneskoester: Sounds reasonable.

bgruening: I'm worried that this disables the linter for too many packages. I think there are a few packages with pre-processing selectors in the `url` section or test section - still we would like to have noarch packages for it. Just a feeling, not sure how often these cases are.

bgruening:
Thanks @daler!

bioconda_utils/lint_functions.py
daler: Sorry didn't get a chance to reply before -- If there's a pre-processing selector in the url or test section, doesn't that mean it's arch-specific?

Either way I think it's ok at least for now to let some packages slide until we get a better feel for what works with noarch, and then tighten up the linter later.

PR #139, opened 2017-06-21
Fix noarch errors. Fix sub-recipe problems.

johanneskoester:
1. Fix noarch building (use correct output folder in the container) (see PR bioconda/bioconda-recipes#5014).
2. Keep conda from building subdirectories by explicitly specifying meta.yaml (see PR bioconda/bioconda-recipes#5011).

### DIFF
--- a/bioconda_utils/docker_utils.py
+++ b/bioconda_utils/docker_utils.py
@@ -103,7 +103,7 @@ conda config --add channels file://{self.container_staging}  > /dev/null 2>&1
 conda build {self.conda_build_args} {self.container_recipe}/meta.yaml 2>&1
 
 # Identify the output package
-OUTPUT=$(conda build {self.container_recipe} --output 2> /dev/null)
+OUTPUT=$(conda build {self.container_recipe}/meta.yaml --output 2> /dev/null)
 
 # Some args to conda-build make it run and exit 0 without creating a package
 # (e.g., -h or --skip-existing), so check to see if there's anything to copy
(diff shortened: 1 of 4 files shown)


bioconda_utils/docker_utils.py
```diff
@@ -91,13 +91,16 @@
...
 
-# The actual building....
-conda build {self.conda_build_args} {self.container_recipe} 2>&1
+# The actual building...
+# we explicitly point to the meta.yaml, in order to keep
+# conda-build from building all subdirectories
+conda build {self.conda_build_args} {self.container_recipe}/meta.yaml 2>&1
 
 # Identify the output package
 OUTPUT=$(conda build {self.container_recipe} --output 2> /dev/null)
```
bgruening: adding meta.yaml here as well?

johanneskoester: Good catch, thanks!

johanneskoester:
Aparently, noarch also generalizes over the python version... There is no Python pinning if `noarch: python` is used. That means, that packages using noarch also have to specify a version constraint on the python dependency instead of using build: skip. Finally, it does not seem to work reliably if you have an entry point, because that contains a hardcoded python version, which is e.g. not updated when you update python after the package has been installed.

johanneskoester:
There are plans to improve this:
https://github.com/conda/conda-build/milestone/3

johanneskoester:
The main issue discussion the limitation of the current noarch implementation is here: https://github.com/conda/conda-build/issues/542

johanneskoester:
@daler, why was the noarch lint disabled? Was it because noarch builds did not work (which this PR hopefully fixes)?

bgruening:
@johanneskoester yes it was disabled because it caused trouble. Hopefully this will fix it.

johanneskoester:
Ok, lets wait with the lint until the first real recipe was build successfully with noarch. Snakemake is not a candidate unfortunately. If you have one, please try once this is merged.

daler:
Awesome, thanks Johannes!

PR #141, opened 2017-06-23
Whitelist MACOSX_DEPLOYMENT_TARGET

johanneskoester:
Such that we can set it in our global travis configuration.


johanneskoester:
xref: #4997

PR #143, opened 2017-06-24
try passing proxies to docker build

daler:
xref https://github.com/bioconda/bioconda-recipes/issues/5027 and https://blog.codeship.com/using-docker-behind-a-proxy


daler:
Confirmed fixed in bioconda/bioconda-recipes#5027

vinjana:
Hi Ryan, I noticed that older versions (1.12.6) of `docker build` do not understand the `--network` option!

daler:
Hmm. Options:

- require later docker verions
- only add `--network` for later versions
- is there an alternative argument that works with a proxy on docker 1.12.6?

vinjana:
The first version with the feature is [1.13.0](https://github.com/docker/docker-ce/blob/4302a48fed6f488be2a53fb4bb1eec52322cc173/components/engine/CHANGELOG.md). At least CentOS 7 still has 1.12.6 per default (which is how I found out about this). I don't see any other alternative options for this feature in 1.12.6, so, if we add the feature only for 1.13.0+ we are on the safe side.

As a side remark: While this fix in bioconda-utils helped me to run the normal simulate_travis tests in docker, I did not succeed with the mulled-build/involucro steps (and can therefore still not reproduce the problems I have on Travis). So if the motivation is to have this feature working for the full simulate_travis.py, the http_proxy stuff needs to be enabled in mulled-build and involucro as well. I actually gave up (for now) and managed to organised access to one of the rare machines on our institute that does not need to go through the proxy (after some detour trying to set up a transparent, cascading squid proxy).

daler:
Thanks for the update. @bgruening can you comment on what it would take to get mulled-build to use proxy env vars?

PR #145, opened 2017-06-29
Update docs for newbie developers (+favicon)

olgabot:
Here's some updates to the documentation for contributing a recipe for the first time, specifically with installing `conda-build`, `bioconda-utils`, and that the commands must be run from the root `conda` environment.


daler:
@olgabot thanks for these much-needed updates.

Rather than install `bioconda-utils` directly, running `simulate-travis.py --install-requirements` in the `bioconda-recipes` dir will install the branch currently being used by `bioconda-recipes`. It's a way to more closely match what's being done on travis-ci. But to directly answer your question, https://github.com/bioconda/bioconda-utils/blob/3993132aee6a6bee8d88763311ec10a781b7cb70/bioconda_utils/bioconda_utils-requirements.txt#L19

daler:
The favicon is great.

Unfortunately since conda-build has to run in the root env, everything else has to be there as well. I know it's a pain but there's no good way around it that I know of besides spinning up a VM. Even using a docker container doesn't work well because of the difficulties of running docker containers inside a docker container.

I end up using my root conda env for all bioconda stuff, and separate envs for anything else.

daler:
@bgruening do you have a good way of isolating the bioconda build env? Maybe a separate conda installation?

daler:
Ah, nice! I'll have to test that out. Maybe we could have `simulate-travis.py` install a separate miniconda installation separate from the main conda installation and use that for everything. That would be really convenient.

daler:
@olgabot could you please merge master into this? Recent updates should fix the failing test.

daler:
Thanks @olgabot!

PR #146, opened 2017-07-04
fix test to handle cases where GITHUB_TOKEN is already set

daler:
@nsoranzo turns out conda-build handles rendering differently based on whether the env var is already set or not. This PR makes the test handle both cases. I haven't worked out exactly why the error is different depending on the situation, but in both cases the env is properly sandboxed.


nsoranzo:
Thanks @daler!

PR #148, opened 2017-07-07
add linting for setuptools args when setuptools is build-dep

daler:
Still WIP because I'd like to add some linting tests to this before merging (due to the problems we had before after merging the noarch lints)

xref https://github.com/bioconda/bioconda-recipes/pull/5124#issuecomment-313716330


johanneskoester:
Great, thanks!

bioconda_utils/lint_functions.py
```diff
@@ -236,6 +236,34 @@ def should_not_be_noarch(recipe, meta, df):
...
+        'setup.py install' in script_line and
+        '--single-version-externally-managed' not in script_line
+    ):
+        return err
+
+    build_sh = os.path.join(recipe, 'build.sh')
+    if not os.path.exists(build_sh):
+        return
+
+    for line in open(build_sh):
```
bgruening: I would read the entire file here at once. We don't know if setup.py and --single-version... is on the same line.

daler: Good point; fixed.

PR #149, opened 2017-07-12
update our numbers for ISMB/ECCB

bgruening:


daler:
Pretty linear!

figure

daler:
For future reference:


```python
"""
Plots the results of::

    git log --pretty=format:%h\t%aN\t%aI" --name-only recipes/* > LOG

"""

import seaborn as sns
import pandas
import os
from matplotlib import pyplot as plt
import datetime

sns.set_style('white')

class chunk(object):
    def __init__(self, block):
        commit, author, time = block[0].split('\t')
        self.author = author
        self.time = datetime.datetime.strptime(time.split('T')[0], "%Y-%m-%d")
        self._block = block
        self.recipes = self._parse_recipes(block[1:])

    def _parse_recipes(self, block):
        recipes = []
        for i in block:
            if not i.startswith('recipes/'):
                continue
            if os.path.basename(i) != 'meta.yaml':
                continue
            recipes.append(os.path.dirname(i.replace('recipes/', '')))
        return set(recipes)

def gen():
    lines = []
    for line in open('LOG'):
        line = line.strip()
        if len(line) == 0:
            yield chunk(lines)
            lines = []
            continue
        lines.append(line.strip())
    yield chunk(lines)

dfs = []
cumulative_recipes = set()
cumulative_authors = set()
for i in sorted(gen(), key=lambda x: x.time):
    if len(i.recipes) == 0:
        continue

    unique_recipes = i.recipes.difference(cumulative_recipes)
    if len(unique_recipes) > 0:
        dfs.append(
            {'time': i.time,
             'author': i.author,
             'recipes': unique_recipes,
             'nadded': len(unique_recipes),
             'new_author': i.author not in cumulative_authors
            },
        )
    cumulative_recipes.update(i.recipes)
    cumulative_authors.update([i.author])

df = pandas.DataFrame(dfs)
df['cumulative_authors'] = df.new_author.astype(int).cumsum()
df['cumulative_recipes'] = df.nadded.cumsum()
fig = plt.figure(figsize=(6, 8))
ax1 = fig.add_subplot(2,1,1)
ax2 = fig.add_subplot(2,1,2)
df.plot('time', 'cumulative_authors', ax=ax1)
df.plot('time', 'cumulative_recipes', ax=ax2)
ax1.set_ylabel('Number of unique authors')
ax2.set_ylabel('Number of unique recipes')
fig.tight_layout()
plt.show()
```

If outputting the log didn't take so long, it would be cool to autogenerate this in the docs.

Issue #150, opened 2017-07-13
Skip lints in extra section of recipe


johanneskoester:
It would be nice if we could persist lint skipping in the recipe, via the extra section. E.g.:

```
extra:
  skip-lints:
      - uses_git_url
      - missing_hash
```

@daler, what do you think?

daler:
Sure, that's a great idea.

PR #151, opened 2017-07-13
Skip lints in extra

daler:
addresses #150, sorry @johanneskoester this is a little late for all your `vg` recipe commits!


johanneskoester:
That was really fast! I appreciate it!

PR #152, opened 2017-07-14
Should fix #5174

simonvh:
If I'm correct this should fix bioconda/bioconda-recipes#5174 , see also bioconda/bioconda-recipes#5131.


daler:
Thanks @simonvh, nice catch!

PR #158, opened 2017-07-28
Update install requirements to include pygithub.

cbrueffer:
Without this bioconda-utils has unmet requirements when installed according to the README and in a fresh Anaconda installation, and a subsequent simulate-travis.py run errors out.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
             ],
         )
     ],
-    install_requires=["argh", "networkx", "pydotplus", "pyyaml", "requests", "jsonschema", "pandas", "colorlog", "jinja2"],
+    install_requires=["argh", "networkx", "pydotplus", "pyyaml", "requests", "jsonschema", "pandas", "colorlog", "jinja2", "pygithub"],
     entry_points={"console_scripts": [
         "bioconda-utils = bioconda_utils.cli:main",
         "bioconductor_skeleton = bioconda_utils.bioconductor_skeleton:main"


cbrueffer:
Steps do reproduce:
1. Install Anaconda 4.4.0
2. Set up bioconda channels
3. conda update --all
4. conda install conda-build
5. pip install git+https://github.com/bioconda/bioconda-utils.git
6. run simulate-travis.py

daler:
Thanks. The core issue here is that the README needs to be updated!

There are actually other dependencies also not listed in setup.py, as we use the `./simulate-travis.py --install-requirements` (see [bioconda contributor instructions](https://bioconda.github.io/contribute-a-recipe.html#update-repo-and-requirements)) install these packages from conda using [this requirements.txt](https://bioconda.github.io/contribute-a-recipe.html#update-repo-and-requirements).

We're probably missing deps in setup.py because in practice everything is already installed via conda, and we haven't been maintaining the list in setup.py. In fact, I think the better option is actually to remove the `install_requires=` line to avoid any ambiguity, and importantly update the README to indicate either `simulate-travis.py` should be used, or to install deps via `conda install --file bioconda_utils/bioconda_utils-requirements.txt`.

cbrueffer:
Thanks for the feedback! Together with the revamped simulate-travis.py workflow you posted either suggested option sounds good to me.

daler:
Thanks @cbrueffer, closing this now (replaced by #160 and #156)

PR #167, opened 2017-08-18
Print build/test output to log in real time

epruesse:
This patch restores the "log output in real time" behavior of plain unwrapped `conda build`, maintaining the log message format and the explicit `decode` call.

Seeing immediately what is going on, and having STDERR and STDOUT interleaved _without_ buffering helps immensely when writing recipes:
 - It will often reduce the length of the development cycle.
 - Knowing that "yes, it is still compiling" at a glance helps against the frustration of "waiting". 
 - Inspecting the output of build and test stages is useful even if they succeeded. (Actually, success of the build stage is ultimately determined by the test, so output should have been dumped on failed tests also). 
 - The buffered interleaving resulted in a peculiar mixup of blocks of STDOUT and STDERR, making it impossible to determine the chronological order in which each line of output was printed.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -134,7 +134,7 @@ def build(
                       [os.path.join(recipe, 'meta.yaml')]
                 logger.debug('command: %s', cmd)
                 with utils.Progress():
-                    p = utils.run(cmd, env=os.environ)
+                    utils.logging_run(cmd, logger, env=os.environ)
 
             build_success = True
 
@@ -161,7 +161,7 @@ def build(
     base_image = 'bioconda/extended-base-image' if use_base_image else None
 
     try:
-        res = pkg_test.test_package(pkg_path, base_image=base_image)
+        pkg_test.test_package(pkg_path, base_image=base_image)
 
         logger.info("TEST SUCCESS %s, %s", recipe, utils.envstr(_env))
         mulled_image = pkg_test.get_image_name(pkg_path)
--- a/bioconda_utils/utils.py
+++ b/bioconda_utils/utils.py
@@ -12,6 +12,7 @@ from collections import defaultdict, Iterable
 from itertools import product, chain
 import logging
 import pkg_resources
+import queue
 import networkx as nx
 import requests
 from jsonschema import validate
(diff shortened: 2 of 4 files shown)


bgruening:
Cool! This is a very nice addition. Can you look at the failing test, or should I do?

epruesse:
@bgruening  It's the `build_docs.sh` that's complaining. Looks like it's missing the encryption key in the environment, so I probably lack the rights to fix it myself.

From log:
```
./build-docs.sh: line 63: !ENCRYPTED_KEY_VAR: unbound variable
```

build-docs.sh:
```
60: # Decrypt and ssh-add key.
61: ENCRYPTED_KEY_VAR="encrypted_${ENCRYPTION_LABEL}_key"
62: ENCRYPTED_IV_VAR="encrypted_${ENCRYPTION_LABEL}_iv"
63: ENCRYPTED_KEY=${!ENCRYPTED_KEY_VAR}  
64: ENCRYPTED_IV=${!ENCRYPTED_IV_VAR}
```

daler:
@epruesse, this is really nice work. Could you add a flag for enabling/disabling? The reason is that when logs get too big travis-ci doesn't even show them and you have to download the raw log. Then we lose the ability to link directly to a particular line of the log, which has been useful in the past. 

Maybe we can run this verbose mode locally, and by default run the quiet[er] mode on travis unless verbose mode is explicitly asked for, likely by a temporary change in the args listed in `.travis.yml`.

The current failure is due to running on a fork, which doesn't have the encrypted var available when cloning the repo over ssh. Should be a quick fix to use https when not on the main repo, I will fix this tomorrow.

epruesse:
@daler: The flag is (sort of) already there. `logging_print`  will optionally take a `logger` object and a `level` flag, so that the logger name and level can be adjusted from the code executing a binary and so that `--logging-level` adjusts the verbosity as usual. 

The question is just which things should log at which level, and which levels are appropriate for local, PR and merge builds.

That said, having build packages where the full output is too long for Travis, I still prefer having the log in full. Travis also goes by bytes, BTW, not lines, so something "unexpanding" `$PREFIX` actually helps a lot (did that for the `arb-bio` package). If the `meta.yaml` was easier to get to, I'd say leave it to the package to decide and put a switch in the `extra` section. Anyway, I think the packages with tons of output tend to be the complicated ones where the maintainer/author would know how to shrink the output in the `build.sh`.

One situation were the full output is actually annoying are the bulk package PRs. So perhaps changing the log level depending on the number of packages build would be appropriate.

The other challenge would be having everything if the tests fail. In that case, there is often enough something wrong with the build itself, but the logs are hidden, complicating debugging. It'd take quite a bit of hacking to "stash" the log and print it later though.

Anyway, I can make "silence" the default if that's your preference. (As long as I can get it at all I'm happy... got into this in the first place because it was slowing me down too much to have no log).

daler:
Good points. @bioconda/core, any preferences on how verbose logs should be (and where)?

bgruening:
Hard to tell, I guess the decision is more a practical, as much as possible but not to much to make travis angry :)

That said, I know its not a useful answer so lets just try a few one and adjust as needed?

johanneskoester:
So, the previous behavior (hidden logs, only shown on failure, regularly print a dot such that travis is happy) was tuned carefully to also work with our bulk updates, where you really don't care about log output of recipes that build fine. So, I would say that this PR definitely should add a command line flag that allows to enable the previous behavior.

bioconda_utils/build.py
```diff
@@ -134,7 +134,7 @@ def build(
                       [os.path.join(recipe, 'meta.yaml')]
                 logger.debug('command: %s', cmd)
                 with utils.Progress():
-                    p = utils.run(cmd, env=os.environ)
+                    utils.logging_run(cmd, logger, env=os.environ)
```
johanneskoester: Why changing the function name? Why the logger argument? We have a global logger object.

bioconda_utils/utils.py
```diff
@@ -176,6 +177,51 @@ def temp_os(platform):
...
+    UnicodeDecodeErrors.
+
+    Raises subprocess.CalledProcessError if the exit code is non-zero.
+
+    Returns the subprocess.CompletedProcess object.
+    """
+    if logger is None:
+        logger = logging.getLogger(__name__)
+
+    p = sp.Popen(cmds, stdout=sp.PIPE, stderr=sp.PIPE,
```
johanneskoester: Why can't we just use the original `sp.run` and omit the pipe arguments so that all output is immediately passed to the terminal?

bioconda_utils/build.py
epruesse: > Why changing the function name?
I chose to add another function rather than replace the existing one to keep the changes limited to the places where I inspected the surroundings of the calling code. When running an external binary, the case where you want stdout/stderr directly pushed to the logger are overall the exception.

That said, it would be easy enough to reduce this to one function. 

> Why the logger argument? We have a global logger object.
I gave the `logging_run` function the option to pass a logger object because those lines originate from a binary specified by the caller, and not from the `util` module itself. For example, shell output marked as originating from `pkg_test` or `build` is more helpful than all coming from `util`. 

You are right in that it does not have any visible effect here, though. While the logger objects are per module as usual (not global), bioconda-utils has a custom configured StreamHandler that doesn't print the `name` of the log records. => no effect at this time.

bioconda_utils/utils.py
epruesse: Good question. I assumed that was undesired because it wasn't done in the first place.

epruesse: Hmm. As you said above, there is good reason to have the output adjusted to the scenario. Bulk builds should have very terse output. New recipes should have reasonable output. Local builds probably want it all. Passing it all though logging allows controlling the verbosity via the log level, rather than switches distributed over the code.

epruesse:
Closing in favor of rewrite in #522

epruesse:
(Please have a look at the implementation over there. It now allows limiting the number of lines emitted by each call to `run.sp` via the command line log settings, as well as logging to a file)

PR #169, opened 2017-08-24
Bioconductor skeleton updates

daler:
This PR moves the `bioconductor_skeleton.py` script from bioconda-recipes to bioconda-utils. It is now called with `bioconda-utils bioconductor-skeleton`. 

Importantly, it also now handles data-heavy packages by building a `post-link.sh` script instead of a `build.sh` script. This is the strategy tested out in https://github.com/bioconda/bioconda-recipes/pull/5485. The post-link script is executed at install-time, not build-time, which means the package data is not included in the conda package. This avoids consuming the limited space on the bioconda channel while still allowing large, data-heavy Bioconductor packages to be installed via conda.

Data packages are detected as those stored under the "data/annotation" subdirectory on the Bioconductor servers (e.g., http://bioconductor.org/packages/3.5/data/annotation/src/contrib/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0.tar.gz). This is automatically detected, so when building recipes you won't have to know or care ahead of time whether a package should use the post-link strategy.

xref https://github.com/bioconda/bioconda-recipes/pull/5428

### DIFF
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -1,0 +1,640 @@
+#!/usr/bin/env python
+
+import shutil
+import tempfile
+import configparser
+from textwrap import dedent
+import tarfile
+import pyaml
+import hashlib
+import os
+import re
+import bs4
+import urllib
+from urllib import request
+from urllib import parse
+from urllib import error
+from collections import OrderedDict
+import logging
+import requests
+from colorlog import ColoredFormatter
+
+logging.getLogger("requests").setLevel(logging.WARNING)
+
+log_stream_handler = logging.StreamHandler()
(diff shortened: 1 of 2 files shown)


johanneskoester:
This is simply amazing @daler! Yet another interesting story for our paper ;-).

daler:
Thanks! I really needed those packages, glad we have a solution now. Hopefully this strategy will hold up to some real-world testing.

bgruening:
I'm looking forward to some real work testing. I'm a little bit surprised that this small idea went through to smoothly I would have expected that the containers scream at us with running out of disc space :)

Thanks for implementing this! You rock!

jerowe:
You are an infrastucture god, @daler. ;-)

PR #175, opened 2017-08-31
Recommend sha256 over md5

peterjc:
Should this label ``shasum -a 256 ...`` as typical on OSX? Is ``sha256sum`` the more common tool on Linux (I have both)?

### DIFF
--- a/docs/source/guidelines.rst
+++ b/docs/source/guidelines.rst
@@ -6,7 +6,7 @@ Guidelines for ``bioconda`` recipes
 bioconda recipe checklist
 -------------------------
 - Source URL is stable (:ref:`details <stable-url>`)
-- md5 or sha256 hash included for source download (:ref:`details <hashes>`)
+- sha256 or md5 hash included for source download (:ref:`details <hashes>`)
 - Appropriate build number (:ref:`details <buildnum>`)
 - `.bat` file for Windows removed (:ref:`details <bat-files>`)
 - Remove unnecessary comments (:ref:`details <comments-in-meta>`)
@@ -37,12 +37,19 @@ TODO: additional info on the various R and bioconductor URLs
 
 Hashes
 ~~~~~~
-Use `md5sum` (Linux) or `md5` (OSX) on a file to compute its md5 hash, and copy
-this into the recipe. A quick way of doing this is:
+We support either sha256 or md5 checksums to verify the integrity of the source
+package. If a checksum is provided alongside the source package, then use that.
+Otherwise we prefer sha256 over md5.
+
+Use ``shasum -a 256 ...`` (or ``sha256sum``  or ``openssl sha256`` etc) on a
+file to compute its sha256 hash, and copy this into the recipe, for example:
 
 .. code-block:: bash
 
-    wget -O- $URL | md5sum
+    wget -O- $URL | shasum -a 256
+
+Likewise use `md5sum` (Linux) or `md5` (OSX) on a file to compute its md5 hash,
+and copy this into the recipe.
 
 .. _buildnum:
 


mbargull:
Maybe alter the `wget` example to `wget -O- $URL | sha256sum` and move it up?

mbargull:
We should clarify that checksums provides alongside the source URL should be used preferably.

peterjc:
I'm happy to update this with an example, but should I use ``sha256sum ...`` or ``shasum -a 256 ...`` which seem to be most common on Linux and Mac respectively?

peterjc:
Rebased, updated to say explicitly to prefer any source provided checksum, and remove the example (since it is wasteful to download the file again just for the checksum), which solves the question of which tool to show for the sha256 checksum.

svenrahmann:
It would still be helpful to give a typical tool and command line call on Linux/Mac. Using wget in the example is indeed wasteful, since people will usually have the file on disk.

peterjc:
Can you comment on which of ``shasum -a 256 ...`` and ``sha256 ...` are most common on Linux and macOS respectively? I failed to get a conclusive answer when I tried to work this out.

nickp60:
Not an expert here, but I tend to use `shasum -a 256` as it seems to be the same across OSX and (at least Ubuntu) Linux.  My mac didn't have sha256sum...

mbargull:
It looks like `shasum -a 256` is common to both platforms. I suppose `sha256sum` will likely be available on Linux. If I understand correctly, on Mac there is no sha256 checksum provided right away but installed separately or with other software. All in all I found
- Linux/Mac: `shasum -a 256`, `openssl sha56`
- Linux: `sha256sum`
- Mac: `gsha256sum`

mbargull:
Would it be sensible to use `shasum -a 256` and give alternatives in a footnote or parentheses, e.g., 

>alternatives: `openssl sha256` or `sha256sum` (Linux) or `gsha256sum` (Mac)

I don't think we should go into details how to obtain (if not already available) any of those. We should try not to clutter up the recommendations with excessive details but give useful examples and the like.

mbargull:
I'd vote to re-include the `wget` example. Since we are talking about calculating the checksum for a file given by an URL, it should be least error-prone to directly calculate it from a fresh download. Concerning being "wasteful": given the fast internet connections nowadays, it might be less wasteful if it leads to

peterjc:
If you leave out the ``wget``, then the examples are trivial and I don't think it is worth padding the documentation further to add:

```
.. code-block:: bash

   sha246sum package.tar.gz
```
or,

```
.. code-block:: bash

   shasum -a 256 package.tar.gz
```

etc.

peterjc:
Confirming on macOS,

```
$ which sha256sum
(i.e. not installed)
$ which shasum
/usr/bin/shasum
```

while on CentOS I have installed both, but it looks like ``sha256sum`` is more likely to be there by default?

```
$ which sha256sum
/usr/bin/sha256sum
$ which shasum
/usr/local/bin/shasum
```

docs/source/guidelines.rst
```diff
@@ -37,12 +37,19 @@ TODO: additional info on the various R and bioconductor URLs
 
 Hashes
 ~~~~~~
-Use `md5sum` (Linux) or `md5` (OSX) on a file to compute its md5 hash, and copy
-this into the recipe. A quick way of doing this is:
+We support either sha256 or md5 checksums to verify the integrity of the source
+package. If a checksum is provided alongside the source package (e.g. in the
+release notes or as a sister file), use that. Otherwise we prefer sha256 over md5.
+
+Use ``shasum -a 256 ...`` (or ``sha256sum``  or ``openssl sha56 ...`` etc) on a
```
mbargull: typo: `openssl sha56` => `openssl sha256`

docs/source/guidelines.rst
```diff
@@ -37,12 +37,19 @@ TODO: additional info on the various R and bioconductor URLs
 
 Hashes
 ~~~~~~
-Use `md5sum` (Linux) or `md5` (OSX) on a file to compute its md5 hash, and copy
-this into the recipe. A quick way of doing this is:
+We support either sha256 or md5 checksums to verify the integrity of the source
+package. If a checksum is provided alongside the source package (e.g. in the
+release notes or as a sister file), use that. Otherwise we prefer sha256 over md5.
```
mbargull: I'd leave out 
> (e.g. in the release notes or as a sister file)

since

> alongside the source package

should suffice.

docs/source/guidelines.rst
mbargull: Maybe use `shasum -a 256 $FILE` instead of `shasum -a 256 ...` to be consistent with the `wget` example's style.
For `openssl sha256` the `...` can be omitted.

mbargull:
If `shasum` is more likely available on macOS and usually (albeit not necessarily) available on Linux, then we can go with that, just as you did.
Since you provides the alternatives, people will easily figure out to use `sha256` if `shasum -a 256` didn't work. Hence (IMHO) we needn't dwell on whether to use one or the other in the examples.

peterjc:
Updated with @mbargull's feedback.

peterjc:
Rebased and squashed into one commit (since GitHub wanted this updated)

mbargull:
Those are helpful changes, thanks @peterjc!

@daler: any objections?

daler:
Looks good, thanks all!

Issue #163, opened 2017-08-14
document quoting of tests


daler:
in https://github.com/bioconda/bioconda-recipes/pull/5515, mulled-build was failing because of a test in `meta.yaml` using single rather than double quotes.

peterjc:
I ran into this before, see https://github.com/bioconda/bioconda-recipes/pull/4799#issuecomment-304683233 where I have to replace single quotes with double quotes

johanneskoester:
@bgruening is this a bug in mulled? Maybe quotes are not escaped properly?

bgruening:
Its a "bug" in our parsing of the commandline for tests. What we are doing here is to extract the tests and pass them via commandline to the container builds ... which can cause troubles with quotes. We need to be smarter here ... I will look at it again as soon as I can.

epruesse:
`conda build` runs the tests by writing the lines into `run_test.sh` (or appending to that file if it exists). If #107 gets implemented, that should be trivial.

Needs someone familiar with the mulled parts of the system though. @bgruening - any chance you can have a look at this?

PR #177, opened 2017-09-13
Add malformed tests check

jtratner:
Currently, if you have a single quote in your test command, it'll break the bioconda build because of shell escaping.  Not sure if the upstream will fix this (or if a fix is really worth it, given that you can always put your test in a test file), but the error is confusing to the user.  This PR adds a single_quote_in_tests check to the linter.

Aside - I haven't been able to get the test suite to run yet (not sure how to let anaconda for python3 and python2 exist).  But I'm hoping Travis will run 'em for me :)

### DIFF
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -143,6 +143,16 @@ def missing_tests(recipe, meta, df):
             }
 
 
+def malformed_tests(recipe, meta, df):
+    if 'test' in meta:
+        # will cause mulled to fail
+        if any("'" in line for line in meta['test']):
+            return {
+                'single_quote_in_tests': True,
+                'fix': 'use double quotes or move to separate test file'
+            }
+
+
 def missing_hash(recipe, meta, df):
     # could be a meta-package if no source section or if None
     try:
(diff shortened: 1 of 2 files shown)


bgruening:
Cool thanks @jtratner. xref: https://github.com/bioconda/bioconda-utils/issues/163
I think the fix might be easier if we can get this to quote the parameters correctly: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/pkg_test.py#L134

jtratner:
@bgruening 
> I think the fix might be easier if we can get this to quote the parameters correctly: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/pkg_test.py#L134

yes, but also shell escaping gets so complicated that it makes my head hurt :-/ (so I picked the dumb way ;P)

bgruening:
That is perfectly fine for the time being! Thanks!

jtratner:
Test failures are unrelated... especially the ENCRYPTED_KEY_VAR for docs - wdyt @bgruening ?

eseiler:
Could the ENCRYPTED_KEY_VAR error be related to #167 ?

> The current failure is due to running on a fork, which doesn't have the encrypted var available when cloning the repo over ssh. Should be a quick fix to use https when not on the main repo, I will fix this tomorrow.

bgruening:
Looks still good to me. Could we have a rebase?

epruesse:
@bgruening - merged and conflicts resolved. It looks like it might be overzealous, but it is KISS.

PR #180, opened 2017-09-21
add six

bgruening:


bgruening:
I tested this locally and on bioconda-recipes in bulk, it at lest does not work more things as currently broken ;)

I will merge this even if not all tests have passed now. The issue is that noarch packages have arrived in conda-forge and they seem to break things...

daler:
Thanks Bjoern -- I'm guessing this problem will affect a whole lot more than just bioconda-utils?

bgruening:
I reported it upstream and fix is underway ... sorry in a Galaxy workshop the entire week.

PR #184, opened 2017-09-30
add dockerfile and setup script

daler:
Now that requirements and versions have settled down, there's not really a need for building a fresh docker container on *every* `./simulate-travis.py` run. This PR adds a Dockerfile and a `setup-channels.sh` that is used by it. This script can double as a `curl | bash` approach for quickly setting up channels in other contexts.

Once this PR is merged and a docker image is built, `bioconda_utils/docker_utils.py` can be updated to use it, removing the need to install requirements and saving considerable time when testing locally. We can still keep the proxies part though, to still support custom building if needed.

@johanneskoester, when you get a chance would you please connect this Dockerfile to dockerhub for automated builds?


johanneskoester:
Nice!

johanneskoester:
@daler @bgruening this is now connected to dockerhub://bioconda/bioconda-utils-build-env. @bgruening does this have implications for your recent openssh work?

bgruening:
I guess so, we need to include the openssh package in the Dockerfile as well.

daler:
added in https://github.com/bioconda/bioconda-utils/pull/218

daler:
Thanks @johanneskoester!

PR #189, opened 2017-10-11
lint: fix: crash at empty build section

mbargull:
Linting for recipes with empty `build:` sections failed since `meta.get('build', {})` is `None` in that case and thus does not support any `dict`-operations. Testcase and fix added.
Also added a debug log for every recipe before executing the lint functions to help debugging batch linting. @johanneskoester


johanneskoester:
@mbargull, I think the tests fail because our docs building only works in the main repo, not a fork. Since, all other tests pass, I will merge this and ignore the failure.

Issue #190, opened 2017-10-12
pin linting considers patch files packages (regression by #188)


epruesse:
Since the merge of #188, patches beginning with the name of a pinned package fail linting:

```yaml
source:
  patches:
    - boost_106400.patch
```

(Kudos to @nsoranzo for finding the cause)

nsoranzo:
xref https://github.com/bioconda/bioconda-recipes/pull/6073
Failed build: https://travis-ci.org/bioconda/bioconda-recipes/builds/287662782

johanneskoester:
Working on this in PR #191

nsoranzo:
Linting passes now, thanks @johanneskoester!

PR #191, opened 2017-10-13
Add tests for pin linting, fix issue #190.

johanneskoester:


bioconda_utils/lint_functions.py
```diff
@@ -288,17 +288,30 @@ def _pin(env_var, dep_name):
...
+        in_requirements = False
+        in_run = False
         for line in open(os.path.join(recipe, 'meta.yaml')):
+            if line.startswith("requirements:"):
+                in_requirements = True
+            elif in_requirements and line.strip().startswith("run:"):
+                in_run = True
+            elif in_requirements and line.strip().startswith("build:"):
+                in_run = False
+            elif not line.startswith(" "):
```
mbargull: Should/could be `elif not line.startswith((" ", "#")):`, right? That way, not indented comments are also handled.

johanneskoester: Good point.

bioconda_utils/lint_functions.py
```diff
@@ -288,17 +288,30 @@ def _pin(env_var, dep_name):
...
+            elif in_requirements and line.strip().startswith("build:"):
+                in_run = False
+            elif not line.startswith(" ") and not line.startswith("#"):
+                in_requirements = False
+                in_run = False
             line = line.strip()
-            if line.startswith('- {}'.format(dep_name)):
-                if env_var not in line: # and '{{' not in line and '}}' not in line:
+            if in_run and line.startswith('- {}'.format(dep_name)):
+                if pin_pattern.search(line):
```
nsoranzo: Shouldn't this line end with ```is None:``` ?

johanneskoester: Yes it should! Thanks!

johanneskoester: That also explains the failing tests.

nsoranzo:
Should the linting be applied to both `run` and `build` requirements or just `run`? Because the tests seem to specify the former while the code implements the latter.

johanneskoester:
In fact it is a bit more complicated. If the dependency does not occur in run, we don't need to pin it at all because it is statically linked. However, my previous implementation did not explicitly check all cases, which I have fixed in the latest commit.

bioconda_utils/lint_functions.py
```diff
@@ -333,7 +357,5 @@ def pin(recipe, meta, df):
     _pin('CONDA_HDF5', 'hdf5'),
     _pin('CONDA_NCURSES', 'ncurses'),
     _pin('CONDA_HTSLIB', 'htslib'),
-
-    # Pinning in env_matrix.yml is not consistent for bzip2.
-    # _pin('CONDA_BZIP2', 'bzip2'),
+    _pin('CONDA_BZIP2', 'bzip2'),
```
nsoranzo: I think that need to be special-cased in the code above. Maybe also a specific test should be added.

johanneskoester: Why? The wrong entry in the env matrix was fixed in PR bioconda/bioconda-recipes#6310

nsoranzo: Ah, sorry, I missed that PR!

PR #193, opened 2017-10-16
try to fix https://github.com/bioconda/bioconda-utils/issues/190

bgruening:
ping @epruesse @nsoranzo 

I'm still travelling and can not test this. What do you think?

### DIFF
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -293,7 +293,7 @@ def _pin(env_var, dep_name):
         # has jinja templating
         for line in open(os.path.join(recipe, 'meta.yaml')):
             line = line.strip()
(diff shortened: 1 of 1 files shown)


bioconda_utils/lint_functions.py
```diff
@@ -293,7 +293,7 @@ def pin(recipe, meta, df):
         # has jinja templating
         for line in open(os.path.join(recipe, 'meta.yaml')):
             line = line.strip()
-            if line.startswith('- {}'.format(dep_name)):
+            if line == '- {}'.format(dep_name):
```
nsoranzo: Wouldn't it be more robust to get the requirements from the `_get_deps()` function?

johanneskoester:
Including additional test cases.

PR #194, opened 2017-10-16
bioconda-utils location changed

Cashalow:
Update the documentation so that the correct path to bioconda-utils is stated


daler:
@Cashalow docs need a github token which only is available in a branch of the main repo (rather than a fork as this one is). #197 has the same changes.

PR #195, opened 2017-10-16
Migrate to get_output_file_paths with conda-build api

yesimon:
`get_output_file_paths` will be totally removed in v4.


bgruening:
@yesimon we are not using yet cb-3. This will also take some time I assume.

PR #198, opened 2017-10-21
new key for bioconda.github.io

daler:
I had added the github token in #196 to the bioconda-utils settings when in fact it needs to be added to bioconda.github.io settings . . . and github does not allow the same deploy key to be used in different repos. This new encrypted deploy key is now only used for bioconda.github.io.


bgruening:
Cool, let's try that.

PR #199, opened 2017-10-22
build: fix: sections may be empty in get_meta_value 

mbargull:
`meta.yaml` sections may be empty and thus will be an empty `str` when parsed by `conda-build`. Hence accessing `meta[key]` in `get_meta_value` will fail since `meta` can be `""` which cannot be subscripted by another `str`, i.e., section name.

Use case that will fail:
```yaml
extra:
  container:
    extended-base: true  # [py27]
```
Due to the selector, `container` will be empty for Python 3 builds.


daler:
Nice catch, thank you!

PR #200, opened 2017-10-22
lint: fix: reason for skip can be commit msg OR skip-lints section

mbargull:
This just fixes the wording for skipped lints.
Previously `"Commit message defines skip lint test..."` has been logged even if the skip information came from the `skip-lints` section instead of from the commit message.


daler:
Another nice catch, thanks!

PR #201, opened 2017-10-23
lint: fix: build numbers not comparable across channels

mbargull:
Build numbers should only be comparable inside a single conda channel.
This PR removes the build number comparison for `in_other_channels`, which in turn can detect some more duplicate packages with this change.

### DIFF
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -31,13 +31,9 @@ def _subset_df(recipe, meta, df):
     name = meta['package']['name']
     version = meta['package']['version']
 
-    build_number = 0
-    build_section = _get_not_none(meta, 'build')
-    build_number = int(build_section.get('number', 0))
     return df[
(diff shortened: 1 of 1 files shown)


johanneskoester:
So, originally I designed this to compare build numbers as well, because we used this feature to find packages that we can safely delete here in bioconda without impacting the user experience. This is why I also checked the build numbers (because they can be pinned as well, e.g., in environments). Since we now don't have storage constraints anymore, we could also omit this constraint. Semantically, build numbers from different channels will be different anyway. @daler @bgruening opinions?

mbargull:
A reasonable pinning on the build numbers which also considers channels will be possible with conda 4.4.0. Currently, as you said,
> Semantically, build numbers from different channels will be different anyway.

hence this PR.

bgruening:
:+1: for removing this constraint.

daler:
Removing the build number constraint sounds good to me too.

PR #207, opened 2017-11-04
minor tweaks

daler:
1) bioconductor skeleton no longer pins `r-base`
2) warning is logged if a recipe has changed but is still in a blacklist


peterjc:
+1 on warnings if a blacklisted recipe is changed, we've accidentally committed a few updates in that situation (e.g. BLAST package)

PR #209, opened 2017-11-05
hotfix

bgruening:
The changes in https://github.com/bioconda/bioconda-utils/pull/207 breaks current bioconda https://travis-ci.org/bioconda/bioconda-recipes/jobs/297651862

ping @daler


daler:
Ugh, sorry about that and thank you for fixing it, I need to add a test that will check that section.

PR #210, opened 2017-11-06
add `git push origin master` to fork repo update docs

dlaehnemann:
When just running the two steps described in the current docs for the fork workflow, the Travis build fails when pushing out changes on a new or updated branch, based on the newest upstream master -- it gets confused by the lack of commits in the fork's master branch. So whenever updating the local master branch from upstream, these changes should also be pushed out to the fork (origin):

`git push origin master`

I don't think this is necessary for the clone, that should only be on the local machine. But I haven't used that workflow yet, so I'm not absolutely sure.

### DIFF
--- a/docs/source/contribute-a-recipe.rst
+++ b/docs/source/contribute-a-recipe.rst
@@ -13,6 +13,7 @@ If you're using a fork (set up as :ref:`above <github-setup>`):
 
     git checkout master
     git pull upstream master
+    git push origin master
 
 If you're using a clone:
 


dlaehnemann:
@Core All tests are passing and I could merge this, but I have no clue if this is the right path for commiting stuff to the online documentation. Could someone give a quick review?

daler:
@dlaehnemann yep, this is the right path for updating the documentation. Thanks for adding this!

dlaehnemann:
Thanks @daler! This is basically also a reminder to myself, when following the documentation after not having contributed for a while... ;)

PR #211, opened 2017-11-07
RFC - adding openssh-clients

bgruening:
This is needed for openmpi builds xref: https://github.com/bioconda/bioconda-recipes/pull/6649

and is also be done by conda-forge in similar cases: https://github.com/conda-forge/pytrilinos-feedstock/blob/89340939dc0591e3a31a0ef92fe4367aeb56a9cf/ci_support/run_docker_build.sh#L66

It will prolong our build-time, so we could do this in the base image from which we derive here. A more elaborated stunt would be to make it more flexible and add this package as a yaml-metadata and we install this on request ... this would mean deeper code changes and I'm not sure how frequent this use-case is.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -158,7 +158,7 @@ def build(
     use_base_image = utils.get_meta_value(
         meta,
         'extra', 'container', 'extended-base')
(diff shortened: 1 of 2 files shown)


bgruening:
Ok, this is then ready to be merged and should work for openmpi packages.

PR #213, opened 2017-11-08
Whitespace cleanups in the generated meta.yaml.

cbrueffer:
Adds some whitespace to bring the bioconductor meta.yaml files closer to the format generally used in Bioconda (from what I've seen).

At least it fixes syntax highlighting in vim :-)


daler:
Thanks!

PR #220, opened 2017-11-11
add notice to https://abi-laboratory.pro/tracker/timeline/gsl/

bgruening:


daler:
Never know about this, what a great resource. Would https://abi-laboratory.pro/tracker/ be a better link than the `gsl`-specific one?

PR #221, opened 2017-11-14
Use more conda-forge pins and slim down the Docker image

mbargull:
This does three things:

1. Add `numpy` explicitly to the requirements (previously implicitly by `pandas`). That way we should get `numpy` from `conda-forge` instead of from https://repo.continuum.io/pkgs which would
   * reduce the dependency on https://repo.continuum.io/pkgs,
   * use the `openblas` build instead of the `mkl` one.

   That way we can avoid downloading/installing that big `mkl` package and thus skim maybe 20-30 seconds setup time as well as a couple of hundreds MB from the Docker image.
   The reason we didn't get `numpy` from `conda-forge` in the first place is described here: https://github.com/conda/conda/issues/6199#issuecomment-344097039.
2. Update to `beautifulsoup4 4.6.0`. The previous version is not available on `conda-forge` since its `beautifulsoup4` package only got added this year. I don't think this update should cause any troubles since the package is only used in `bioconda_utils.bioconductor_skeleton` to parse some HTML files. Nonetheless, for reference, the commit which introduced the `4.4.1` pin as well as Beautiful Soup's changelog:
   https://github.com/bioconda/bioconda-utils/pull/18/files#diff-c07a161f6233d35a615a6849711b2d1dR3
   http://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/NEWS.txt
3. Remove tarballs and unused packages from Conda's package cache.
   * This will further decrease the size of the Docker image.
   * It is a fast operation that shouldn't really increase the setup time.
   * It is ok to remove the tarballs since we don't use/need them for anything after their installation.
   * It is ok to remove the unused packages since that are those which came with Miniconda from the `defaults` channels and won't be re-downloaded afterwards as `conda-forge` packages will always be used instead.

### DIFF
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -16,6 +16,7 @@ docutils
 sphinx_rtd_theme
 involucro=1.1.2
 pandas=0.19.2
+numpy
 pygithub=1.29
 colorlog=2.10.*
 six=1.10.0
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -1,6 +1,6 @@
 anaconda-client=1.6.3
 argh=0.26.2
-beautifulsoup4=4.4.1
+beautifulsoup4=4.6.0
 conda=4.3.21
 conda-build=2.1.16
 galaxy-lib=17.9.7
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -16,7 +16,7 @@ docutils
 sphinx_rtd_theme
 involucro=1.1.2
 pandas=0.19.2
-numpy
+numpy=1.12.1
 pygithub=1.29
 colorlog=2.10.*
 six=1.10.0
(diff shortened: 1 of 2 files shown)


mbargull:
Docker image shrinks from `913 MB` to `549 MB`:
https://hub.docker.com/r/bioconda/bioconda-utils-build-env/tags/

daler:
@mbargull this is great! I'm not worried about the bs4 changes, these can be fixed next time we update bioconductor packages if needed. Thanks for the fixes and detailed write-up.

Only question I have is: should we pin numpy?

mbargull:
Right, that was also my question that I forgot to add ;D
I don't see any harm in pinning it to `1.12.1` which I believe was the version that was previously picked by Conda. Should I add that pin?

bgruening:
conda-forge is currently changing the strategy of pinning numpy this strict. So I think this is not really needed, but can also not hurt.

Thanks for this detective work!

daler:
Agreed, I don't see the harm in pinning.

chambm:
Will this change cascade down to bioconda recipes themselves? Currently I reinstall multiqc with the non-mkl numpy for the same reason: to trim my docker image back by 500mb.

mbargull:
@chambm: Sadly, I don't think so. (I believe, but haven't looked into it:) For the Biocontainers the packages will be installed into a separate Conda environment (which usually is a good thing). That separate environment then does not track the `blas_openblas` feature, hence Conda will do its usual thing and prefer the `numpy` that doesn't add a new "feature" into the environment, i.e., the one coming from https://repo.continuum.io/pkgs/ including its `mkl` dependency.

PR #222, opened 2017-11-16
adding in environmental variable to pkg_test

jerowe:
@bgruening 

Here is a really simple and possibly stupid way to fix the hardcoded biocontainers problem.

Let me know if you think there is a better way to do this.

### DIFF
--- a/bioconda_utils/pkg_test.py
+++ b/bioconda_utils/pkg_test.py
@@ -126,11 +126,15 @@ def test_package(
     tests = get_tests(path)
     logger.debug('Tests to run: %s', tests)
 
+    target = 'biocontainers'
+    if os.environ.get('QUAY_TARGET') is not None:
+        target = os.environ.get('QUAY_TARGET')
+
     cmd = [
         'mulled-build',
         'build-and-test',
         spec,
-        '-n', 'biocontainers',
+        '-n', target,
         '--test', tests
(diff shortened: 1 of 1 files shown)


jerowe:
I think it would be best to get it from the command line argument, but I need to sit down and figure out how to do that.

bioconda_utils/pkg_test.py
```diff
@@ -126,11 +126,15 @@ def test_package(
     tests = get_tests(path)
     logger.debug('Tests to run: %s', tests)
 
+    target = 'biocontainers'
+    if os.environ.get('QUAY_TARGET') is not None:
```
bgruening: Can we call it CONTAINER_NAMESPACE or something?

bgruening:
A proper fix with command line args would be better I agree. But I'm also not opposed to merge this one for the time being.

bioconda_utils/pkg_test.py
jerowe: We can name it whatever you like . ;-)

jerowe: Changed it.

jerowe:
I can confirm that this works. ;-)

jerowe:
@bgruening , I changed this so that it gets the original command line argument.

Issue #176, opened 2017-09-11
Using bioconda-utils as a site specific package builder


jerowe:
pinging @bioconda/core 

This is mostly just a shout out for the great work you've done on infrastructure. ;-)

We are going to be using the bioconda-utils package and build system (including mulled builds and whatnot), as a way of building our analysis/lab/project packages, which are just meta packages of bioconda and sometimes R packages. Initially we were just using conda-envs, but that has proven to be problematic if things change upstream, so instead, we will just build packages and leave them there. 

I've been taking a look at the code, and it looks likes its ready for being site specific. I think I should only have to change some of the variables in the scripts/travis-run.sh script, and of course the .travis needs to have our own github token and things, but I will let you know if you run into any snags.

johanneskoester:
Great. I think we are all happy to merge any changes that make bioconda-utils more configurable and remove hardcoded values.

johanneskoester:
Btw. for [Snakemake](https://snakemake.bitbucket.io), I have developed a way to snapshot a conda env and package it in a portable way for e.g. upload to [zenodo](https://zenodo.org). If that is of interest for you, one could probably also think about separating this out into a standalone tool.

jerowe:
@johanneskoester  - that could be interesting. I still use conda-envs quite a bit. For a researcher the conda-env interface is more straight forward than creating a recipe, building it, and uploading it.

For the main infrastructure though I want the rest of what bioconda utils has to offer, mainly creating the docker, hopefully soon singularity containers, and when I can finally sit down and do it the easybuild configurations those too.

jerowe:
So I finally sat down and got started on this today. I didn't even end up needing to change any code, just a bit of the travis-run.sh in the bioconda-recipes, so that was great! I would say I'm about 90% there. I just have a few questions.

What does the github integration do? Just comments? If its something that I want I will have to open a PR, because it does have a hardcoded 'bioconda' user within the code, that IMO a site should be able to change to their own user.

I see that the build_recipes takes an argument 'label', but I'm not sure how to get at those from the command line. Is it possible to add my own custom labels from the existing code/interface, or will I need to write an addon to the anaconda_upload function? 

The bioconda-utils is a really robust build system, and I think its a major perk for the community to be able to use the same build system with minimal changes and be create their own site specific meta packages.

daler:
[...]

PR #223, opened 2017-11-16
expose labels to cli

daler:
xref https://github.com/bioconda/bioconda-utils/issues/176

@jerowe this will attach the label to all recipes built, e.g., 
```
bioconda-utils --label test --packages "*-testing"
```

Do you need to specify per-package labels? Might need a different mechanism for that if so.


daler:
How about supplying a mapping of recipe:label in an additional YAML file? Could even map patterns of recipes to labels, and have the same package in different labels:

```yaml
label1:
  - bowtie2
  - samtools

label2:
  - samtools
  - bioconductor-*
 ```
I guess another option would be storing the label[s] in the recipe's meta.yaml, but I think you'd want to maintain separation of recipe and deployment.

epruesse:
ping @daler @jerowe - do we still need this or can it be closed?

daler:
Closing, not sure this is the right direction to go in for labels.

PR #226, opened 2017-11-17
add docs for extended base

### LINKED ISSUE #225: Documentation for extended-base image
I believe it's currently not documented when to use `extra: {container: {extended-base: True}}`.
Use cases from the top of my head:
- Unicode support, e.g., dependency on `click`
- `libGL.so.1` dependency (sometimes through `matplotlib`, I think)
- `openssl` dependency, e.g., through `openmpi`

daler:
fixes #225


mbargull:
Thanks!

PR #231, opened 2017-11-30
Install Bioconductor data packages in conda R library.

jdblischak:
On Ubuntu, a user library is automatically set, e.g. `~/R/x86_64-pc-linux-gnu-library/3.4`. When I install a Bioconductor data package, e.g. GenomeInfoDb, it gets install in my Ubuntu user library instead of in the conda R library.

This PR updates the `post-link.sh` file created by `bioconductor-skeleton.py` to install the Bioconductor data package in the conda R library path. I tested that this works on Ubuntu.

### DIFF
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -812,7 +812,7 @@ def write_recipe(package, recipe_dir, config, force=False, bioc_version=None,
             fi
 
             # Install and clean up
-            R CMD INSTALL --build $TARBALL
+            R CMD INSTALL --library=$PREFIX/lib/R/library --build $TARBALL
             rm $TARBALL""")
         with open(os.path.join(recipe_dir, 'post-link.sh'), 'w') as fout:
             fout.write(dedent(post_link_template))


jdblischak:
Actually, I think I may have missed the larger issue here. The problem is that `post-link.sh` uses the first `R` it finds on `PATH`. Thus my current solution installs the package in the right place, but it uses the wrong R to build and install the package. Investigating a better option.

bgruening:
@jdblischak why do you have multiple R versions in your PATH ;)

jdblischak:
>The problem is that `post-link.sh` uses the first `R` it finds on `PATH`. Thus my current solution installs the package in the right place, but it uses the wrong R to build and install the package.

Sorry. I think I was actually wrong about that. I used a fresh Docker container with no system R installed, and the conda environment properly used the newly installed R even though it wasn't yet in the PATH. We could add `$PREFIX/bin/R` to be 100% sure it was using the conda R, but from my testing that doesn't seem necessary.

mbargull:
I might be wrong, but I believe the `post-link.sh` scripts are not necessarily run from inside the environment the package is being installed into. That would mean 
```bash
source activate
conda create -n my-env my-data-package
```
or
```bash
source activate
conda create -n my-env
conda install -n my-env my-data-package
```
will behave differently -- i.e., use another `R` or have none available -- from
```bash
source activate
conda create -n my-env
source activate my-env
conda install my-data-package
```
. Hence we'd need to workaround that. I'm not sure we can or should use `source activate "$PREFIX"` in the post link script as I don't know whether the environment is fully created at that point or might be in an unfinished/inconsistent state.
So it might be best to just only use `"$PREFIX/bin/R"` and provide it with all the information it would get from an activated environment.  (But I don't speak R and know nothing about `R`'s prerequisites...)

mbargull:
Oh, and using `wget` would also be problematic for the same reasons...

jdblischak:
@mbargull Thanks for sharing your ideas. I am putting together some reproducible examples to help us confirm the behavior.

> I might be wrong, but I believe the post-link.sh scripts are not necessarily run from inside the environment the package is being installed into.

That was my intuition as well, which is why I thought that `$PREFIX/bin/R` would be necessary. However I've tested it in a docker environment with no system R installed and in an docker environment with a different version of R installed system-wide, and both were able to properly use the R in the newly created environment despite it not being activated. I have no idea how conda makes this work, and frankly it makes me a little uneasy that it does. In the typical `build.sh` script for an R package, it uses `$R`. However I don't know how that is defined or whether `$R` is available for `post-link.sh`.

> I'm not sure we can or should use source activate "$PREFIX" in the post link script as I don't know whether the environment is fully created at that point or might be in an unfinished/inconsistent state.
So it might be best to just only use "$PREFIX/bin/R" and provide it with all the information it would get from an activated environment. (But I don't speak R and know nothing about R's prerequisites...)

We shouldn't need to activate the entire environment (and as you mention, it's probably not possible to activate an environment while it is being created). As long as we point to the R executable and specify the path to install the package, it should work. If there were other non-R dependencies that would be needed to build the package, then we could run into trouble, but Bioconductor data packages presumably don't have any non-R dependencies.

> Oh, and using wget would also be problematic for the same reasons...

Good point! One option would be to add `wget` as a dependency for all Bioconductor data packages and then refer to `$PREFIX/bin/wget` in `post-link.sh`. A second option would be to use `curl`, though  I don't know how ubiquitous `curl` is across different Unix-alike distributions.

daler:
@jdblischak @mbargull thanks for working on this.

Just chiming in here . . . for the bioconductor data packages created by the skeleton script, `wget` is automatically added as a run dependency, which, based on my tests, was sufficient for the `post-link.sh` step. Same with `r-base`.

However, https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html disagrees, explicitly saying that the post-link script should not depend on any packages. Despite this, I could not figure out any other way to avoid swamping our channel storage quota with large data packages. And it seems to work most of the time. But if you can come up with a better way that would be fantastic!

As far as `$R` vs `$PREFIX/bin/R` and what gets installed where, I think the biggest thing to be aware of is if the `R_LIBS` environment variable exists. And possibly `R_HOME`. @jdblischak maybe setting these would change the behavior of your test docker containers? I did not test these different combinations when writing the original post-link.sh. My initial guess would be that if we temporarily unset those env vars in `post-link.sh` and use `$PREFIX/bin/R`, it could avoid problems.

mbargull:
It's very strange indeed that it works in all or some cases...
I guess to really get what's going on, we'll have to either dig into `conda`'s code and see what's being done there or do some experimentation, e.g., by creating some dummy packages that show us the outputs from `set` or `export` in their `post-link.sh`. (Although I myself won't have time to look into that anytime soon, I believe..)

jdblischak:
> @jdblischak @mbargull thanks for working on this.

@daler No problem! I like your solution to the problem of Bioconductor data packages, and I think we'll be able to work around this edge case.

> Just chiming in here . . . for the bioconductor data packages created by the skeleton script, wget is automatically added as a run dependency, which, based on my tests, was sufficient for the post-link.sh step. Same with r-base.

Ah yes, sorry. Forgot that you already had this covered.

> However, https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html disagrees, explicitly saying that the post-link script should not depend on any packages. Despite this, I could not figure out any other way to avoid swamping our channel storage quota with large data packages. And it seems to work most of the time. But if you can come up with a better way that would be fantastic!

Yes, it does work most of the time, and the reason it is failing here is not because the post-link script depends on newly installed packages.

> As far as $R vs $PREFIX/bin/R and what gets installed where, I think the biggest thing to be aware of is if the R_LIBS environment variable exists. And possibly R_HOME. @jdblischak maybe setting these would change the behavior of your test docker containers? I did not test these different combinations when writing the original post-link.sh. My initial guess would be that if we temporarily unset those env vars in post-link.sh and use $PREFIX/bin/R, it could avoid problems.

I think my solution of adding `--library=$PREFIX/lib/R/library` fixes the problem without having to manipulate the environment variables that the user has set. This seems safer to me. Do you disagree?

jdblischak:
To create a consistent testing environment that isn’t affected by my local setup, I created a Docker image. You can run it with:

```
docker run -it jdblischak/ubuntu-bioconda
```

If the user has not created an R user library, then post-link.sh works correctly. This is true even if `wget` and `R` are not installed.

```
which wget
which R
conda create -y -n test1 bioconductor-genomeinfodbdata
ls /opt/conda/envs/test1/lib/R/library/ | grep GenomeInfoDbData
```

If a user library exists (even if R is not installed), post-link.sh writes Bioconductor data packages to the user library.

```
which R
mkdir -p ~/R/x86_64-pc-linux-gnu-library/3.4
conda create -y -n test2 bioconductor-genomeinfodbdata
ls /opt/conda/envs/test2/lib/R/library/ | grep GenomeInfoDbData
ls ~/R/x86_64-pc-linux-gnu-library/3.4
```

However, this can be fixed by adding `--library=$PREFIX/lib/R/library` in post-link.sh.

```
rm ~/R/x86_64-pc-linux-gnu-library/3.4/GenomeInfoDbData/ -r
sed -i s/'R CMD INSTALL --build $TARBALL'/'R CMD INSTALL --library=$PREFIX\/lib\/R\/library --build $TARBALL'/ ~/bioconda-recipes/recipes/bioconductor-genomeinfodbdata/post-link.sh
conda build --R 3.4.1 ~/bioconda-recipes/recipes/bioconductor-genomeinfodbdata
conda create -y -n test3 —-use-local bioconductor-genomeinfodbdata
ls /opt/conda/envs/test3/lib/R/library/ | grep GenomeInfoDbData
ls ~/R/x86_64-pc-linux-gnu-library/3.4
```

And just as a sanity check, I confirmed that this change in behavior is not not due to the difference in installing a local versus remote package.

```
sed -i s/'R CMD INSTALL --library=$PREFIX\/lib\/R\/library --build $TARBALL'/'R CMD INSTALL --build $TARBALL’/ ~/bioconda-recipes/recipes/bioconductor-genomeinfodbdata/post-link.sh
conda build --R 3.4.1 ~/bioconda-recipes/recipes/bioconductor-genomeinfodbdata
conda create -y -n test4 —-use-local bioconductor-genomeinfodbdata
ls /opt/conda/envs/test4/lib/R/library/ | grep GenomeInfoDbData
ls ~/R/x86_64-pc-linux-gnu-library/3.4
```

bgruening:
I spend some time and tried different things and I could not see any disadvantage of adding `--library=$PREFIX/lib/R/library` to the script. However, I also haven't seen much problem with the current setup. So :+1: from me to this change.

One thing I want to mention is that under Windows R will install everything into `R/library` and not `lib/R/library`. This might fool us as soon as we create noarch packages.

jdblischak:
> I spend some time and tried different things and I could not see any disadvantage of adding --library=$PREFIX/lib/R/library to the script. However, I also haven't seen much problem with the current setup. So 👍 from me to this change.

@bgruening Thanks for testing this out!

> One thing I want to mention is that under Windows R will install everything into R/library and not lib/R/library. This might fool us as soon as we create noarch packages.

Good point. Hopefully since bioconda only supports macOS and Linux, we can assume that the noarch bioconda Bioconductor packages will not be installed on Windows. Though that reminds me, I should do some more testing of this PR on macOS.

jdblischak:
**tl;dr** I think this is ready to merge. `post-link.sh` uses the executables defined in the newly created environment by default.

I did some more investigation into which executables are being used by `post-link.sh`. To do this, I used the following idea from the [docs](https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html) that @daler sent:

> The scripts should not write to stdout or stderr unless an error occurs, but they may write to $PREFIX/.messages.txt, which is shown after conda completes all actions.

Thus I added the following 3 lines to `post-link.sh`:

```
which R >> $PREFIX/.messages.txt
Rscript --version >> $PREFIX/.messages.txt 2>&1
which wget >> $PREFIX/.messages.txt
```

I tested the following 3 scenarios on Ubuntu 16.04:

* No `wget` and no `R`
* `wget` and `R` (3.2.3) installed with Apt
* `wget` and `R` (3.2.3) installed with Apt *and* `wget` and `R` (3.3.2) installed in the root conda environment

In all 3 scenarios I get the same result. The executables in the newly created environment are used:

```
/opt/conda/envs/test5/bin/R
R scripting front-end version 3.4.1 (2017-06-30)
/opt/conda/envs/test5/bin/wget
```

mbargull:
@bgruening:
>  One thing I want to mention is that under Windows R will install everything into R/library and not lib/R/library. This might fool us as soon as we create noarch packages.

I guess in that case `post-link.bat` instead of `post-link.sh` would have to come into play either way.

mbargull:
@jdblischak's findings are also inline with the associated code which prepends the post-link's directory (i.e., environment's `bin` folder):
https://github.com/conda/conda/blob/6807e91d36a125d19248acdecac960c0c7c2dd55/conda/core/link.py#L492

bgruening:
@daler ok to merge?

daler:
@jdblischak @mbargull @bgruening thanks so much for this. I don't have time right now to double-check the tests, but you all are doing more detailed tests than I originally did. I would add that a pre-unlink.sh (https://github.com/bioconda/bioconda-utils/issues/234) should really be added before building/rebuilding packages, but that doesn't have to block this PR.

daler:
(oh, and to be clear: yes, good to merge!)

Issue #235, opened 2017-12-04
Coordinating rebuilding Bioconductor data packages


jdblischak:
This Issue is to coordinate all the changes being made to `bioconductor-skeleton` in its handling of Bioconductor data packages (and subsequently rebuilding them).

cc @bgruening @mbargull @daler

To do:

- [x] Install in conda R library even if user has a user-specific R library (#231)
- [x] Create `pre-unlink.sh` to uninstall the data package (#234, #238)
- [x] Check md5 checksum of downloaded package on Linux (#233)
- [x]  Determine order for re-building packages (`bioconda-utils dag` accepts multiple globs, not a list of package names, so I'm not sure the easiest way to do this)
- [x] Fix tarball download issues (#230, #236, #237)

By my count, there are 63 Bioconductor data packages that would need to be updated:

```
$ cd bioconda-recipes
$ git log -1 --pretty=oneline
fe2c4f1377c0b55ef4550b53f5a0e247686cc751 update seqkit to v0.7.2 (#6985)
$ find recipes/bioconductor-* -name post-link.sh | wc -l
63
```

jdblischak:
The order in which we rebuild these is not critical. The CI builds will pass since they already passed previously. However, we won't be able to rebuild them all at once, so to minimize the effect on downstream users, I used `conda build-all` to order the packages based on their dependencies (this only made a difference for a few of the packages, which is why it is mostly alphabetical order). I suggest that once we finalize the recent updates to the recipe skeleton, we submit PRs with 5-10 recipes each (please advise if that is too many to fit into the CI time limits) in the following order (note that the total is 66 since some packages have multiple versions):

```
TMP=/tmp/bioc
mkdir -p $TMP
for pkg in recipes/bioconductor-*
do
  if [[ -e $pkg/post-link.sh ]]; then
    cp -R $pkg $TMP
  fi 
done
conda build-all $TMP --dry-run --matrix-conditions "r-base ==3.4.1" | \
  grep bioconductor- | \
  cut -d'-' -f1-3 | \
  sed s/\\t/'- [ ] '/
```

- [ ] bioconductor-affydata-1.26.0
- [ ] bioconductor-all-1.20.0
- [ ] bioconductor-bladderbatch-1.14.0
- [ ] bioconductor-bsgenome.celegans.ucsc.ce10-1.4.0
- [ ] bioconductor-bsgenome.celegans.ucsc.ce6-1.4.0
[...]

Issue #234, opened 2017-12-03
pre-unlink script for Bioconductor data packages


mbargull:
While looking at the fixes @jdblischak provides for the Bioconductor skeleton I noticed that only a `post-link.sh` but not a corresponding `pre-unlink.sh` is created.
I guess Conda won't be able to track changes make by `post-link.sh`, thus uninstalling a data package may not remove it properly.
(Note that acc. to https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html
> `pre-unlink`—Executed before the package is removed. An error is indicated by a nonzero exist and causes the removal to fail.

which would mean one should take care to make those not fail easily to not create "uninstallable" packages -- however `conda` will actually only output a warning and carry on, see:
https://github.com/conda/conda/blob/6807e91d36a125d19248acdecac960c0c7c2dd55/conda/core/link.py#L494-L517
)

jdblischak:
> thus uninstalling a data package may not remove it properly.

@mbargull I did a quick test to confirm your suspicion. While removing the Bioconductor package with `conda remove` makes it forgotten by conda, the package still exists untouched and able to be loaded by R.

PR #237, opened 2017-12-04
bioconductor skeleton: fixes for tarball download handling

mbargull:
Incorporates #230. Addresses #236.
There were some errors in checking the response status codes.

PR #238, opened 2017-12-05
Add template pre-unlink.sh for Bioconductor data packages

jdblischak:
xref: #234 #235 

This PR adds a template `pre-unlink.sh` file for Bioconductor data packages that runs prior to removing the package from the conda environment. Without this file, the R package remains installed and accessible in `$PREFIX/lib/R/library` even after it has been removed.

It also makes two minor edits to the `post-link.sh` file:

1. Removes indentation from the line that defines `MD5`
1.  The subdirectory within `$PREFIX/share` that was created to download the tarball is deleted instead of only deleting the tarball

### DIFF
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -776,10 +776,9 @@ def write_recipe(package, recipe_dir, config, force=False, bioc_version=None,
         post_link_template += urls
         post_link_template += dedent(
             '''
-        )
+            )
             MD5="{proj.md5}"
-            '''.format(proj=proj, urls=urls)
-        )
+            '''.format(proj=proj, urls=urls))
         post_link_template += dedent(
             """
             # Use a staging area in the conda dir rather than temp dirs, both to avoid
@@ -817,7 +816,7 @@ def write_recipe(package, recipe_dir, config, force=False, bioc_version=None,
 
             # Install and clean up
             R CMD INSTALL --library=$PREFIX/lib/R/library --build $TARBALL
-            rm $TARBALL""")
+            rm -r $STAGING""")
         with open(os.path.join(recipe_dir, 'post-link.sh'), 'w') as fout:
             fout.write(dedent(post_link_template))
         pre_unlink_template = "rm -r $PREFIX/lib/R/library/{0}\n".format(package)
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -815,11 +815,12 @@ def write_recipe(package, recipe_dir, config, force=False, bioc_version=None,
             fi
 
             # Install and clean up
-            R CMD INSTALL --library=$PREFIX/lib/R/library --build $TARBALL
-            rm -r $STAGING""")
+            R CMD INSTALL --library=$PREFIX/lib/R/library $TARBALL
+            rm $TARBALL
+            rmdir $STAGING""")
         with open(os.path.join(recipe_dir, 'post-link.sh'), 'w') as fout:
             fout.write(dedent(post_link_template))
-        pre_unlink_template = "rm -r $PREFIX/lib/R/library/{0}\n".format(package)
+        pre_unlink_template = "R CMD REMOVE --library=$PREFIX/lib/R/library/ {0}\n".format(package)
         with open(os.path.join(recipe_dir, 'pre-unlink.sh'), 'w') as fout:
             fout.write(pre_unlink_template)
 
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -820,5 +820,8 @@ def write_recipe(package, recipe_dir, config, force=False, bioc_version=None,
             rm $TARBALL""")
         with open(os.path.join(recipe_dir, 'post-link.sh'), 'w') as fout:
             fout.write(dedent(post_link_template))
+        pre_unlink_template = "rm -r $PREFIX/lib/R/library/{0}\n".format(package)
+        with open(os.path.join(recipe_dir, 'pre-unlink.sh'), 'w') as fout:
+            fout.write(pre_unlink_template)
 
     logger.info('Wrote recipe in %s', recipe_dir)


bioconda_utils/bioconductor_skeleton.py
```diff
@@ -813,8 +812,11 @@ def write_recipe(package, recipe_dir, config, force=False, bioc_version=None,
 
             # Install and clean up
             R CMD INSTALL --library=$PREFIX/lib/R/library --build $TARBALL
-            rm $TARBALL""")
+            rm -r $STAGING""")
```
mbargull: Can we always be sure `$STAGING` has no other content than `$TARBALL`? If not, I'd rather use
```bash
rm $TARBALL
rmdir $STAGING
```

bioconda_utils/bioconductor_skeleton.py
```diff
@@ -813,8 +812,11 @@ def write_recipe(package, recipe_dir, config, force=False, bioc_version=None,
...
         with open(os.path.join(recipe_dir, 'post-link.sh'), 'w') as fout:
             fout.write(dedent(post_link_template))
+        pre_unlink_template = "rm -r $PREFIX/lib/R/library/{0}\n".format(package)
```
mbargull: Does a removal of that directory suffice or does
```bash
R CMD INSTALL --library=$PREFIX/lib/R/library --build $TARBALL
```
add or modify any other files?
(I know next to nothing about R, so someone more knowledgeable than me should be review this 😉)

bioconda_utils/bioconductor_skeleton.py
jdblischak: `$STAGING` was created at the beginning of this script specifically to store the tarball. To my knowledge other conda processes will not use `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM`. But even if they did, these files would no longer be needed if the package is removed.

If we do want to allow the possibility of other files to exist there after the package is removed, we'd want to do:

```
rm $TARBALL
rmdir --ignore-fail-on-non-empty $STAGING
```

Otherwise the script would fail due to the directory not being empty.

mbargull: > `$STAGING` was created at the beginning of this script specifically to store the tarball. To my knowledge other conda processes will not use `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM`. But even if they did, these files would no longer be needed if the package is removed.

I agree, but who knows what people might do ¯\\_(ツ)\_/¯. I just like being explicit when deleting stuff -- but I'm fine either way as that's only being pedantic 😉.

> Otherwise the script would fail due to the directory not being empty.

No, the `post-link.sh` is not executed with the `-e` switch, hence it will not exit on error. Otherwise we'd also have to take care of the previous commands (`mkdir`, `wget`, `rm`) which can also fail any time. IMHO, we should even explicitly not use `--ignore-fail-on-non-empty` to not suppress error output.

bioconda_utils/bioconductor_skeleton.py
jdblischak: I did a quick test. The only files added by `R CMD INSTALL` are in `lib/R/library`. There are other files left after removing the package, but they are all created by conda.

```
$ conda create -y -n test bioconductor-genomeinfodbdata
$ find /opt/conda/ -iname *genomeinfodbdata*
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0.tar.bz2
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0/bin/.bioconductor-genomeinfodbdata-post-link.sh
/opt/conda/envs/test/conda-meta/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0.json
/opt/conda/envs/test/share/bioconductor-genomeinfodbdata-0.99.1-0
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/help/GenomeInfoDbData.rdb
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/help/GenomeInfoDbData.rdx
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/scripts/updateGenomeInfoDbData.R
/opt/conda/envs/test/bin/GenomeInfoDbData_0.99.1_R_x86_64-pc-linux-gnu.tar.gz
/opt/conda/envs/test/bin/.bioconductor-genomeinfodbdata-post-link.sh
$ conda remove -y -n test bioconductor-genomeinfodbdata
$ find /opt/conda/ -iname *genomeinfodbdata*
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0.tar.bz2
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0/bin/.bioconductor-genomeinfodbdata-post-link.sh
/opt/conda/envs/test/share/bioconductor-genomeinfodbdata-0.99.1-0
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/help/GenomeInfoDbData.rdb
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/help/GenomeInfoDbData.rdx
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/scripts/updateGenomeInfoDbData.R
/opt/conda/envs/test/bin/GenomeInfoDbData_0.99.1_R_x86_64-pc-linux-gnu.tar.gz
```

The files in `pkgs/` are also left behind when removing a normal R recipe:

```
$ conda install -y -n test r-registry
$ find /opt/conda/ -iname registry
/opt/conda/pkgs/r-registry-0.3-r3.4.1_0/lib/R/library/registry
/opt/conda/pkgs/r-registry-0.3-r3.4.1_0/lib/R/library/registry/R/registry
/opt/conda/envs/test/lib/R/library/registry
/opt/conda/envs/test/lib/R/library/registry/R/registry
$ conda remove -y -n test r-registry
$ find /opt/conda/ -iname registry
/opt/conda/pkgs/r-registry-0.3-r3.4.1_0/lib/R/library/registry
/opt/conda/pkgs/r-registry-0.3-r3.4.1_0/lib/R/library/registry/R/registry
```

jdblischak:
Thanks for the review @mbargull!

bioconda_utils/bioconductor_skeleton.py
jdblischak: > I agree, but who knows what people might do ¯\_(ツ)_/¯. I just like being explicit when deleting stuff -- but I'm fine either way as that's only being pedantic

I actually like the 2-step process. It was what I was going to do originally, but feared it was too verbose. I'm happy to switch to using `rmdir`.

> No, the post-link.sh is not executed with the -e switch, hence it will not exit on error.

Good point! I always forget that Bash doesn't stop on error by default.

bioconda_utils/bioconductor_skeleton.py
mbargull: The `pkgs` folder is just Conda's package cache, no worries about that. But
```
/opt/conda/envs/test/share/bioconductor-genomeinfodbdata-0.99.1-0
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/help/GenomeInfoDbData.rdb
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/help/GenomeInfoDbData.rdx
/opt/conda/envs/test/lib/R/library/GenomeInfoDbData/scripts/updateGenomeInfoDbData.R
/opt/conda/envs/test/bin/GenomeInfoDbData_0.99.1_R_x86_64-pc-linux-gnu.tar.gz
```
are not created by `conda` itself but likely by `R`.
Does `R` have an uninstall command, i.e., some direct pendant to `R CMD INSTALL`?

jdblischak: > Does R have an uninstall command, i.e., some direct pendant to R CMD INSTALL?

Yes, it does. We could switch to it, but it has the same effect in this case as the current `rm -r` (i.e. remove the directory in `lib/R/library`):

```
$ conda create -y -n test bioconductor-genomeinfodbdata
$ source activate test
$ R CMD REMOVE --library=/opt/conda/envs/test/lib/R/library/ GenomeInfoDbData
$ find /opt/conda/ -iname *genomeinfodbdata*
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0.tar.bz2
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0
/opt/conda/pkgs/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0/bin/.bioconductor-genomeinfodbdata-post-link.sh
/opt/conda/envs/test/conda-meta/bioconductor-genomeinfodbdata-0.99.1-r3.4.1_0.json
/opt/conda/envs/test/share/bioconductor-genomeinfodbdata-0.99.1-0
/opt/conda/envs/test/bin/GenomeInfoDbData_0.99.1_R_x86_64-pc-linux-gnu.tar.gz
/opt/conda/envs/test/bin/.bioconductor-genomeinfodbdata-post-link.sh
```

The directory in `share` is the directory `$STAGING` created by `post-link.sh` (which as we discussed above, this PR also proposes removing) and the files in `bin/` are added by conda.

mbargull: I'd go for `R CMD REMOVE`, but only because I don't know what `R`'s `INSTALL`/`REMOVE` can actually do..
So regardless if we use one or the other removal command, for the `GenomeInfoDbData` example only `$PREFIX/bin/GenomeInfoDbData_0.99.1_R_x86_64-pc-linux-gnu.tar.gz` will remain when this PR is adopted. This looks like some compilation file from `R` (who knows why the heck this ends up in `bin`...). To confirm this, `diff`ed the environment pre-install, post-install and post-removal:
```
$ diff -rqx conda-meta init/ installed/
Only in installed/bin: .bioconductor-genomeinfodbdata-post-link.sh
Only in installed/bin: GenomeInfoDbData_0.99.1_R_x86_64-pc-linux-gnu.tar.gz
Only in installed/lib/R/library: GenomeInfoDbData
Only in installed/share: bioconductor-genomeinfodbdata-0.99.1-0
$ diff -rqx conda-meta installed/ removed/
Only in installed/bin: .bioconductor-genomeinfodbdata-post-link.sh
$ diff -rqx conda-meta init/ removed/
Only in removed/bin: GenomeInfoDbData_0.99.1_R_x86_64-pc-linux-gnu.tar.gz
Only in removed/lib/R/library: GenomeInfoDbData
Only in removed/share: bioconductor-genomeinfodbdata-0.99.1-0
```

mbargull: Ah ok, IIUC, `R CMD INSTALL --build` installs the package first and then creates the tarball in `bin`. Since I don't believe we need/want that tarball anyway, shouldn't the command just be
```bash
R CMD INSTALL --library=$PREFIX/lib/R/library $TARBALL
```
without the `--build`?

jdblischak:
@mbargull I've updated this PR based on your feedback:

1. Do not use the `--build` flag from the install command
1. Remove tarball and staging directory in two steps
1. Use `R CMD REMOVE` to remove the package

mbargull:
@bioconda/r: Is this good to go or do some of you have any objections?

jdblischak:
Note that I wasn't able to test my most recent commit because #237 was merged into my PR: https://github.com/bioconda/bioconda-utils/pull/237#issuecomment-350775885

jdblischak:
I cleaned up the commit history using `git cherry-pick` (I don't love the automatic merge commits that get added, but I see why that would be useful in some cases).

I performed the following test to confirm that Bioconductor data packages are installed in the correct directory (even if a user directory exists) and are properly removed with `conda remove`.

```
mkdir -p ~/R/x86_64-pc-linux-gnu-library/3.4
bioconda-utils bioconductor-skeleton recipes/ config.yml bovineprobe
conda build --R 3.4.1 recipes/bioconductor-bovineprobe/
conda create -y -n test --use-local bioconductor-bovineprobe
source activate test
Rscript -e "library(bovineprobe)"
# Installed in the correct location
ls ~/R/x86_64-pc-linux-gnu-library/3.4
conda remove -y bioconductor-bovineprobe
# There is little trace of the package after removal
find /opt/conda/ -name "bovineprobe"
```

After removal, here are the files still around:

```
# These are because I built the package locally
/opt/conda/conda-bld/src_cache/bovineprobe_2.18.0.tar.gz
/opt/conda/conda-bld/linux-64/bioconductor-bovineprobe-2.18.0-r3.4.1_0.tar.bz2
/opt/conda/lib/python3.5/site-packages/bioconda_utils-0.10.0-py3.5.egg/bioconda_utils/cached_bioconductor_tarballs/bovineprobe_2.18.0.tar.gz
# These are files that conda leaves behind.
/opt/conda/pkgs/bioconductor-bovineprobe-2.18.0-r3.4.1_0.tar.bz2
/opt/conda/pkgs/bioconductor-bovineprobe-2.18.0-r3.4.1_0
/opt/conda/pkgs/bioconductor-bovineprobe-2.18.0-r3.4.1_0/bin/.bioconductor-bovineprobe-post-link.sh
/opt/conda/pkgs/bioconductor-bovineprobe-2.18.0-r3.4.1_0/bin/.bioconductor-bovineprobe-pre-unlink.sh
```

mbargull:
> I cleaned up the commit history using git cherry-pick (I don't love the automatic merge commits

Clicking those GitHub "Update branch" buttons is just like taking the car instead of a bike: Quite convenient but dirty.. thanks for the cleanup :wink:

> After removal, here are the files still around:

Great, all of those are expected/intended to remain. So, this PR should be ready to be merged!

jdblischak:
@daler Please review and merge

daler:
Thank you!

PR #241, opened 2017-12-15
Recursive cran functionality added to bioconductor-skeleton.py 

MathiasHaudgaard:
Elixir in Aarhus has added a feature to bioconductor-skeleton so if there are any cran dependencies, there will be made recipes for these dependencies. However, we still have the following problems:
- What if you want to make the same recipe just with a different version. Right now that situation results in an error because the script sees that the recipe is already made.
- Not sure if we're handling r-base correct.

Feel free to comment and give suggestions :-)

### DIFF
--- a/bioconda_utils/cran_skeleton.py
+++ b/bioconda_utils/cran_skeleton.py
@@ -68,9 +68,12 @@ def clean_build_file(package, no_windows=False):
     path = package + '/build.sh'
     with open(path, 'r') as build:
         lines = list(build.readlines())
-        lines = filter_lines_regex(lines, r'^mv\s.*$', '')  # Remove lines with mv commands
-        lines = filter_lines_regex(lines, r'^grep\s.*$', '')  # Remove lines with grep commands
-        lines = filter_lines_regex(lines, r'^\s*#.*$', '')  # Removes the lines consisting of only comments
+        # Remove lines with mv commands
+        lines = filter_lines_regex(lines, r'^mv\s.*$', '')
+        # Remove lines with grep commands
+        lines = filter_lines_regex(lines, r'^grep\s.*$', '')
+        # Removes the lines consisting of only comments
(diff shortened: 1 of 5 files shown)


maintainers.yaml
```diff
@@ -0,0 +1,9 @@
+extra:
+   recipe-maintainers:
```
jdblischak: Are the GitHub usernames case sensitive?

jdblischak:
> What if you want to make the same recipe just with a different version. Right now that situation results in an error because the script sees that the recipe is already made.

@MathiasHaudgaard Are you referring to the original recipe or the dependencies? What happens when you set `force = True` for `write_recipe`?

maintainers.yaml
bgruening: I think so too.

MathiasHaudgaard: I'll change that :-)

bioconda_utils/cran_skeleton.py
```diff
@@ -0,0 +1,136 @@
+from optparse import OptionParser
```
bgruening: Can we use argparse here?

bioconda_utils/cran_skeleton.py
```diff
@@ -0,0 +1,136 @@
...
+import subprocess as sp
+import re
+from itertools import zip_longest
+import argparse
+import yaml
+
+
```
bgruening: Can we add a short description about the purpose of this file?
This is the file which you started at the ELIXIR hackathon in Paris, right? The port of conda_r_helper_script?

If so we could just add, that this file cleans-up normal cran skeletons to make it conda-forge compliant.

bgruening:
@jdblischak @daler should we add some tests before merging?

daler:
@bgruening @jdblischak tests would be good, I'll work on that.

daler:
Closed via #242

PR #242, opened 2017-12-26
Updates to recursive bioconductor skeleton

daler:
This is an update of #241, I thought it would be cleaner to submit as a separate PR since I made a lot of changes relative to that PR.

Changes relative to #241 include:
- a stable place for maintainers.yaml so it will be installed with bioconda-utils
- BioCProjectPage.dependencies now uses a dict mapping `conda-pkg-name: CRAN package name` which simplifies calls in `cran_skeleton.py`
- lots of docstrings and comments
- ability to force recursive pkgs even if they exist
- ability to skip recursive pkgs if they already exist in conda-forge or bioconda
- using `conda_build.api.skeleton` rather than calling out to shell
- initial tests

In running the tests, I think I'm running across a bug in `conda-build skeleton`. It's tricky because the failure is intermittent, which makes me suspect that something in conda-build skeleton is assuming a particular dict order. The problem seems to be that a package will sometimes pick up the R version pinning and assign it to the package itself. The NotImplementedError that is sometimes raised is from [here](https://github.com/conda/conda-build/blob/2.1.x/conda_build/skeletons/cran.py#L576).

One workaround is to skip the package if NotImplementedError is raised, but that makes it hard to have a robust test. Another would be to keep restarting the tests until they pass. And the best would be to fix the underlying issue, but I haven't dug into the conda-build code for that yet.

I'll continue adding tests and try to come up with a better workaround and/or fix upstream.

cc @bgruening @ArneKr @MathiasHaudgaard @dansondergaard


daler:
Update: For the CRAN intermittent failures I mentioned above, it seems to be the `lattice` package causing issues. I'm now allowing that to fail, but ensuring that other recipes get built. This allows the tests to be stable, and it's less important that upstream gets fixed. If this kind of failure happens in practice, a warning is logged and it's easy enough to write that recipe manually.

Other changes:

- ~91% test coverage across `bioconductor_skeleton` and `cran_skeleton`
- updates and fixes to deal with newer pytest versions that integrate pytest-catchlog but does some overwriting of sys.stderr (basically a workaround for https://github.com/pytest-dev/pytest/issues/2079)
- fixed a case where github token could potentially leak (https://github.com/pytest-dev/pytest/issues/2079) when not running on travis
- new CLI command `clean_cran_skeleton`

daler:
@bgruening @ArneKr @MathiasHaudgaard @dansondergaard anything else you'd like to add before merging?

bgruening:
Wow, this is way more then I expected - thanks a lot for so many new cool improvements!!!

Happy new year @daler!

dansondergaard:
Great work, thanks! :) Happy new year to all.

MathiasHaudgaard:
@boll3 do you still have this error? 
I've just built the package genomeIntervals with no errors.

boll3:
The 'no_windows' argument was removed and everything worked fine from there on.

PR #243, opened 2018-01-03
lint: disable uses_setuptools

mbargull:
The `uses_setuptools` lint suggests to remove `setuptools` dependencies.
But `setuptools` can be a run requirement if a Python package
* uses `pkg_resources` (which is part of `setuptools`),
* uses `setup(entry_points={"console_scripts": [...]})` but doesn't have all those entry points replaced by Conda-created entry points in `meta.yaml` (which thus automatically use `pkg_resources` unlike the Conda-created ones),
* (very unlikely: imports `setuptools` (other than in `setup.py`)).

The previous wording `'setuptools may not be required'` let contributor to unconditionally remove `setuptools` from run/build requirements, I suppose because
* `may not` is more of a "prohibition" phrase,
* `be required` doesn't state we're only talking about run requirements.

This PR fixes and clarifies that wording. But it also disabled `uses_setuptools` all together since that might be the "safest bet" to avoid broken packages -- a surplus dependency isn't nice but also doesn't hurt as much as a missing one...

### DIFF
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -341,7 +341,9 @@ registry = (
     uses_git_url,
     uses_javajdk,
     uses_perl_threaded,
-    uses_setuptools,
+    # removing setuptools from run requirements should be done cautiously:
+    # it breaks packages that use pkg_resources or setuptools console scripts!
+    # uses_setuptools,
     has_windows_bat_file,
 
     # should_be_noarch,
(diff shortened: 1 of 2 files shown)


daler:
Makes sense, and thanks for the clarification. Could you please add these bullet points to [the docs](https://github.com/bioconda/bioconda-utils/blob/master/docs/source/linting.rst) so that we have this info somewhere, and maybe also point out in the docs that this is currently disabled?

mbargull:
Good point! While at it, I also marked the other two disabled linting functions (`already_in_bioconda`, `should_be_noarch`) accordingly.

PR #247, opened 2018-01-09
Add 'HOST_USER_ID' to Docker env vars.

acaprez:
The condaforge/linux-anvil image now runs everything using a
'conda' user account instead of root. Setting this var is
needed to allow mounted volumes to have proper rw permissions.


acaprez:
@daler - ah, so just set it in the `bioconda-utils` code itself?   I was thinking of setting it either in `simulate-travis.py` or `travis-run.sh` and then letting it be inherited, but your way is cleaner I think.

daler:
Yeah, this way hopefully the bioconda-utils tests will pass, and anyone using `bioconda-utils build` will get the fix.

acaprez:
Right -- makes sense.  I just updated my branch with an attempted fix doing it that way.

johanneskoester:
The test somehow hangs. Any idea what could be wrong?

acaprez:
Hmm, it gets to 33% ok and then hangs.  I went ahead and restarted the test just now in case it was a one-time issue.  Otherwise I'll have to defer to someone who's more familiar with the test setup for help troubleshooting.

johanneskoester:
@daler do you have the time? I might get a chance to look at it tomorrow otherwise.

acaprez:
Looks like the linux tests finished ok with the capture disabled (still waiting on OS X).  So maybe the build timeout just needs to be increased?

daler:
@johanneskoester sorry, swamped with teaching and prepping for teaching thru Friday. I'll do it on Sat if it's not fixed by then.

johanneskoester:
Great work, thanks guys!

PR #250, opened 2018-01-14
Skip mulled-build test for conda-dependent recipes

daler:
xref https://github.com/bioconda/bioconda-recipes/pull/7241, ping @johanneskoester and @bgruening. 

Rather than special-case `bioconda-utils`, this skips mulled-build for any recipe depending on conda. Should allow a `bioconda-utils` conda package to be built.


bgruening:
Cool, thanks. I'm now back from workshops and can hopefully be more useful in the next weeks.

PR #255, opened 2018-01-23
Catch any error from running "docker --version"

peterjc:
If Docker is not installed, simulate-travis.py fails in a somewhat confusing way.

This still shows a nasty exception, but a meaningful error is logged first.

### DIFF
--- a/bioconda_utils/docker_utils.py
+++ b/bioconda_utils/docker_utils.py
@@ -370,7 +370,14 @@ class RecipeBuilder(object):
         #  `docker version` command (note the missing dashes) is not consistent
         # between different docker versions. The --version string is the same
         # for docker 1.6.2 and 1.12.6
-        s = sp.check_output(["docker", "--version"]).decode()
+        try:
+            s = sp.check_output(["docker", "--version"]).decode()
+        except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


bioconda_utils/docker_utils.py
```diff
@@ -370,7 +370,14 @@ def _build_image(self):
...
         # for docker 1.6.2 and 1.12.6
-        s = sp.check_output(["docker", "--version"]).decode()
+        try:
+            s = sp.check_output(["docker", "--version"]).decode()
+        except FileNotFoundError as e:
+            logger.error('DOCKER FAILED: Error checking docker version, is it installed?')
+            raise e
+        except sp.CalledProcessError as e:
+            logger.error('DOCKER FAILED: Error checking docker version.')
+            raise e
```
mbargull: Can you remove the ` as e` / ` e` in the `except` and `raise` expressions, resp.? That way we get an unaltered stack trace.

peterjc:
Rebased and revised based on @mbargull's suggestion to avoid alteration to the stack trace (using plain ``except`` and ``raise``).

mbargull:
Thanks @peterjc!

PR #258, opened 2018-01-28
fix liniting of pinnings

bgruening:
I think this is broken since https://github.com/bioconda/bioconda-utils/pull/191

### DIFF
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -295,6 +295,7 @@ def _pin(env_var, dep_name):
         not_pinned = set()
         pinned = set()
         for line in open(os.path.join(recipe, 'meta.yaml')):
+            line = line.rstrip("\n")
             dedented_line = line.lstrip(' ')
             if line.startswith("requirements:"):
                 in_requirements = True
@@ -302,11 +303,11 @@ def _pin(env_var, dep_name):
                 section = "run"
             elif in_requirements and dedented_line.startswith("build:"):
                 section = "build"
-            elif dedented_line and not line.startswith(' ') and not line.startswith("#"):
+            elif line and not line.startswith(" ") and not line.startswith("#"):
                 in_requirements = False
                 section = None
             if in_requirements and dedented_line.startswith('- {}'.format(dep_name)):
-                if pin_pattern.search(line.strip()) is None:
+                if pin_pattern.search(dedented_line) is None:
                     not_pinned.add(section)
                 else:
                     pinned.add(section)
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -295,19 +295,18 @@ def _pin(env_var, dep_name):
         not_pinned = set()
         pinned = set()
         for line in open(os.path.join(recipe, 'meta.yaml')):
+            dedented_line = line.lstrip(' ')
             if line.startswith("requirements:"):
                 in_requirements = True
-            elif in_requirements and line.strip().startswith("run:"):
+            elif in_requirements and dedented_line.startswith("run:"):
                 section = "run"
-            elif in_requirements and line.strip().startswith("build:"):
+            elif in_requirements and dedented_line.startswith("build:"):
                 section = "build"
-            elif not line.startswith(" ") and not line.startswith("#") and line.strip():
+            elif dedented_line and not line.startswith(' ') and not line.startswith("#"):
                 in_requirements = False
                 section = None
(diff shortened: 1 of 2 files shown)


bioconda_utils/lint_functions.py
```diff
@@ -301,10 +301,11 @@ def pin(recipe, meta, df):
                 section = "run"
             elif in_requirements and line.strip().startswith("build:"):
                 section = "build"
-            elif not line.startswith(" ") and not line.startswith("#"):
+            elif not line.startswith(" ") and not line.startswith("#") and line.strip():
```
mbargull: To avoid redundancy and be more conservative, can we use:
```diff
+            elif line and not line.startswith(" ") and not line.startswith("#"):
```
? Note that only the [space character is considered as indenation](http://www.yaml.org/spec/1.2/spec.html#id2777534) so we should use `line.lstrip(" ")`. But since we already test for `line.startswith(" ")`, the only missing case is the empty string.

bioconda_utils/lint_functions.py
```diff
@@ -301,10 +301,11 @@ def pin(recipe, meta, df):
...
                 in_requirements = False
                 section = None
             line = line.strip()
```
mbargull: For the sake of correctness, I think, we should also use
```python
line = line.lstrip(" ")
```
here.

test/test_linting.py
```diff
@@ -678,6 +678,20 @@ def test_lint_pin():
...
+          meta.yaml: |
+            package:
+              name: a
+              version: '0.1'
+
+            requirements:
+              build:
+                - zlib
+   
+              run:
```
mbargull: To avoid that someone who does some "beautifications" accidentally removes the purposely included empty line, maybe add a comment in the following line, e.g.
```diff
+              run:  # ^-- DO NOT remove the preceding line; it's part of the test case!
```

bioconda_utils/lint_functions.py
mbargull: Ah, I just now noticed the `line.strip()`s in the lines above. Let's just avoid that redundancy altogether and do a
```python
dedented_line = line.lstrip(" ")
```
and use `dedented_line` everywhere `line.strip()` is used.

bioconda_utils/lint_functions.py
```diff
@@ -295,18 +295,18 @@ def pin(recipe, meta, df):
...
                 section = "build"
-            elif not line.startswith(" ") and not line.startswith("#"):
+            elif dedented_line and not line.startswith(' ') and not line.startswith("#"):
                 in_requirements = False
                 section = None
-            line = line.strip()
-            if in_requirements and line.startswith('- {}'.format(dep_name)):
-                if pin_pattern.search(line) is None:
+            if in_requirements and dedented_line.startswith('- {}'.format(dep_name)):
+                if pin_pattern.search(line.strip()) is None:
```
mbargull: `dedented_line` instead of `line.strip()` should suffice as we're only searching a substring with the regex anyway

bioconda_utils/lint_functions.py
```diff
@@ -295,18 +295,18 @@ def pin(recipe, meta, df):
...
             if line.startswith("requirements:"):
                 in_requirements = True
-            elif in_requirements and line.strip().startswith("run:"):
+            elif in_requirements and dedented_line.startswith("run:"):
                 section = "run"
-            elif in_requirements and line.strip().startswith("build:"):
+            elif in_requirements and dedented_line.startswith("build:"):
                 section = "build"
-            elif not line.startswith(" ") and not line.startswith("#"):
+            elif dedented_line and not line.startswith(' ') and not line.startswith("#"):
```
mbargull: here, I'd actually use `line` to avoid confusion 😉

bioconda_utils/lint_functions.py
```diff
@@ -295,18 +295,18 @@ def pin(recipe, meta, df):
         not_pinned = set()
         pinned = set()
         for line in open(os.path.join(recipe, 'meta.yaml')):
+            dedented_line = line.lstrip(' ')
```
mbargull: Ah, I forgot, this also needs a
```python
line = line.rstrip("\n")
```
at the top of the loop body.

mbargull:
@bgruening, I added that `line.rstrip("\n")` I forgot before and rebased on `master`. Should be ready to merge once the CI tests finish.

johanneskoester:
Thanks @bgruening and @mbargull!

Issue #260, opened 2018-01-29
Fix gcc/perl issues == re-instate entrypoint in Docker image?


mbargull:
I took some time so get on track about the `perl`/`gcc` issues (https://github.com/conda-forge/perl-feedstock/pull/12, https://github.com/conda-forge/perl-feedstock/issues/16) with `defaults::gcc` and the `gcc` in `condaforge/linux-anvil`.
(@jerowe: Now I understand what you meant in https://github.com/bioconda/bioconda-recipes/pull/7410#issuecomment-361273181 ! 😄)

Now I'm (nearly) certain that https://github.com/bioconda/bioconda-utils/commit/2bfbbbf137513b51e2ae9dc8a33257c97c064fe1 could have caused the recent issues with `perl` recipes (https://github.com/bioconda/bioconda-recipes/issues/7415, https://github.com/bioconda/bioconda-recipes/pull/7408, ...).
It looks like it is crucial to use the image's entrypoint as it enables the `devtoolset-2` (put on `PATH`, etc., I assume), see: https://github.com/conda-forge/docker-images/blob/dae46482a0b87f4bc2ff7ad519dc9ecd76949be9/linux-anvil/entrypoint_source#L8.

@johanneskoester: I haven't yet looked at want you meant with "leads to an endless chown" in your commit message, so I'm only opening this issues, but have no fix, yet.

jerowe:
@mbargull , glad its clearer. Once you responded I realized my saying the toolchain was sourced was a very ambiguous statement. ;-)

If sourcing the toolchain is a problem we could always just preface the PATH? 

`export PATH=/opt/rh/devtoolset-2/root/usr/bin/:$PATH`

I have the feeling that if we don't add this in we are breaking binary compatibility with conda-forge. Specifically with perl things, but since conda-forge is a higher channel priority than defaults anything that is pulled in from there that is compiled will also be broken. I believe this included scientific python packages like numpy, scipy, matplotlib, but I haven't dug too deep into those.

mbargull:
Well, I hope that this "leads to an endless chown" Johannes observed might've only been a coincidence. If we could just re-use the entrypoint, that would certainly be the easiest way to retain compatibility.

mbargull:
@johanneskoester: What were the exact problems that entrypoint caused? Could/should we just try to reinstate it and see what happens?

johanneskoester:
[...]

PR #262, opened 2018-01-31
docker: enable devtoolset-2

mbargull:
Following up https://github.com/bioconda/bioconda-utils/issues/260#issuecomment-361892262:
This should activate the `devtoolset-2` and thus make the included `gcc` available.
It should mirror the behavior from https://github.com/conda-forge/docker-images/blob/92fedcfdb44ea530a712632cd4ccffbd4bebe0f8/linux-anvil/Dockerfile#L85 minus the `conda` user creation and `. /opt/conda/bin/activate root`. Would there be reason for us to want to activate `/opt/conda`?.
Also, it uses `tini` just because `condaforge/linux-anvil` does so too -- I'm not really familiar with all of that. Instead of using [`exec` with `gosu`](https://github.com/conda-forge/docker-images/blob/92fedcfdb44ea530a712632cd4ccffbd4bebe0f8/linux-anvil/entrypoint#L23) this does a plain `exec` as no `conda` user is created.
I haven't tested this out yet, so maybe a WIP. If someone wants to try it out, be my guest. I'll be able to do so only in a couple of hours.


mbargull:
Merged in #263 to avoid test failures.

docker-entrypoint
```diff
@@ -0,0 +1,15 @@
...
+# copied from condaforge/linux-anvil
+
+# Enable the compiler toolset.
+#
+# Must do this before activating conda or
+# the system Python interpreter is not used.
+# This is a problem for many reasons. One being
+# that the Python scripts used to activate the
+# toolset are not Python 3 compatible.
+. scl_source enable devtoolset-2
```
johanneskoester: To verify that this works, I think you should add a test case in our test suite that builds a recipe which fails if gcc from devtoolset-2 is not present.

mbargull: Yes, was going to do that. I need to run the image first (haven't done so, yet) to see what the best way to ensure that would be.

mbargull: The added test case checks for `gcc (GCC) 4.8.2 20140120 (Red Hat 4.8.2-15)` exactly. This might be overly specific, but this way we'll hopefully notice unintended changes right away.

mbargull:
This now does everything `condaforge/linux-anvil` does, minus the user creation.

bioconda_utils/docker_utils.py
```diff
@@ -198,6 +198,7 @@ def __init__(
         use_host_conda_bld=False,
         pkg_dir=None,
         keep_image=False,
+        image_build_dir=None,
```
mbargull: I had to add this to be able to test the repository's `Dockerfile`.
Since `Dockerfile` does `COPY . /tmp/repo`, those to-be-copied files have to exist in the image build directory. Previously that directory couldn't be prepopulated with any other files and only contained `Dockerfile` and `requirements.txt`.

daler: 👍

daler:
@mbargull this looks great! Nice test, too. Having the overly-specific gcc test might be a good canary for when things change in linux-anvil.

PR #263, opened 2018-01-31
test: pytest>=3.4 compat: capture >=INFO messages

mbargull:
`pytest 3.4`(just released on `conda-forge`) introduced a [breaking change](https://docs.pytest.org/en/latest/logging.html#incompatible-changes-in-pytest-3-4) so that `caplog` doesn't capture `INFO` messages on default in `test/test_utils.py::TestSubdags::test_subdags_more_than_recipes`, which thus failed.

### DIFF
--- a/pip-test-requirements.txt
+++ b/pip-test-requirements.txt
@@ -1,2 +1,2 @@
-pytest
+pytest>=3.3
 pytest-cov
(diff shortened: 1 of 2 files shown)


mbargull:
The CI did a good catch (=fail) on this one in another PR ;D

PR #264, opened 2018-02-03
lint: dont' require pin for run-only deps

mbargull:
The linter currently request to pin run-only dependencies. But if those dependencies are not required/directly pulled in during the build phase, the built package shouldn't have those run constraints.
(The second commit doesn't change behavior and is just to make the code a bit clearer (I hope).)


bgruening:
@mbargull can you include a test that would previously fail?

mbargull:
Ah, forgot to `git add` that; thanks for noticing!

bgruening:
LGTM, thanks for the test. @mbargull feel free to push to the other PR directly if you think something still needs to be fixed. I'm not sure when I will be back today.

Issue #256, opened 2018-01-26
Build system docs need to be updated for CircleCI


nsoranzo:
`docs/source/build-system.rst` still mentions Travis and related files.

xref. https://github.com/bioconda/bioconda-recipes/pull/7329

johanneskoester:
Solved by the referred PR.

PR #267, opened 2018-02-06
Update docs to reflect Circle CI changes.

johanneskoester:


peterjc:
Should solve #256 and https://github.com/bioconda/bioconda.github.io/issues/11 too I think.

tomkinsc:
Looks like `./simulate-travis.py` is still mentioned in section 2:
https://github.com/bioconda/bioconda-utils/pull/267/files#diff-f03f3ea7f67d3e74421ec0e6e90e89cdR32

PR #278, opened 2018-02-24
More updates to docs

epruesse:
You guys merged too fast :)

 - Fix parallel build
 - Add doc_url and dev_url links (only showing doc_url if different from home)
 - Document stuff & cleanup

### DIFF
--- a/docs/source/templates/readme.rst_t
+++ b/docs/source/templates/readme.rst_t
@@ -6,15 +6,21 @@
 
 {{ about.summary | escape }}
 
-======== ===========
-Home     {{ about.home }}
-Versions {{ versions.keys() | sort | join(", ") }}
-License  {{ about.license }}
-Recipe   {{ gh_recipes }}{{ recipe_path }}
-{% if extra.identifiers %}
-Links    {{ extra.identifiers | as_extlink | join(", ") }}
+============= ===========
+Home          {{ about.home }}
+Versions      {{ versions.keys() | sort | join(", ") }}
+License       {{ about.license }}
+Recipe        {{ gh_recipes }}{{ recipe_path }}
+{% if about.doc_url and about.doc_url != about.home %}
+Documentation {{ about.doc_url }}
 {% endif %}
-======== ===========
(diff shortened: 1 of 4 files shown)


docs/source/templates/readme.rst_t
```diff
@@ -6,15 +6,21 @@
...
-Home     {{ about.home }}
-Versions {{ versions.keys() | sort | join(", ") }}
-License  {{ about.license }}
-Recipe   {{ gh_recipes }}{{ recipe_path }}
+============= ===========
+Home          {{ about.home }}
+Versions      {{ versions.keys() | sort | join(", ") }}
+License       {{ about.license }}
+Recipe        {{ gh_recipes }}{{ recipe_path }}
+{% if about.doc_url and about.doc_url != about.home %}
```
bgruening: We could `rstrip('/')` here to be slightly more reliable.

docs/source/templates/recipes.rst_t
```diff
@@ -15,19 +15,19 @@ Available packages
...
+    {%- for r in recipes -%}
     <tr>
-        {% for k in keys %}
+        {%- for k in keys -%}
         <td>{{ r[k] }}</td>
-        {% endfor %}
+        {%- endfor -%}
     </tr>
-    {% endfor %}
+    {% endfor -%}
```
bgruening: `{%-`

epruesse: That's intentional, so that each row has its own line. Otherwise, the entire table becomes one line, making it rather hard to look at the HTML should one ever need to.

docs/source/templates/readme.rst_t
epruesse: Which one? Wouldn't it be better to use linting to enforce correct strings?

docs/source/templates/recipes.rst_t
epruesse: (For reference: the dash adjacent to the percent sign in jinja loop statements makes jinja drop white space in that direction)

PR #279, opened 2018-03-01
pkg_test: pin Miniconda version for mulled to avoid conda 4.4.10

mbargull:
fix https://github.com/bioconda/bioconda-recipes/pull/8044#issuecomment-369404165


mbargull:
The test failed locally (that is, without using `circleci` CLI client), but I don't really know why. Maybe it'll run fine on the CI...

mbargull:
Well, the undescriptiveness of that test failure is topnotch! :angry:

mbargull:
The created container has `/usr/local/conda-version.txt` but it's empty.. Why??

mbargull:
`/opt/conda/bin/conda --version` definitely writes to stdout.

mbargull:
Ah, it's not that the `mulled` test failed but the test from the build step, which obviously has another `conda` version.

mbargull:
Needs https://github.com/galaxyproject/galaxy-lib/pull/99 to work.

bioconda_utils/pkg_test.py
```diff
@@ -104,6 +110,8 @@ def test_package(
     base_image : None | str
         Specify custom base image. Busybox is used in the default case.
 
+    conda_image : None | str
+        Specify conda image to install the package with.
```
bgruening: `...conda docker image to install the package with during the mulled based tests.` ?

bioconda_utils/pkg_test.py
```diff
@@ -79,7 +84,8 @@ def test_package(
     name_override=None,
     channels=["conda-forge", "defaults"],
     mulled_args="",
-    base_image=None
+    base_image=None,
+    conda_image=MULLED_CONDA_IMAGE,
```
bgruening: cosmetic: leading comma

test/test_pkg_test.py
```diff
@@ -56,7 +57,9 @@ def _build_pkg(recipe):
     build.build(
         recipe=r.recipe_dirs['one'],
         recipe_folder='.',
-        env=env_matrix)
+        env=env_matrix,
+        mulled_test=mulled_test,
```
bgruening: same here, not sure if there is a lint rule for it :)

test/test_pkg_test.py
```diff
@@ -86,3 +89,31 @@ def test_pkg_test_custom_base_image():
...
+        one:
+          meta.yaml: |
+            package:
+              name: one
+              version: 0.1
+            test:
+              commands:
+                - '[ "${PREFIX%%/}" != "/usr/local" ] || cat "/usr/local/conda-version"'
+                - '[ "${PREFIX%%/}" != "/usr/local" ] || grep -F 4.3.11 "/usr/local/conda-version"'
+          post-link.sh: |
```
bgruening: Very clever! :+1:

bioconda_utils/pkg_test.py
mbargull: Yes, will update once we have a new `galaxy-lib` for this.

bioconda_utils/pkg_test.py
mbargull: Ohhh, I like my trailing commas :wink:. See, if it would have been there in the first place, the diff would only be one-line and only touch what was added (or deleted in another case). I think we have some mixed styles in here; some time in the future we can unify that and with some linter assistance :wink:

mbargull:
Huh, actually the test should have failed since
```bash
[ "${PREFIX%%/}" != "/usr/local" ] || grep -F 4.3.11 "/usr/local/conda-version"
```
needed to fail if setting `CONDA_IMAGE` didn't work. I'm confused.

test/test_pkg_test.py
mbargull: The `${PREFIX%%/}` didn't work because of the `${PREFIX}` string replacement in `pkg_test.get_tests`. Fixed with a new commit.

bioconda_utils/bioconda_utils-requirements.txt
```diff
@@ -3,7 +3,7 @@ argh=0.26.*
 beautifulsoup4=4.6.*
 conda=4.3.33
 conda-build=2.1.18
-galaxy-lib=17.9.*
+galaxy-lib >=18.5.4,<19.0a0
```
mbargull: `galaxy-lib 18.5.4` includes the `CONDA_IMAGE` configurability.
Do we need to care whether `pip` is fine with this syntax? As is, this is only used as input for `conda install --file`.

mbargull:
This is ready to go, I think.
As for the CI adjustments, those are the ones necessary to pick up the updated `galaxy-lib` dependency. I created a followup PR #282 for further adjustments, but those don't have to be part of this one.

bioconda_utils/bioconda_utils-requirements.txt
mbargull: Although technically unrelated, I increased the lower bound to `18.5.5` so that we can test out if https://github.com/galaxyproject/galaxy-lib/pull/100 works.

mbargull:
Before this gets merged, I want to add a test for https://github.com/galaxyproject/galaxy-lib/pull/100. I'll do that once I return from lunch.

mbargull:
Instead of a separate test, I just added the single quotes to `test_pkg_test_conda_image` since those Docker-based tests take ages to finish :-/

mbargull:
@bgruening are you fine with the changes I made after you approved previously?
(BTW, no need to "Update branch"/`git merge master` here, can be merged right away.)

bgruening:
@mbargull yes of course this looks great. Like the test. You rock!

PR #281, opened 2018-03-05
Fix for building rust recipes and a corresponding test case

johanneskoester:
For some reason, the SSL certificate bundle is not properly set to be recognized by cargo (the rust build tool). This PR tries to fix this in the docker template and adds a corresponding test case.

I still don't really understand what is going on here, because it works fine when testing our build env container manually...

### DIFF
--- a/bioconda_utils/docker_utils.py
+++ b/bioconda_utils/docker_utils.py
@@ -79,6 +79,9 @@ BUILD_SCRIPT_TEMPLATE = \
 #!/bin/bash
 set -e
 
+# Set SSL certificates
+export SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt
+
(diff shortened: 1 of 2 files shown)


mbargull:
This is due to a bug in `conda-build >=2.1.18,<3.0.10` that assigns empty values to `REQUESTS_CA_BUNDLE` and `SSL_CERT_FILE` instead of having those variables completely unset.
This has caused trouble with Rust and Julia packages thus far. For now we fixed that in the recipes' `build.sh`:
https://github.com/bioconda/bioconda-recipes/pull/7818/files#diff-328a10f3007ed83abc4579d9bcf913e6R3
https://github.com/bioconda/bioconda-recipes/pull/7973/files#diff-732a32d76d1f4cd81ac3d4604382952eR3

PR #283, opened 2018-03-07
clarify patch policy

daler:
xref https://github.com/bioconda/bioconda-recipes/issues/8183


daler:
Yes, thanks @charlesreid1!

PR #292, opened 2018-04-20
Fix identifier parsing

johanneskoester:


johanneskoester:
Fix for novoalign recipe is in bioconda/bioconda-recipes#8699

johanneskoester:
Build docs was just cancelled because it kept an open SSH connection, but it passed.

Issue #284, opened 2018-03-19
Conda-Build 3 support


johanneskoester:
1. [x] Update pinning to conda build 3
2. [x] Replace our env matrix by https://github.com/conda-forge/conda-forge-pinning-feedstock
3. [x] add test case for new `outputs:` feature of CB3 meta.yaml
----
3. [ ] update docs to include the new compiler polices
4. [ ] adopting rules to include noarch packages
----
5. [x] point out `host:`/`build:` section differences in docs
6. [ ] Adjust/add lint functions
   1. [ ] replace `build:` by `host:` when applicable
   2. [ ] replace `build: [gcc]` by `build: {{ compiler("c") }}` or `build: {{ compiler("cxx") }}`
   3. [ ] allow `source:` to be a list (@bgruening: I assume `cargo-port` will need similar adjustments)
   5. [ ] in the docs and the lint explanation, explain how to use the new compilers with the old ABI if necessary (e.g. to stay compatible with conda-forge that might be slower in rebuilding with the new compilers).
   6. [ ] prohibit `fn:` in `source:`, now redundant if `url:` points to file (which it really should)

bgruening:
I added two other points, feel free to remove them.

mbargull:
Sorry for the delay. Added some more points. Will get this done in the next days.

kyleabeauchamp:
I made a PR with a prototype BWA recipe upgraded for CB3:

https://github.com/bioconda/bioconda-recipes/pull/8737

kyleabeauchamp:
Then there's the question of how to go about updating the pins (e.g., do we recommend that authors manually change the pins at their convenience and increment the build number).

epruesse:
CB3 seems to work :)

PR #293, opened 2018-04-23
Migrate to conda-build 3.

johanneskoester:
This is a start to finally do the migration to CB3. Feel free to contribute, as my time is quite limited at the moment (sorry).

xref: #284

* [x] adapt bioconda-utils
* [x] update docs as suggested in #284
* [x] try to implement all lints from #284


johanneskoester:
@mbargull in case you already have some local changes for CB3 or just your conceptual work, please move that into this PR.

johanneskoester:
Seems we are running into this bug: https://github.com/conda/conda/issues/6691

johanneskoester:
@bgruening the restriction of the galaxy-lib version (`<18.6.0a0`) triggers a bug in the latest conda package (see above). Is the restriction crucial?

bgruening:
No and if so, I will fix galaxy-lib.

johanneskoester:
Update: this works fine. I am now migrating to the new way of handling pinnings.

kyleabeauchamp:
Thanks for tackling this one!

daler:
@johanneskoester I think I figured out the remaining failures. It stems from [using `--skip-existing`](https://github.com/bioconda/bioconda-utils/pull/293/files#diff-877ea8e79f39898f92d9e5f6dea153fdR92) when passed to the docker container during tests.

I believe the problem arises because the test fixtures that build packages (e.g, [`multi_build`](https://github.com/bioconda/bioconda-utils/pull/293/files#diff-85c7da7d97a696e8017da5742a876475R128)) are scoped to the module level and so do not clean up until after all tests consuming them have completed. The packages built by these fixtures live in the host's conda-bld directory. That conda-bld dir is mapped to the docker container, and added as a local channel inside the container. When conda-build runs in the container it sees the package in one of its channels and because of `--skip-existing`, doesn't re-build the package in the container's conda-bld dir (which is separate from the host's mapped conda-bld dir).

So the failure -- at least for the multi_build with docker -- is because the container is trying to copy a package filename that wasn't created due to being skipped.

Possible solutions:

1. Don't use `--skip-existing`
    - this is the easiest fix
    - not sure of the implications (and it looks like you added this in recently)

2. Change the build script used by docker to first check if a package was built before trying to copy it over.
    - while this would let the tests pass, it's not actually testing correctly-built packages
    - we would need to rewrite tests to test for gracefully skipping existing recipes as well as separate tests that ensure a package is gone before running the test

3. Refactor tests to NOT use module-scoped fixtures, to force a new built recipes for each test
    - this will extend the testing time (not sure by how much)
    - this may be the best way of isolating tests from each other

daler:
(I temporarily removed `--skip-existing` in ff204fe  to verify it fixes the circleci tests)

johanneskoester:
The skip existing is needed because we otherwise will build all variants although they might be already available in one of the channels. Thanks for digging into this. I think we can solve that by passing the pkg_paths to the container script. I will take care of that later today.

daler:
Ah, I see. I'm not sure pkg_paths will solve this issue, but it would great if that works. In parallel, I'll try get a test refactor working locally that we can use as a fallback in case providing pkg_paths doesn't work.

By the way, thanks so much for working on this migration!

johanneskoester:
@daler problem solved. What was really missing is that I did not pass `--skip-existing` to the `conda build --output` command when copying files. So, what should be the strategy for merging? There are some additional lints to be written (#284), but these are all not mandatory. We could merge now and gradually add the missing documentation, or also update all docs and lints in this PR.

daler:
@johanneskoester nice!

I think we should at least get docs in place to point folks to when there's issues or confusion (which I think is unavoidable with a change like this). I can write them. I'm not that familiar myself with cb3, but this will be my chance to learn it. In the meantime we could add also the bulk of the lints. The lints would be useful so there is an obvious warning that the system is a bit different now.

johanneskoester:
Ok, sounds like a plan. I will be busy until next tuesday, so if you get the chance to work on this, I will see what is left by tuesday and go on from there. Thanks!

bioconda_utils/bioconda_utils-requirements.txt
```diff
@@ -21,3 +21,4 @@ colorlog=3.1.*
 six=1.11.*
 alabaster=0.7.*
 git=2.14.*
+conda-forge-pinning
```
daler: @johanneskoester should we pin this so that we can maintain control over when to move to the next set of variants?

daler:
@johanneskoester are we planning on using the `{{ compiler('c') }}` constructs? If so we need to add the `anaconda` channel (and update docs, and coordinate with bioconda-recipes setup). Looks like the compliers aren't on defaults or conda-forge.

kyleabeauchamp:
In the previous ticket (https://github.com/bioconda/bioconda-utils/issues/284) I believe we planned on using those constructs, right?

daler:
@kyleabeauchamp ah yes thank you, I hadn't revisited that issue in a while and forgot it was on the list.

docs/source/cb3.rst
```diff
@@ -0,0 +1,208 @@
...
+
+.. _global-pinning:
+
+Global pinning
+~~~~~~~~~~~~~~
+
+**Summary:**
+
+- Previously we pinned packages using the syntax ``- zlib {{ CONDA_ZLIB }}*``
+- Instead, we should now pin packages with the syntax ``- zlib {{ zlib }}``.
```
bgruening: I don't think we should do this. We should wait until zlib and co have run-exports defined and work in the meantime to get ours as well supported for run_exports (https://github.com/bioconda/bioconda-recipes/pull/8831). This would mean don't have to define zlib at all as run-time dependency.

test/test_linting.py
```diff
@@ -852,3 +772,34 @@ def test_invalid_identifiers():
...
+            meta.yaml: |
+              package:
+                name: a
+                version: 0.1
+              requirements:
+                build:
+                  - numpy
+                  - python
+                run:
+                  - numpy
```
bgruening: I think the new way specifically for numpy is:
```
requirements:
        build:
             - numpy
             - python
        run:
             - {{ pin_compatible('numpy') }}
```

test/test_utils.py
```diff
@@ -960,11 +783,166 @@ def test_build_container_default_gcc(tmpdir):
...
+    r = Recipes(
+        """
+        one:
+          meta.yaml: |
+            package:
+              name: one
+              version: 0.1
+            requirements:
+              run:
+                - htslib {{ htslib }}
```
bgruening: @daler from where do you have this `{{ htslib }}` syntax? It should work without it, does it not?

test/test_utils.py
```diff
@@ -960,11 +783,166 @@ def test_build_container_default_gcc(tmpdir):
...
+    r = Recipes(
+        """
+        one:
+          meta.yaml: |
+            package:
+              name: one
+              version: 0.1
+            requirements:
+              run:
+                - zlib {{ zlib }}
```
bgruening: We should aiming for not including zlib at all at run-time. This can be archived with run_exports, see here for example: https://github.com/bioconda/bioconda-recipes/pull/8831

However, this means we need to wait a little bit longer -  conda-forge is working on this atm.

bgruening:
@johanneskoester @daler awesome work, I added a few comments.
Concerning the new compilers: If we decide to use them from this PR onwards, which I agree, we should wait a few more weeks. On conda-forge we made the experience that a few new compiled packages where incompatible with old once. So conda-forge is currently rebuilding everything with the new compilers (on anaconda-build-bots afaik). As soon as this goes online, we should be fine re-compiling our entire stack with the new compilers.

bioconda_utils/pkg_test.py
```diff
@@ -148,6 +149,13 @@ def test_package(
         cmd += ['--name-override', name_override]
     cmd += channel_args
     cmd += shlex.split(mulled_args)
+
+    # galaxy-lib always downloads involucro, unless it's in cwd or its path is explicitly given.
+    # TODO: This should go into galaxy-lib. Once it is fixed upstream, remove this here.
```
mbargull: xref: https://github.com/galaxyproject/galaxy-lib/issues/110

.circleci/setup.sh
```diff
@@ -28,6 +28,9 @@ if [[ ! -d $WORKSPACE/miniconda ]]; then
     bash miniconda.sh -b -p $WORKSPACE/miniconda
 
     # step 2: setup channels
+
+    # Added to support the new compiler packages from Anaconda 5
+    conda config --system --add channels anaconda
```
bgruening: I think we can remove the default channel here. anaconda is default + extra afaik or even a symlink.

mbargull: We should **not** use the `anaconda` channel. That channel should not be used under normal circumstances. IIRC, this is just a mirror (or intermediate source? -- can't remember) of `defaults` (or rather the `pkgs/main` part, I think).
`defaults` has everything we need, including the compilers.

mbargull: The default compilers used by `conda-build 3` should be packages from:

<details><summary>

```
$ for s in linux-64 osx-64 win-64; do conda search "*_${s}[subdir=${s}]" ; done
```

</summary>

```                                                                                                                             
Loading channels: done
# Name                  Version           Build  Channel             
binutils_impl_linux-64          2.28.1      h04c84fa_2  pkgs/main           
binutils_impl_linux-64          2.28.1      had2808c_3  pkgs/main           
binutils_linux-64          2.28.1      he4fe6c7_1  pkgs/main           
binutils_linux-64           5.4.0     had2808c_24  pkgs/main           
binutils_linux-64           7.2.0              19  pkgs/main           
binutils_linux-64           7.2.0              24  pkgs/main           
binutils_linux-64           7.2.0              25  pkgs/main           
binutils_linux-64           7.2.0              26  pkgs/main           
binutils_linux-64           7.2.0     hc67e822_10  pkgs/main           
binutils_linux-64           7.2.0     hc67e822_11  pkgs/main           
binutils_linux-64           7.2.0     hc67e822_12  pkgs/main           
binutils_linux-64           7.2.0     hc67e822_13  pkgs/main           
binutils_linux-64           7.2.0     hc67e822_14  pkgs/main           
binutils_linux-64           7.2.0     hc67e822_15  pkgs/main           
binutils_linux-64           7.2.0      hc67e822_2  pkgs/main           
binutils_linux-64           7.2.0      hc67e822_3  pkgs/main           
binutils_linux-64           7.2.0      hc67e822_7  pkgs/main           
binutils_linux-64           7.2.0      hc67e822_8  pkgs/main           
binutils_linux-64           7.2.0      hc67e822_9  pkgs/main           
duma_linux-64            2.5.15      h28ad9a3_3  pkgs/main           
duma_linux-64            2.5.15      h49555fe_2  pkgs/main           
duma_linux-64            2.5.15      hec2b5b5_1  pkgs/main           
gcc-dbg_linux-64           7.2.0              19  pkgs/main           
gcc-dbg_linux-64           7.2.0              24  pkgs/main           
gcc-dbg_linux-64           7.2.0              25  pkgs/main           
gcc_impl_linux-64           5.4.0      habb00fd_3  pkgs/main           
gcc_impl_linux-64           7.2.0      habb00fd_3  pkgs/main           
gcc_impl_linux-64           7.2.0      hc5ce805_2  pkgs/main           
gcc_linux-64              5.4.0     h98af8de_24  pkgs/main           
gcc_linux-64              7.2.0              19  pkgs/main           
gcc_linux-64              7.2.0              24  pkgs/main           
gcc_linux-64              7.2.0              25  pkgs/main           
gcc_linux-64              7.2.0              26  pkgs/main           
gcc_linux-64              7.2.0     h196ecd4_10  pkgs/main           
gcc_linux-64              7.2.0     h217957b_13  pkgs/main           
gcc_linux-64              7.2.0      h22f992b_7  pkgs/main           
gcc_linux-64              7.2.0      h29fd5c9_3  pkgs/main           
gcc_linux-64              7.2.0     h4118e58_11  pkgs/main           
gcc_linux-64              7.2.0      h60973fd_9  pkgs/main           
gcc_linux-64              7.2.0      h6f34251_8  pkgs/main           
gcc_linux-64              7.2.0     haf1f6fa_15  pkgs/main           
gcc_linux-64              7.2.0      hc7b1ceb_1  pkgs/main           
gcc_linux-64              7.2.0     hd763dfe_12  pkgs/main           
gcc_linux-64              7.2.0      hecb3f9c_2  pkgs/main           
gcc_linux-64              7.2.0     hf1c97a4_14  pkgs/main           
gdb_linux-64             7.12.1      h0d70bb3_3  pkgs/main           
gdb_linux-64             7.12.1      hd7315ee_1  pkgs/main           
gdb_linux-64             7.12.1      hf9d566d_2  pkgs/main           
gfortran-dbg_linux-64           7.2.0              19  pkgs/main           
gfortran-dbg_linux-64           7.2.0              24  pkgs/main           
gfortran-dbg_linux-64           7.2.0              25  pkgs/main           
gfortran_impl_linux-64           5.4.0      hdf63c60_3  pkgs/main           
gfortran_impl_linux-64           7.2.0      hb3c8cce_2  pkgs/main           
gfortran_impl_linux-64           7.2.0      hdf63c60_3  pkgs/main           
gfortran_linux-64           5.4.0     h98af8de_24  pkgs/main           
gfortran_linux-64           7.2.0              19  pkgs/main           
gfortran_linux-64           7.2.0              24  pkgs/main           
gfortran_linux-64           7.2.0              25  pkgs/main           
gfortran_linux-64           7.2.0              26  pkgs/main           
gfortran_linux-64           7.2.0      h0d86fa5_2  pkgs/main           
gfortran_linux-64           7.2.0      h36698f7_3  pkgs/main           
gfortran_linux-64           7.2.0     h3a66261_14  pkgs/main           
gfortran_linux-64           7.2.0      h6c08c0b_8  pkgs/main           
gfortran_linux-64           7.2.0      h6fbc3aa_9  pkgs/main           
gfortran_linux-64           7.2.0      h939f3e1_1  pkgs/main           
gfortran_linux-64           7.2.0     h9b3fb79_10  pkgs/main           
gfortran_linux-64           7.2.0     ha0c5ce7_12  pkgs/main           
gfortran_linux-64           7.2.0     hbff877f_13  pkgs/main           
gfortran_linux-64           7.2.0      hc3a7fdf_7  pkgs/main           
gfortran_linux-64           7.2.0     hc6796cf_11  pkgs/main           
gfortran_linux-64           7.2.0     hd729bf5_15  pkgs/main           
gxx-dbg_linux-64           7.2.0              19  pkgs/main           
gxx-dbg_linux-64           7.2.0              24  pkgs/main           
gxx-dbg_linux-64           7.2.0              25  pkgs/main           
gxx_impl_linux-64           5.4.0      hdf63c60_3  pkgs/main           
gxx_impl_linux-64           7.2.0      hd3faf3d_2  pkgs/main           
gxx_impl_linux-64           7.2.0      hdf63c60_3  pkgs/main           
gxx_linux-64              5.4.0     h98af8de_24  pkgs/main           
gxx_linux-64              7.2.0              19  pkgs/main           
gxx_linux-64              7.2.0              24  pkgs/main           
gxx_linux-64              7.2.0              25  pkgs/main           
gxx_linux-64              7.2.0              26  pkgs/main           
gxx_linux-64              7.2.0     h2e50c1c_15  pkgs/main           
gxx_linux-64              7.2.0      h33f4dac_3  pkgs/main           
gxx_linux-64              7.2.0      h3aab023_7  pkgs/main           
gxx_linux-64              7.2.0     h41d9e3f_12  pkgs/main           
gxx_linux-64              7.2.0     h4bc98f9_14  pkgs/main           
gxx_linux-64              7.2.0      h57110d4_9  pkgs/main           
gxx_linux-64              7.2.0     h5e79354_11  pkgs/main           
gxx_linux-64              7.2.0      h60a51f1_2  pkgs/main           
gxx_linux-64              7.2.0     h7798672_13  pkgs/main           
gxx_linux-64              7.2.0      h8cc0f3c_1  pkgs/main           
gxx_linux-64              7.2.0     hc01b44c_10  pkgs/main           
gxx_linux-64              7.2.0      he93fcd3_8  pkgs/main           
Loading channels: done
# Name                  Version           Build  Channel             
clang-dbg_osx-64           4.0.1      h1ce6c1d_5  pkgs/main           
clang-dbg_osx-64           4.0.1      h1ce6c1d_7  pkgs/main           
clang-dbg_osx-64           4.0.1      h1ce6c1d_8  pkgs/main           
clang_osx-64              4.0.1               9  pkgs/main           
clang_osx-64              4.0.1     h1ce6c1d_10  pkgs/main           
clang_osx-64              4.0.1      h1ce6c1d_5  pkgs/main           
clang_osx-64              4.0.1      h1ce6c1d_7  pkgs/main           
clang_osx-64              4.0.1      h1ce6c1d_8  pkgs/main           
clang_osx-64              4.0.1      h5d4359d_1  pkgs/main           
clang_osx-64              4.0.1      h60dbf0d_2  pkgs/main           
clang_osx-64              4.0.1      h6c06975_3  pkgs/main           
clang_osx-64              4.0.1      h9d74b7d_5  pkgs/main           
clang_osx-64              4.0.1      hd786faf_0  pkgs/main           
clangxx-dbg_osx-64           4.0.1               5  pkgs/main           
clangxx-dbg_osx-64           4.0.1               7  pkgs/main           
clangxx-dbg_osx-64           4.0.1      h22b1bf0_8  pkgs/main           
clangxx_osx-64            4.0.1               5  pkgs/main           
clangxx_osx-64            4.0.1               7  pkgs/main           
clangxx_osx-64            4.0.1               9  pkgs/main           
clangxx_osx-64            4.0.1      h04dcba0_2  pkgs/main           
clangxx_osx-64            4.0.1     h22b1bf0_10  pkgs/main           
clangxx_osx-64            4.0.1      h22b1bf0_8  pkgs/main           
clangxx_osx-64            4.0.1      h22b6db4_0  pkgs/main           
clangxx_osx-64            4.0.1      h557b0ca_5  pkgs/main           
clangxx_osx-64            4.0.1      h633f504_1  pkgs/main           
clangxx_osx-64            4.0.1      hbbb0622_3  pkgs/main           
gfortran_osx-64           4.8.5      h044dcb3_2  pkgs/main           
gfortran_osx-64           4.8.5      h22b1bf0_3  pkgs/main           
gfortran_osx-64           4.8.5      h878a615_0  pkgs/main           
gfortran_osx-64           4.8.5      hd9f5454_1  pkgs/main           
Loading channels: done
# Name                  Version           Build  Channel             
intel-fortran_win-64          13.1.6      h09e79d1_0  pkgs/main           
intel-fortran_win-64          16.0.4      hc0ab4e9_0  pkgs/main           
intel-fortran_win-64        2017.0.4      h739a657_0  pkgs/main           
vs2008_win-64      9.00.30729.1      hb7fc0b5_0  pkgs/main           
vs2008_win-64      9.00.30729.1      hb7fc0b5_1  pkgs/main           
vs2015_win-64        14.0.25123      h17c34da_3  pkgs/main           
vs2015_win-64        14.0.25123      hf5d5b87_1  pkgs/main           
vs2015_win-64        14.0.25123      hf5d5b87_2  pkgs/main           
vs2017_win-64            15.5.2               0  pkgs/main           
vs2017_win-64            15.5.2               1  pkgs/main           
```
</details>

daler: If I remove the `anaconda` channel, and try building this recipe:

```yaml
package:
  name: one
  version: 0.1
requirements:
  build:
    - {{ compiler('c') }}
  host:
    - python
  run:
    - python
```

I get the error `conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-64: {"libgcc-ng[version='>=4.9']"}
`. However if I re-enabling the `anaconda` channel lets the recipe build. I must be doing something wrong, but not sure what.

test/test_utils.py
daler: Thanks Bjoern. I approached this entirely from reading the conda docs. It wasn't clear to me from the docs that's what `run_exports` is doing -- that's really nice!

test/test_utils.py
daler: Thanks, yes it seems to work. Was not clear from the docs that this would be true. I was following the lead of [this example](https://conda.io/docs/user-guide/tasks/build-packages/variants.html#general-pinning-examples) where they specify `boost {{ boost }}`. Is that syntax just for when there are multiple variants specified?

docs/source/cb3.rst
daler: After reading the `run_exports` docs some more (and your other comments below), let me completely rewrite those docs!

.circleci/setup.sh
daler: It was, but looking closer I found the problem. 

Due to a longstanding issue with the solver hanging on so many duplicated R packages across channels, I had edited my `~/.condarc` according to this as-yet-unresolved issue: https://github.com/conda/conda/issues/5536:

```yaml
default_channels:
  - https://repo.continuum.io/pkgs/free
```

Changing those lines to:

```yaml
default_channels:
  - https://repo.continuum.io/pkgs/free
  - https://repo.continuum.io/pkgs/main
```

solves the problem. However, since I run everything in the separate "bootstrapped" bioconda-utils miniconda installation, I guess I just learned that my home directory's `.condarc` gets merged into any conda env I use. Good to know.

docs/source/cb3.rst
daler: @bgruening just pushed some updates/fixes to this section. Any better?

.circleci/setup.sh
mbargull: Alternatively, you could also try:
```
conda config --add pinned_packages 'r-base >=3.4'
```
, xref https://github.com/conda/conda/issues/5536#issuecomment-386935346.

----
`~/.condarc` (and others, see: https://github.com/conda/conda/blob/4.5.2/conda/base/constants.py#L25-L43) are pesky little beasts that I always (kind of) deactivate via
```
mkdir -p ~/.conda/condarc.d
touch ~/.conda/condarc ~/.conda/.condarc ~/.condarc
chmod a-w ~/.conda/condarc.d ~/.conda/condarc ~/.conda/.condarc ~/.condarc
```
before `conda` gets a chance to create them.
Sure, they are helpful for end users in some cases, but I find to use `conda config --env` and `conda config --system` much less headache-inducing when operating multiple Conda installations...

docs/source/cb3.rst
```diff
@@ -169,38 +170,94 @@ Global pinning
 **Summary:**
 
 - Previously we pinned packages using the syntax ``- zlib {{ CONDA_ZLIB }}*``
-- Instead, we should now pin packages with the syntax ``- zlib {{ zlib }}``.
+  in both the ``build`` and ``run`` dependencies.
+- Instead, we should now just specify package names in the ``host`` section,
+  e.g., as simply ``zlib``. They are pinned automatically. No need to add them
+  to run dependencies, as they will be added automatically.
```
mbargull: > No need to add them to run dependencies, as they will be added automatically.

This is only true if the package, e.g., `zlib`, is built with `run_exports` **and** it is ensured that the package build with that `run_exports` is used in `host`. For a package like `zlib`, which has few (=only one) dependencies the following is unlikely to happen, but imagine the following case:
```yaml
# conda_build_config.yaml
upstream: 1.0
pin_run_as_build:
  upstream:
    max_pin: x.x
```
```yaml
package:
  name: downstream
requirements:
  host:
    - upstream
    - other-package <2.0
```
```yaml
package:
  name: upstream
  version: 1.0
build:
  number: 0
requirements:
  run: []
```
```yaml
package:
  name: upstream
  version: 1.0
build:
  number: 1
  run_exports:
    - {{ pin_subpackage('upstream', max_pin='x.x') }}
  requirements:
    run:
      - other-package >=2.0
```

Even though we have a build of `upstream 1.0` that provides `run_exports`, `conda`'s solver will pick up the older build without `run_exports` due to a dependency conflict (and since `conda` doesn't always use the builds with the highest numbers :unamused:). Hence, `downstream` will not have that dependency added to `requirements/run`!
If, as a backup to those cases, we have `pin_run_as_build` and a plain `- upstream` dependency in `requirements/run`, we can avoid that case.
Thus, I think we should recommend
```yaml
package:
  name: downstream
requirements:
  host:
    - upstream
    - other-package <2.0
  run:
    - upstream
```
over
```yaml
package:
  name: downstream
requirements:
  host:
    - upstream
    - other-package <2.0
```

mbargull: Also note that `conda-forge-pinning` only specifies `zlib: 1.2`. Hence, we can create a real world example:
```yaml
package:
  name: downstream
requirements:
  host:
    - bwa 0.7.16
    - zlib
```
=> `bwa 0.7.16` requires `zlib 1.2.8.*` and hence prevents `zlib 1.2.11` (which has \[on `defaults`\] or will have \[on `conda-forge`\] `run_exports`) to be picked up and instead will use `zlib 1.2.8`. Thus, without `requirements: run: ['zlib']` we won't get the `zlib` run requirement.

test/test_linting.py
johanneskoester: We use the conda-forge feedstock for the pinning (i.e. pinning at the variant level). Hence, I think this is needed here:

```
requirements:
  host:
    - numpy {{ numpy }}
  run:
    - numpy {{ numpy }}
```

The `pin_compatible` seems to be only needed when overwriting the variants matrix on a per-recipe base.

docs/source/cb3.rst
johanneskoester: I agree, packages that are needed as run dependencies should always be added to the run section. This also improves readability, as you immediately see what will be used at runtime.

johanneskoester:
@daler, we have been using the new compilers with our channel setup already. I thought that the anaconda channel is part of defaults, hence, this should already work without changing the channel setup.

docs/source/cb3.rst
bgruening: Imho we should do what conda-forge is doing here. They are currently integrating the run_dependencies and updating the pinnings accordingly. I don't follow the `bwa` example, as we need to rebuild against bwa either way, against new compiler and against new zlib. Only if we do this, we remove the run dependency. Or do you mean that bwa as tool needs 1.2.8 of zlib? In this case we could 1) rebuild old zlib, 2) patch bwa or 3) make an exception.
While I agree with Johannes that this magic is bad for readability, I see the libgcc (strong run_export) mess we had over the years finally solved with this and I prefer adapting conda-forge strategy for consistency.

Please note that it is not yet entirely clear what conda-forge will be doing and these discussions are all ongoing so if we really disagree we should start discussing in the conda-forge channel.

mbargull: Read the `bwa 0.7.16` as "some random, possibly transitively, additional dependency that causes `conda` to pick up some older `zlib` version without `run_exports`".
I also agree with the readability aspect but this is not the reason why I think the explicit `run` requirements are needed. Only relying on the implicit `run_exports` and `conda` not picking up older versions or builds (within the version constraints given by the package keys in the `conda_build_config.yaml` files) is too brittle. @bgruening, can you point me to some conda-forge discussion where the `run_exports`-only strategy is proposed or discussed? I assume they are aware of the issue I'm pointing at, but if not completely, I'd like to chime in.

test/test_linting.py
mbargull: We should be fine with 
```yaml
  host:
    - numpy
```
instead of
```yaml
  host:
    - numpy {{ numpy }}
```
(or rather
```yaml
  host:
    - numpy {{ numpy }}.*
```
if `numpy` is given as a plain version as in `conda-forge-pinning`) since the top-level keys in the `conda_build_config.yaml` files act two-fold: For one, they can be used as arbitrary Jinja2 variables like with `numpy {{ numpy }}.*` where they just get replaced by the given value (for the currently enumerated build variant). Secondly, they are interpreted as versions (or rather Conda match specifications, I think) for packages with the same name in `build`/`host` requirements.

As for `pin_compatible('numpy')`, this is a special case since `numpy` deliberately does not define `run_exports` in https://github.com/AnacondaRecipes/numpy-feedstock/blob/master/recipe/meta.yaml (which I expect `conda-forge` to adapt, too) and also has no `pin_run_as_build` entry. Hence, a plain `numpy` requirement will remain a plain `numpy`.
Although I said earlier that we should avoid `pin_compatible` on a per-recipe basis, with `numpy` we have to add it, though, due to the lacking `run_exports`/`pin_run_as_build`.
(The reason for leaving out `run_exports`/`pin_run_as_build` for `numpy` is that it often does not require the pinning (= if used as a pure Python dependency) and only sometimes requires it (= if a package links to `numpy`'s C-API).)

mbargull:
@johanneskoester, the compiler problem @daler had was just due to some sneaky `~/.condarc` that had `pkgs/main` (i.e., the "newer" part of `defaults` that includes the compilers) removed, see https://github.com/bioconda/bioconda-utils/pull/293#discussion_r186310547.

johanneskoester:
@daler how close are we in terms of docs and lints?

daler:
Docs mostly done. Haven't heard any updates on our above discussion on whether to always include `run:` dependencies even though it might be redundant with `run_exports` for a package also listed in `host:` deps.

Since it looks like everyone's leaning towards keeping deps in both `host` and `run`, I'll update the docs accordingly.

The following lints from #284 I'm not sure about, if you could take care of them:

- *replace build: by host: when applicable* I'm not sure all the cases that are applicable
 - *allow source: to be a list (@bgruening: I assume cargo-port will need similar adjustments)* I'm not sure about policy decisions on this, specifically under what conditions we should encourage `source:` to be a list. 
- *in the docs and the lint explanation, explain how to use the new compilers with the old ABI if necessary (e.g. to stay compatible with conda-forge that might be slower in rebuilding with the new compilers).* I'm not fully clear on under what conditions we need new compilers with old ABI.

I can do these:

- *replace build: [gcc] by build: {{ compiler("c") }} or build: {{ compiler("cxx") }}* 
- *prohibit fn: in source:, now redundant if url: points to file (which it really should)* 

I will also add a `noarch` test and bring back the linting and docs for `noarch`.

daler:
@johanneskoester now the only lints/docs remaining are:

> * replace `build:` by `host:` when applicable 
> *  allow `source:` to be a list 
> * in the docs and the lint explanation, explain how to use the new compilers with the old ABI if necessary (e.g. to stay compatible with conda-forge that might be slower in rebuilding with the new compilers).

johanneskoester:
* @daler I think we can skip the lint for the host section. As I read the conda docs, conda build will simply fail if things are specified wrongly. I am also unsure how a lint would look like, because, e.g., python can be both a build tool and a host dependency, so it is not really possible to come up with a white or blacklist here.
* is there a use case for source to be a list?
* Regarding the old ABI thing, we could probably wait and see if issues arise?

daler:
> @daler I think we can skip the lint for the host section. As I read the conda docs, conda build will simply fail if things are specified wrongly. I am also unsure how a lint would look like, because, e.g., python can be both a build tool and a host dependency, so it is not really possible to come up with a white or blacklist here.

OK, that was my conclusion too.

> is there a use case for source to be a list?

Not that I know of.

> Regarding the old ABI thing, we could probably wait and see if issues arise?

Agreed.

@mbargull @bgruening do you have anything else you'd like to add? I think another read-through of the docs in `cb3.rst` would help to make sure I've written things accurately. Other than that, I think it might be ready to try out, and we can hotfix issues as they arise.

bgruening:
I will read through it tomorrow, sorry not much time here at the moment.

mbargull:
>* @daler I think we can skip the lint for the host section. As I read the conda docs, conda build will simply fail if things are specified wrongly. I am also unsure how a lint would look like, because, e.g., python can be both a build tool and a host dependency, so it is not really possible to come up with a white or blacklist here.

I'm a little on the fence on this one. Currently, I think *at some point* we would want to introduce a lint function that requires `host:` to be always present. If we make `host:` mandatory (even if empty), we could ensure more uniform build setups, i.e., always have two environments (`build` and `host`) activated. Currently, for backwards compatibility, `conda-build` will merge them in certain cases, e.g., if no `host:` is given or no `{{ compiler(...) }}` is used.
I think the only advantage of not adding such lint function right away is that we don't have to change every recipe just now (which is a weak argument, TBH).
>* is there a use case for source to be a list?

Yes, but it only affects a small number of recipes, essentially all recipes that use `curl` or `wget` during the build to circumvent the previous limitation to only have a single source. I made the (small) adjustments needed for this.
>* Regarding the old ABI thing, we could probably wait and see if issues arise?

I agree. Also, I'm not sure, but I could imagine these issues to need solutions on a case by case-basis.

.ropeproject/config.py
```diff
@@ -0,0 +1,112 @@
+# The default ``config.py``
```
mbargull: Was the addition of this file intentional? (Don't mind, don't care, just making sure :wink:)

bioconda_utils/docker_utils.py
```diff
@@ -134,6 +120,7 @@
 """
 FROM bioconda/bioconda-utils-build-env
 {self.proxies}
+RUN /opt/conda/bin/conda install -y conda={conda_ver} conda-build={conda_build_ver}
```
mbargull: This is fine for now. I think it would make sense to create a `bioconda-build-setup` (or similar) package at some point which consists of `bioconda_utils-conda_build_config.yaml` and has requirements like
```yaml
requirements:
  run:
    - conda 4.5.4
    - conda-build 3.10.2
    - conda-forge-pinning 2018.05.07
```
and possibly a `post-link.sh` with
```bash
"${CONDA_EXE}" config --env --add channels defaults --add channels conda-forge --add channels bioconda
```
or something.
That way, we could just do a `conda create bioconda-build-setup=version` and don't have to juggle with copying those configuration files manually

bioconda_utils/lint_functions.py
```diff
@@ -85,268 +71,286 @@ def _has_preprocessing_selector(recipe):
...
     Does the package exist in any other non-bioconda channels?
     """
-    results = _subset_df(recipe, meta, df)
-    channels = set(results.channel).difference(['bioconda'])
-    if len(channels):
-        return {
-            'exists_in_channels': channels,
-            'fix': 'consider deprecating',
-        }
+    for meta in metas:
```
mbargull: Regarding code complexity (or just cosmetics?):
Instead of adding loops explicitly, we could just add a decorator like
```python
def lint_multiple_metas(lint_function):  # better name?
    def lint_metas(recipe, metas, df, *args, **kwargs):
        lint = partial(lint_function, recipe)
        for meta in metas:
            ret = lint(meta, df, *args, **kwargs)
            if ret is not None:
                return ret
    return lint_metas

@lint_multiple_metas
def in_other_channels(recipe, metas, df):
[...]
```

bioconda_utils/docker_utils.py
johanneskoester: I don't think so. The nice thing of the current approach is that we can synchronize host and container conda/conda-build without additional facilities. With such a recipe, it will only get out of sync.

bioconda_utils/lint_functions.py
johanneskoester: Yes, indeed, that's a good idea. But it should not stop us from merging. If nobody has the time for doing it, we can postpone it.

bioconda_utils/docker_utils.py
mbargull: We could have those perfectly in sync if we install `bioconda-build-setup=version` to either.

(With `conda >=4.4` we could even `conda create bioconda-build-setup=version` and use a completely separate environment for `conda-build` on the host which does not depend on the packages introduced by `bioconda-utils`. I'm not sure about if that approach is that valuable, but at least we could use a smaller Docker image thatn `bioconda-utils-build-env` -- maybe (I have no clue) this could reduce the build time.)

bioconda_utils/lint_functions.py
mbargull: done

.circleci/config.yml
```diff
@@ -97,6 +108,8 @@ workflows:
     jobs:
       - test-linux:
           context: org-global
+      - test-linux (long_running):
```
mbargull: I separated the two longest running tests so that CircleCI can run them parallel to the others, roughly cutting the clock time by 50 percent. I also added this `test-linux (long_running)` to the required checks in GitHub afterwards. Hope that's fine by you.

mbargull:
Regarding the `host:` lint, let's postpone. Looks like there is a small bug currently (https://github.com/conda/conda-build/issues/2887) that partially negates my previous statement
> If we make `host:` mandatory (even if empty), we could ensure more uniform build setups, i.e., always have two environments (`build` and `host`) activated.

.circleci/config.yml
daler: @mbargull good idea. I ended up having to merge this into https://github.com/bioconda/bioconda-utils/pull/300. Github was expecting the long-running job but circleci didn't run it since it wasn't configured in .circleci/config.yaml yet. Result was a hung test.

bioconda_utils/docker_utils.py
johanneskoester: Ok, we can keep this in mind, but I still don't see the immediate benefit over what we have here. Seems much simpler to me than another package.

johanneskoester:
Ok, I think this is ready for merging isn't it? Maybe it would be good to coordinate that. Is there a day next week where all of us have time to work on fixes if problems occur?

I suggest the following plan:

1. Release a final CB2 version of bioconda utils before merging.
2. Pin to that version in bioconda/bioconda-common.
4. Temporarily modify the bulk branch to use this branch of bioconda-utils.
5. Update packages in there, and fix any errors or problems that occur.
6. Merge this PR.
7. Release a new version of bioconda-utils.
8. Pin to that version in bioconda/bioconda-common.

daler:
@johanneskoester that sounds good. This would be a good week for me, any day, to stand by for fixes.

bgruening:
I'm super excited, thanks a lot for working on this and sorry to be so silent the last weeks.
Friday I'm back and we could do this ... but anytime earlier would also be fine.

:+1: for the PR.

johanneskoester:
Ok, @daler, @bgruening, Friday afternoon CET?

johanneskoester:
@bgruening bulk seems to be diverged from master. Since you have done the last commits, can you check and see whether everything has been merged into master and is in sync, so that we have a clean base to start with? I don't want to wipe your work.

mbargull:
Oh, I think we need to prepare the Docker tags before using this for `bulk` PR. The default (which get also used for `bioconda-recipes`) for `bioconda_utils.docker_utils.RecipeBuilder.dockerfile_template` uses `bioconda-utils-build-env:latest`.
Should we change at https://github.com/bioconda/bioconda-utils/blob/46866050d909559d27436fda7daa030d5a5e18cf/bioconda_utils/docker_utils.py#L121
```diff
 DOCKERFILE_TEMPLATE = \
 """
-FROM bioconda/bioconda-utils-build-env
+FROM {self.docker_base_image}
 {self.proxies}
 RUN /opt/conda/bin/conda install -y conda={conda_ver} conda-build={conda_build_ver}
"""
```
and set
```python
self.docker_base_image = "bioconda/bioconda-utils-build-env:{}".format(
    os.environ.get("BIOCONDA_UTILS_TAG", "latest"))
```
in `bioconda_utils.docker_utils.RecipeBuilder.__init__`?
For that to we'd need Docker Hub to create tags not only for branches but also Git tags, e.g.,
`bioconda-utils-build-env:v0.12.4`.

mbargull:
What do you think? Should I push https://github.com/mbargull/bioconda-utils/commit/3fc9fd994db481e3c21ce4eaf6e86b7c147b948a to this PR?

mbargull:
Working on `bulk` whilst setting `BIOCONDA_UTILS_TAG=cb3-migration` there should work with that commit, I think.
The changes to Docker Hub would have to be made at latest before merging this PR.

bgruening:
Ok, online. Anything what I can help? Working on bulk?

johanneskoester:
@mbargull, that is a good plan. Can you push your commit?
@bgruening, could you clean up bulk and sync it with master?

bgruening:
@johanneskoester https://github.com/bioconda/bioconda-recipes/pull/9040 something seems to be wrong with the CI configuration. We can fix this later. Do you want me to merge it with admin power?

johanneskoester:
Done.

johanneskoester:
Ok, I will now activate cb3-migration on the bulk branch and create a new PR for tracking.

daler:
@johanneskoester @bgruening `bulk` is not building; I thought https://github.com/bioconda/bioconda-utils/pull/293/commits/2fb47ea9edc109ed5ff488b02752a79335006220 would fix it (which it did locally). I must be missing something because it's still not working on circleci. I will be able to return to this tonight or tomorrow.

mbargull:
@daler, that's because CircleCI's cache didn't "see" those changes. I just disabled it temporarily while we get some changes in here.

mbargull:
Oh, I just noted that `conda-forge-pinning`'s newer versions include an, I think, important change:
https://github.com/conda-forge/conda-forge-pinning-feedstock/compare/793426c3b16a049fc0bdaba1a731103b9548c6f7...798fe4884150408a3f03796c815ed66c3b4cb856
namely https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/58, i.e., restricted pinning of `boost`. I think we should update to `conda-forge-pinning 2018.05.22`, do you agree?

mbargull:
And it looks like there could possibly be a new `boost`-related release for `conda-forge-pinning` soon: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/71 (though, that will just be "normal business" on updating their pinnings).

bioconda_utils/utils.py
```diff
@@ -482,9 +482,7 @@ def get_inner_deps(dependencies):
     for meta, recipe in metadata:
         name = meta["package"]["name"]
         dag.add_edges_from((dep, name)
-                           for dep in set(get_inner_deps(chain(
-                               get_deps(meta, "build"),
-                               get_deps(meta, "host")))))
+                           for dep in set(get_inner_deps(get_deps(meta, "host"))))
```
mbargull: I get the failing case (dependency circle on those Perl recipes), but I don't think this change makes sense. Can't we just blacklist those recipes?

mbargull:
> I think we should update to `conda-forge-pinning 2018.05.22`, do you agree?

Since I think no packages from https://github.com/bioconda/bioconda-recipes/pull/9041 got actually built, yet, I just updated the pinning.

docs/source/cb3.rst
mbargull: xrefs from a linked Gitter conversation:
https://gitter.im/conda-forge/conda-forge.github.io?at=5b1006bfb6eece791dca9b6e
https://gitter.im/conda-forge/conda-forge.github.io?at=5b1014c8e26c847ac8a799ec
https://gitter.im/conda-forge/conda-forge.github.io?at=5b10221ba7abc8692efc7b69

bioconda_utils/utils.py
```diff
@@ -721,115 +771,88 @@ def changed_since_master(recipe_folder):
...
+        existing_build_ids = channel_packages.get(pkg_key, set())
+        for meta in pkg_metas:
+            if meta.build_id() not in existing_build_ids:
+                new_metas.append(meta)
+            else:
+                existing_metas.append(meta)
+        pkg_build_ids = set(meta.build_id() for meta in pkg_metas)
+        for divergent_build in (existing_build_ids - pkg_build_ids):
+            divergent_builds.add(
+                '-'.join((pkg_key.name, pkg_key.version, divergent_build)))
```
mbargull: TODO/FIXME: This misses a corner case where packages have mismatches *only* in the `osx-64` subdir since `channel_packages` only contains data from `linux-64` and `noarch` when we have a `noarch` build.
Solution could be to always fail when we do a `noarch` build and there are builds with the same number in `linux-64` or `osx-64`.
Otherwise we'd have to also render with `CONDA_SUBDIR=osx-64` (at least when we do a `noarch` build) to be able to inspect those build strings. But we don't have to go through this trouble since we do not support the case where we have builds with the same number in `noarch` and simultaneously other subdirs <sup>(That's actually valid, albeit the use cases are rather exotic, currently.)</sup>.

test/test_utils.py
```diff
@@ -187,18 +193,21 @@ def test_upload(single_upload):
...
+    for pkg in single_build:
+        pkg_test.test_package(pkg)
 
 
 @pytest.mark.long_running
 def test_multi_build(multi_build):
     for v in multi_build.values():
-        assert os.path.exists(v)
+        for pkg in v:
+            assert os.path.exists(pkg)
```
mbargull: I assume this test currently fails because of https://github.com/bioconda/bioconda-utils/pull/293/commits/a6dabe3cdc470e0c2c1c6ddd8a9e52a941e95589.

bioconda_utils/linting.py
```diff
@@ -213,15 +182,18 @@ def lint(recipes, config, df, exclude=None, registry=None):
         skip_dict[recipe].append(func)
 
     hits = []
-    for recipe in recipes:
+    for recipe in sorted(recipes):
```
mbargull: :+1:

bioconda_utils/lint_functions.py
```diff
@@ -225,8 +225,10 @@ def has_windows_bat_file(recipe, metas, df):
 @lint_multiple_metas
 def should_be_noarch(recipe, meta, df):
     deps = _get_deps(meta)
+    no_compilers = sum(1 for i in deps if
+                        i.startswith('toolchain_')) == 0
```
mbargull: We'll have to make this one a bit more generic in the future when the Anaconda built compilers are used.
A similarly hardcoded but more general solution could be
```python
    no_compilers = all(
        not dep.startswith(('clang_', 'clangxx_', 'gcc_', 'gxx_', 'gfortran_', 'toolchain_'))
        for dep in deps)
```

bioconda_utils/linting.py
```diff
@@ -232,14 +204,14 @@ def lint(recipes, config, df, exclude=None, registry=None):
                  'severity': 'ERROR',
                  'info': result})
             continue
-        logger.debug('lint {}'.format(recipe))
```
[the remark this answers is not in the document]
daler: Hah, I had done the same thing, I thought I had added my local edits and was cleaning them up again. Added back in.

bioconda_utils/lint_functions.py
daler: done, thanks.

mbargull:
> Maybe we should just update to the latest sphinx version?

I think it's tripping on some recipe that has some malformation or at least unexpected content.
Last time I checked, we were already using the latest `sphinx`. (I also checked that the failing build was not related to the Sphinx update, too. Hence I assume it's something in the recipes.)
I'll change `build-docs.sh` on another branch to use only one job -- maybe that'll show some more information. Following that, we can merge into master, sure.

mbargull:
There is at least one recipe (`atactk`) that uses a list for `extra/notes`. Although this should be a string, I added a change to support this anyway since I've seen `extra/notes` being a list in multiple PRs in the past (which I guess makes sense since "notes" is a plural, hence contributors make it a list)

mbargull:
Would be nice to get the error output from https://circleci.com/gh/bioconda/bioconda-utils/1037 when using multiple jobs, though...

bioconda_utils/build.py
```diff
@@ -384,6 +385,15 @@ def build_recipes(
             for n in nx.algorithms.descendants(subdag, name):
                 skip_dependent[n].append(recipe)
             continue
+        except UnsatisfiableError as e:
```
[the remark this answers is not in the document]
mbargull: Yep, I'm also a bit puzzled we didn't hit this before...
Looks like it was only fixed on `conda-build:master` lately in: https://github.com/conda/conda-build/pull/2974

.circleci/build-docs.sh
```diff
@@ -58,7 +58,9 @@ rm -r *
 
 # build docs and copy over to tmpdir
 cd ${DOCSOURCE}
-make clean html SPHINXOPTS="-j2" 2>&1 | grep -v "WARNING: nonlocal image URL found:"
+# TODO: reenable "-j2" when docs build fine
```
mbargull: Looking at the build times, it doesn't make much sense to use multiple jobs here. It's less than a minute faster with `-j2` but also "obfuscates" the progress/error output..

bgruening:
Bulk is green and we have rebuild our pkgs, a substantial amount is still blacklisted and we will work on this in the coming weeks.
But I think it is important to merge this to get the changes onto our homepage.

johanneskoester:
@bioconda/core I will shortly update bioconda-recipes:master to point to the new release 0.13 of bioconda-utils. I am just waiting for the docker build to finish.

johanneskoester:
@bioconda/core done. Migration is complete (at least from the bioconda-utils side)!

PR #296, opened 2018-05-04
Add Haskell guidelines

luispedro:
@bioconda/haskell : I decided to share some of the little tricks I learned while working on the ngless recipe yesterday, but maybe this should be a group effort and if you have other tips, add them here.


bgruening:
ping @eggzilla

eggzilla:
Nice! Thanks for the write-up :-)

PR #311, opened 2018-06-19
exclude less important lints on build

mbargull:
- renamed `--prelint` to `--lint` since this happens always happens pre-build.
- added `--lint-only`/`--lint-exclude` to `build` command (equivalent to `--only`/`--exclude` from `lint` command).
- ~if none of `--lint-only`, `--lint-exclude` is given, all lints not in `lint_functions.registry_build` are excluded.~ (EDIT: didn't make sense since the build lint is also run on PRs)

### DIFF
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -22,3 +22,4 @@ six=1.11.*
 alabaster=0.7.*
 git=2.14.*
 conda-forge-pinning=2018.05.22
+python>=3.6
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -312,9 +312,7 @@ def lint(recipe_folder, config, packages="*", cache=None, list_funcs=False,
 @arg('--lint-only', nargs='+',
      help='''Only run this linting function. Can be used multiple times.''')
 @arg('--lint-exclude', nargs='+',
-     help='''Exclude this linting function. Can be used multiple times.
-     If neither --lint-exclude nor --lint-only is given, this defaults to
-     exclude many less importand lints. (cf. `lint_functions.registry_build`)''')
+     help='''Exclude this linting function. Can be used multiple times.''')
 def build(
     recipe_folder,
     config,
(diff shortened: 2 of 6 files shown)


bioconda_utils/cli.py
```diff
@@ -309,10 +305,16 @@ def lint(recipe_folder, config, packages="*", cache=None, list_funcs=False,
 @arg('--keep-image', action='store_true', help='''After building recipes, the
      created Docker image is removed by default to save disk space. Use this
      argument to disable this behavior.''')
-@arg('--prelint', action='store_true', help='''Just before each recipe, apply
+@arg('--lint', '--prelint', action='store_true', help='''Just before each recipe, apply
```
mbargull: Left `--prelint` as an alias here so that this can be swapped out without issues. The alias should be removed when `bioconda-recipes` doesn't use it anymore.

bioconda_utils/bioconda_utils-requirements.txt
```diff
@@ -22,3 +22,4 @@ six=1.11.*
 alabaster=0.7.*
 git=2.14.*
 conda-forge-pinning=2018.05.22
+python>=3.6
```
mbargull: For some reason one of the CI builds picked up `python 3.5.*` instead of `3.6`.
That one choked on https://github.com/bioconda/bioconda-utils/blob/41f39c06607fe142d47e5ba244cd88444e5367cd/bioconda_utils/utils.py#L217 complaining that `os.path.join` doesn't like the `*Path` objects.

mbargull:
I think we can merge this no matter if we'll be using it or not since all of those options aren't activate on default.

bgruening:
@mbargull this error seems to use py36:
```
Traceback (most recent call last):
  File "/home/circleci/project/test/test_utils.py", line 210, in test_multi_build
    assert os.path.exists(pkg)
AssertionError: assert False
 +  where False = <function exists at 0x7fb448fddae8>('/home/circleci/project/miniconda/conda-bld/linux-64/two-0.1-0.tar.bz2')
 +    where <function exists at 0x7fb448fddae8> = <module 'posixpath' from '/home/circleci/project/miniconda/lib/python3.6/posixpath.py'>.exists
 +      where <module 'posixpath' from '/home/circleci/project/miniconda/lib/python3.6/posixpath.py'> = os.path
```

mbargull:
Already merged as part of gh-312.

PR #313, opened 2018-07-04
Handle dependency cycles

mbargull:
The first three commits are some minor fixes.
The last two fix handling of dependencies by considering all requirements from `build`, `host`, as well as `run` environments. To avoid the previous workaround of only considering `host` due to cyclic dependencies of some existing recipes, this detects those dependency cycles and marks all packages of a cycle as `failed` (+ skipping any downstream packages, of course).

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -1,5 +1,4 @@
 import subprocess as sp
-from itertools import chain
 from collections import defaultdict, namedtuple
 import os
 import logging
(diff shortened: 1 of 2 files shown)


bioconda_utils/utils.py
```diff
@@ -846,6 +845,16 @@ def check_recipe_skippable(recipe, channel_packages, force=False):
         ),
         Counter()
     )
+    if num_existing_pkg_builds == Counter():
+        # No packages with same version + build num in channels: no need to skip
+        return False
```
mbargull: If
```python
num_existing_pkg_builds == num_new_pkg_builds == Counter()
```
this previously caused an additional
> ... the same number of builds are in channel(s) ...

info message to be displayed. But since
```python
num_new_pkg_builds == Counter()
```
means there is nothing to do anyway,
> Nothing to be done for recipe ...

is output in any case so that the former message is just noise.
ref: https://github.com/bioconda/bioconda-recipes/pull/9686#issuecomment-401799336

bioconda_utils/utils.py
```diff
@@ -492,7 +492,7 @@ def get_deps(meta, sec):
         deps = reqs.get(sec)
         if not deps:
             return []
-        return [dep.split()[0] for dep in deps if dep is not None]
+        return [dep.split()[0] for dep in deps if dep]
```
mbargull: `load_meta_fast` does not evaluate the Jinja functions. Hence, `{{ compiler(...) }}` or `{{ pin_compatible(...) }}` may be empty strings here.
Changing
```python
if dep is not None
```
to
```
if dep
```
just prevents raising `IndexError` in for `dep.split()[0]`.
If we want to handle this correctly, we'd have to evaluate the Jinja constructs.

bioconda_utils/build.py
```diff
@@ -322,14 +320,45 @@ def build_recipes(
...
     else:
-        # take connected components as subdags
-        subdags = sorted(map(sorted, nx.connected_components(dag.to_undirected(
-        ))))
+        # take connected components as subdags, remove cycles
+        subdags = []
+        for cc_nodes in nx.connected_components(dag.to_undirected()):
+            cc = dag.subgraph(sorted(cc_nodes))
+            nodes_in_cycles = set()
+            for cycle in list(nx.simple_cycles(cc)):
```
mbargull: I did some quick tests with the existing packages by clearing the build blacklist to get the cyclic dependency of the Perl packages that caused the troubles before.
Didn't write tests yet, though, sorry :man_facepalming: :face_with_head_bandage:

mbargull:
NB: we should rename all `dag`s to `dependency_graph` up until the point when we removed the cycles. Doesn't make much sense to call them DAGs :laughing:

bioconda_utils/utils.py
johanneskoester: This is fine with me.

johanneskoester:
@mbargull, I guess you will be busy the next days with travel preparation and the job application. I think it should be fine to merge this if you agree, although there is no test case yet.

bgruening:
Sorry but I think this broke the builds, e.g. https://circleci.com/gh/bioconda/bioconda-recipes/16499

```
  File "/home/circleci/project/miniconda/lib/python3.6/site-packages/conda/core/package_cache_data.py", line 605, in execute
    raise CondaMultiError(exceptions)
conda.CondaMultiError: Exceeded 30 redirects.
```

mbargull:
Yep, looks like that has been some temporary fluke: The current `bulk` run is past the point were the previous one failed.

PR #322, opened 2018-08-01
add libdeflate pinning

johanneskoester:


kyleabeauchamp:
Thanks, this never even crossed my mind!

PR #324, opened 2018-08-08
bump conda-build to 3.11.0

rvalieris:
this version of conda-build includes this PR: https://github.com/conda/conda-build/pull/3021
it affects recipes that use multiple urls.


rvalieris:
@bioconda/core is this ok ? what else do I need to do to makes these updates available for bioconda-recipes builds ?

bgruening:
That looks good to me.
ping @daler

mbargull:
Updated to `conda 4.5.9`:
- The changes are mostly bug fixes and there is no reason not to update.
Updated to `conda-build 3.12.1`:
- Given the changes, it should be fine to update this too.
- Of interest to some people might be the addition of
  > add the "extra" field of a package's meta.yaml file into the output package's info/about.json file https://github.com/conda/conda-build/pull/3048

daler:
Sounds good to me. And after the change, may need to update https://github.com/bioconda/bioconda-common/blob/master/common.sh as well.

bgruening:
There is also worthwhile to mention and should help some people: https://github.com/conda/conda/pull/7601
Please note it is discussed to enable this for the next release.

bgruening:
General question for the pinning update. Do we need to bump the build versions if we rebuild packages just for the pinning?

mbargull:
> There is also worthwhile to mention and should help some people: https://github.com/conda/conda/pull/7601

IMHO, that part is of most interest for `conda-forge` Windows users -- so Bioconda users don't profit *that* much as is :wink:. The problematic with `numpy` and `openblas` from `conda-forge` is still an issue. I'll write up on this on the Conda issue.

daler:
@bgruening I was about to ask exactly the same thing. Just packages depending on freetype and libuuid would need to be rebuilt, right?

mbargull:
> Do we need to bump the build versions if we rebuild packages just for the pinning?

You mean for packages that depend on those pinned ones that are affected by an update of `conda-forge-pinning` (and/or our pinning file)? Yes, we should -- and also automate this at some point when someone finds the time to do so..

bgruening:
> @bgruening I was about to ask exactly the same thing. Just packages depending on freetype and libuuid would need to be rebuilt, right?

Yes, all pkgs that depend on freetype and co ... the nightly builds and/or bulk will take care of this, but if we need to bump the build number this currently needs manual intervention. We need to automate this.

mbargull:
For future updates of `freetype` we don't necessarily have to rebuild much since the pinning was relaxed :+1:

mbargull:
Okay, updated the `conda-forge-pinning` requirement.
For reference, these are the recipes with `freetype` as a build/host requirement:
```
connectome-workbench/meta.yaml
graphicsmagick/1.3.23/meta.yaml
graphicsmagick/1.3.26/meta.yaml
python-bioext/meta.yaml
rdkit/2015.09.2/meta.yaml
rdkit/2016.03.3/meta.yaml
umi_tools/meta.yaml
```

`libuuid` is used for the `ucsc-*` packages but we don't need to rebuild those given the newly introduced loose pin.

bgruening:
Thanks @mbargull!

bgruening:
I updated the build numbers as well.

mbargull:
Thanks, but we'll need another release cycle for this since we don't use the `cb3-migration` branch directly anymore but use the release tags again.
I.e., we need to tag a new release, then update `bioconda-common`, and then bump the `freetype`-depending packages' build numbers.

rvalieris:
so just to make sure I got it right, to update bioconda-utils:

1. make the commit on bioconda-utils
2. make a new release of bioconda-utils
3. update bioconda-common with the new release
4(?). docker image updates automatically ?

bgruening:
@rvalieris thanks for doing the release!

Issue #330, opened 2018-09-04
Channel priority: [bioconda, conda-forge] or [conda-forge, bioconda]


mbargull:
@bioconda/core: What do you think about changing the channel priority (as suggested in the docs and used for the CI) from
```
- bioconda
- conda-forge
- defaults
```
to
```
- conda-forge
- bioconda
- defaults
```
? We actively try to avoid (via linter checks in Bioconda as well as conda-forge) name clashes between different packages in both channels. Hence the cases where we have packages with the same name in both channels are usually when we have outdated version of packages in `bioconda` for which newer versions are now maintained in `conda-forge`. Due to `conda`'s channel priority we then express a preference for the older version in `bioconda` if that channel has higher priority. This is usually not what the user would expect and also something that puts excessive load on the dependency solver, e.g.,
trying to run
```
conda create -nenv -c bioconda -c conda-forge qualimap=2.2.2a=3
```
takes 14 hours or so, while
```
conda create -nenv -c conda-forge -c bioconda qualimap=2.2.2a=3
```
finishes in seconds. Plus, `-c bioconda -c conda-forge qualimap=2.2.2a=3` will give you the old `r=3.3.2` (due to outdated R packages in `bioconda`) which is also probably not what a user would expect.

Personally, I'm very much in favor of changing the channel priority and don't see much benefit in having `bioconda` on top. What do think?

----
latest refs:

----
@abalter in https://github.com/conda/conda/issues/7646#issuecomment-418315927: 
> @mbargull -- you made [this suggestion](https://github.com/conda/conda/issues/7684#issuecomment-415588307) on a different issue I raised, and it solved the problem. It also solved other problems I was having with building environments with `bioconda` bioconductor packages. I should probably raise the issue with bioconda.
> 
> > The current state of the Bioconda docs still says to give bioconda the higher priority. In reality (due to the issue you just encountered) in often makes sense to use
> ```
> channels:
>   - conda-forge
>   - bioconda
>   - defaults
> ```
> >though. (defaults has to be in the mix because conda-forge is not yet self-contained and depends on packages only available in defaults.)

----
@mbargull in https://github.com/conda/conda/issues/7684#issuecomment-415588307:
> > @mbargull does work on both projects so I'm pinging him in-case he has some advice.
> 
[...]

PR #333, opened 2018-09-12
Remove deprecated log message in case of empty linting job.

johanneskoester:
This would previously have thrown an unitialized variable error. In addition this PR fixes a syntax error in index.rst, leading to a broken paper link.

PR #334, opened 2018-09-12
Use conda-forge with a higher priority than bioconda.

bgruening:
xref: https://github.com/bioconda/bioconda-utils/issues/330

### DIFF
--- a/bioconda_utils/pkg_test.py
+++ b/bioconda_utils/pkg_test.py
@@ -130,7 +130,7 @@ def test_package(
     if "local" not in channels:
         raise ValueError('"local" must be in channel list')
 
-    channel, extra_channels = [
+    channel, *extra_channels = [
         'file://{0}'.format(conda_bld_dir) if channel == 'local' else 'channel'
         for channel in channels
     ]
--- a/bioconda_utils/pkg_test.py
+++ b/bioconda_utils/pkg_test.py
@@ -83,7 +83,7 @@ def get_image_name(path):
 def test_package(
     path,
     name_override=None,
-    channels=["conda-forge", "defaults"],
+    channels=("conda-forge", "bioconda", "local", "defaults"),
     mulled_args="",
     base_image=None,
     conda_image=MULLED_CONDA_IMAGE,
@@ -99,9 +99,9 @@ def test_package(
     name_override : str
         Passed as the --name-override argument to mulled-build
 
-    channels : None | str | list
-        The local channel of the provided package will be added automatically;
-        `channels` are channels to use in addition to the local channel.
+    channels : list
+        List of Conda channels to use. Must include an entry "local" for the
+        local build channel.
 
     mulled_args : str
         Mechanism for passing arguments to the mulled-build command. They will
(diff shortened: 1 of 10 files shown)


daler:
Found a couple more. I don't know what `example_config.yaml` is doing (I can't find where it's used) but figured I'd fix the priority in it and keep it rather than remove it right now.

johanneskoester:
What happened to this long-running test case? I see the same error in #333...

mbargull:
> What happened to this long-running test case? I see the same error in #333...

fixed in gh-335.

bioconda_utils/pkg_test.py
```diff
@@ -127,13 +127,17 @@ def test_package(
...
-    extra_channels.extend(channels)
-    channel_args = ['--extra-channels', ','.join(extra_channels)]
+    if "local" not in channels:
+        raise ValueError('"local" must be in channel list')
+
+    channel, extra_channels = [
+        'file://{0}'.format(conda_bld_dir) if channel == 'local' else 'channel'
+        for channel in channels
+    ]
+    # '--channel' currently defaults to 'bioconda' in mulled-build and is
```
mbargull: xref: https://github.com/galaxyproject/galaxy-lib/pull/132

bioconda_utils/pkg_test.py
```diff
@@ -99,9 +99,9 @@ def test_package(
     name_override : str
         Passed as the --name-override argument to mulled-build
 
-    channels : None | str | list
-        The local channel of the provided package will be added automatically;
-        `channels` are channels to use in addition to the local channel.
+    channels : list
```
mbargull: I don't think the `None | str` option is used anywhere, so I think it's fine to remove them.

bioconda_utils/pkg_test.py
```diff
@@ -83,7 +83,7 @@ def get_image_name(path):
 def test_package(
     path,
     name_override=None,
-    channels=["conda-forge", "defaults"],
+    channels=("conda-forge", "bioconda", "local", "defaults"),
```
mbargull: The order `"bioconda", "local"` is what we currently have (since `bioconda` gets prepended to everything). Granted, that order is debatable. What we *actually* would want is to have `bioconda` and `local` at the same priority. `conda` currently does not provide such functionality. If we wanted to have that, we would either have to
1. put such functionality into `conda`, or
2. spawn a local HTTP server that transparently merges `bioconda` and `local`.

The second option might be easier since for the first option one would have to think about and coordinate the user interface changes with the Conda devs.

johanneskoester: I would think it should be local first, shouldn't it? At least before bioconda. Because those are the packages that where created during that run, and we surely want to prefer those over old stuff in bioconda.

mbargull: Merging `bioconda` and `local` would replicate what you get after the packages have been uploaded, so I think this would be the "ideal" solution, maybe.
`(bioconda, local)` is just the status quo, but I completely agree that `(local, bioconda)` is the better choice.

johanneskoester:
For me, this is ready to merge. Any objections?

bgruening:
Yes, thanks!

johanneskoester:
Good job guys! Thanks for taking care of this!

PR #339, opened 2018-09-24
Add lints that prevent jinja variable misuse. Add subcommand for bumping build number.

johanneskoester:
# Part one
Current conda-build supports accessing package version in meta.yaml via {{ PKG_VERSION }}.
Therefore, it is no longer reasonable to define your own jinja variables. For other parts
like name, build number and checksum, it was never reasonable because these either don't change
during updates (name) nor do they occur more than once.
By enforcing non-jinja style here, the recipe becomes much more readable.
Moreover, it becomes easier to update programatically.

# Part two
This PR adds a subcommand `bioconda-utils bump-build-number`, that takes a list of recipes and increases their build number by one. This helps us to rebuild stuff when the version pinnings change. Ideally, I would like to automate this, e.g. by finding out which pinning changed, and scanning through all packages with a corresponding dependency. However, this is not done yet.

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -632,6 +632,16 @@ def pypi_check(recipe_folder, config, loglevel='info', packages='*', only_out_of
         print('\t'.join(map(str, result)))
 
 
+@arg('recipe', nargs="+", help='Path to recipes')
+@arg('--loglevel', default='debug', help='Log level')
+def bump_build_number(recipes, loglevel="info"):
+    """
+    Bumps the build number in given recipes.
+    """
+    utils.setup_logger('bioconda_utils', loglevel)
+    utils.bump_build_number(recipes)
+
+
 def main():
     argh.dispatch_commands([
         build, dag, dependent, lint, duplicates,
--- a/bioconda_utils/lint_functions.py
+++ b/bioconda_utils/lint_functions.py
@@ -3,6 +3,7 @@ import glob
 import os
 import re
 
+import yaml
 import pandas
 import numpy as np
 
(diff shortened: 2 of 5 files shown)


johanneskoester:
@bioconda/core, please have a look at this, opinions welcome. In particular regarding part one.

johanneskoester:
An example recipe that implements the changed strategy from part one can be found here: https://github.com/bioconda/bioconda-recipes/pull/11042

daler:
Thanks @johanneskoester, this is a good idea. Will take some effort to retrofit existing recipes, but using jinja vars at the top for version/name/hash never sat well with me, and you describe exactly why in the lint docs.

johanneskoester:
Indeed, I am very happy as well that we finally can get rid of them. I suggest merging ASAP, and then fix old recipes when we anyway have to rebuild them (e.g., because of the compiler update on conda-forge).

bgruening:
I have the feeling that this diverges us more from the rest of the conda universe. I __do__ think its a good change but we should talk to conda-forge and push it their in parallel and even more important I think we should push for skeleton changes. Currently all the skeleton generators do generate the opposite, isn't it?

It would be a bad UX if people using skeletons and need to change a lot of stuff to make our linter happy.

Thanks @johanneskoester for working on this, I do like it - I just think we should discuss this at a different level and not go for a bioconda only solution.

johanneskoester:
Yes I agree @bgruening. Might be that conda-forge does not want to change this quickly, because I think they have a lot of tooling that relies on those jinja variables. Nevertheless it is worth a try. Apart from that, I think we can solve the issue with the skeletons by providing a PR that adds an option to generate a jinja-free recipe from every skeleton generator. This way, we only have to tell people, e.g., use `conda skeleton pypi --profile plain-yaml somepackage`.

johanneskoester:
I would take care of that PR if you agree that this would be a feasible approach to proceed fast.

bgruening:
> I would take care of that PR if you agree that this would be a feasible approach to proceed fast.

That would be fantastic!!! Thanks Johannes!

johanneskoester:
Do you want to approach conda-forge about this, or shall we ask our connection office @mbargull? :-)

mingwandroid:
I strongly prefer the jinja2 style with these values at the top.

Primrarily because it makes diff'ing and merging so much easier if these variables are in one block at the top and the rest of the recipe doesn't change.

Issue #316, opened 2018-07-23
linting fn is too strict


bgruening:
> 1	recipes/r-bipartite	should_not_use_fn	{'should_not_use_fn': True, 'fix': 'URL should specify path to file, which will be used as the filename'}		URL should specify path to file, which will be used as the filename	True

What was the intend of this change? I think we need fn in R packages, or to be more precise in all packages that do have multiple URLs.

Can we disable that or check for a list or URLs?

bgruening:
ping @mbargull @daler

mbargull:
Sorry I didn't reply before, I'm still lacking behind on a **ton** of issues...
I [fixed](https://github.com/conda/conda-build/pull/3021) that in `conda-build` a few days before you posted this issue :smile:.
The updates from gh-324 should take care of this!

epruesse:
Closing - we have only very few recipes with `fn:` and most seem to be indeed unnecessary.

PR #343, opened 2018-10-02
Have linting obey blacklist, modify bioconductor skeleton, fix cleaning of CRAN recipes

### LINKED ISSUE #342: bioconductor-skeleton produced meta.yaml fail linting
Both `conda skeleton cran` and `bioconda-utils bioconductor-skeleton` produced `meta.yaml` files currently set download filenames, which causes linting to fail. Fixing the bioconductor skeleton should be easy enough. Perhaps `bioconda-utils clean-cran-skeleton` could fix the `conda skeleton cran` produced files.

dpryan79:
Fixes #342 among others.

At the moment, `bioconda-utils lint` effectively ignores the blacklist because it's not stripping `recipes/` from recipes paths before checking to see if they're blacklisted (c.f., [this circleci log](https://circleci.com/gh/bioconda/bioconda-recipes/28181)).

This also changes the bioconductor skeleton to add:

    extra:
      skip-lints:
        - should_not_use_fn

The equivalent is now added to CRAN recipes by `bioconda-utils clean-cran-skeleton`. This is related to #316, which apparently we're still seeing even though we're using a `conda-build` version that should be patched (presumably this is a regression).


bgruening:
@dpryan79 have you tried upgrading to conda-build 3.15.1 (https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/bioconda_utils-requirements.txt#L5). Does this help?

dpryan79:
I've tried 3.15.1, though I can give it another go to be sure.

dpryan79:
Weird, I must have been changing the conda-build version in a different conda environment last night, because now 3.15.1 works for me without the filename stuff. I'll redo a bunch of tests and update recipes and update this PR after further testing (likely Thursday).

bgruening:
Perfect! Thanks @dpryan79 for double checking!

dpryan79:
I've now reverted all of the file name stuff and it seems to work locally.

I suspect a change in the conda API is causing one test to fail. If I start with a clean conda environment and run only `test/test_utils.py::test_skip_dependencies` it succeeds the first time I run it and then fails thereafter. I'll see if I can track down the cause.

dpryan79:
The following commit removed the ability of `conda index` to remove deleted files from the cache, which happened between 3.14.4 and 3.15.0. This is causing the current build failure.

dpryan79:
I've changed one of the tests to simply avoid the new `conda index` behavior.

dpryan79:
Nope, it's not WIP any more, I forgot to change that.

dpryan79:
Relatedly, I'm going to make a small example of the `conda index` issue I mentioned above and open an issue on conda-build.

dpryan79:
Ah, the issue is known https://github.com/conda/conda-build/issues/3167

bgruening:
Thanks @dpryan79! I will pin a release.

dpryan79:
@bgruening Thanks, you'll at the very least need to update bioconda-common to point to the commit hash. I don't actually know if that works, so you might just tag this and then put that tag in bioconda-common.

bgruening:
I know, this was more a question how urgent you want to have it.

bgruening:
Done, both!

PR #344, opened 2018-10-09
Expose check_channels on the CLI and check the first 2 by default.

dpryan79:
This should fix builds in `bulk` and any other environment where `--git-range` is not specified. Previously, the first channel in the channels list in the config file was checked to ensure that packages that already existed weren't built. Since the channel order was changed, this now means that `conda-forge` is checked against rather than `bioconda`. It's probably good to check against `conda-forge`, but especially for the `bulk` branch we need to check the `bioconda` channel as well. This PR causes up to the first 2 channels in the config file to be checked against before building/uploading. This will fix builds in the `bulk` branch. Additionally, this adds a `--check-channels` option to `bioconda-utils build`, so this can be directly changed by users (or the script given to circleci). The `bioconda-utils build` option isn't really needed, it's just for convenience in the future (so I'm happy to remove it if desired).

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -318,6 +318,11 @@ def lint(recipe_folder, config, packages="*", cache=None, list_funcs=False,
      help='''Only run this linting function. Can be used multiple times.''')
 @arg('--lint-exclude', nargs='+',
      help='''Exclude this linting function. Can be used multiple times.''')
+@arg('--check-channels', nargs='+',
+     help='''Channels to check recipes against before building. Any recipe
+     already present in one of these channels will be skipped. The default is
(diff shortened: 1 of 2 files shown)


bioconda_utils/cli.py
```diff
@@ -318,6 +318,11 @@ def lint(recipe_folder, config, packages="*", cache=None, list_funcs=False,
      help='''Only run this linting function. Can be used multiple times.''')
 @arg('--lint-exclude', nargs='+',
      help='''Exclude this linting function. Can be used multiple times.''')
+@arg('--check-channels', nargs='+',
+     help='''Channels to check recipes against before building. Any recipe
```
bgruening: Not really recipe but pkg/version combination afaik.

PR #355, opened 2018-11-16
Allow conversion to e.g. win-64 for pure python packages via 'conda convert'.

johanneskoester:
This PR adds automatic conversion to e.g. windows to bioconda-utils. This will only work for pure python packages. Target platforms have to be specified like this:

```
extra:
  convert-to:
      - win-64
```

If that section is empty or not present, conversion will be skipped. This allows us to enable windows builds in a very controlled way without runtime overhead (conversion is a matter of seconds).

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -85,7 +85,7 @@ def build(
         If not None, then apply linting just before building.
 
     convert_to_platforms : bool
-        If True, convert to platforms given in extra->convert->platforms
+        If True, convert to platforms given in extra->convert-to
     """
 
     if lint_args is not None:
@@ -204,7 +204,7 @@ def build(
 
     # convert to given platforms
     converted_pkgs = []
-    for platform in meta.get_value('extra/convert/platforms', []):
+    for platform in meta.get_value('extra/convert-to', []):
         for pkg in pkg_paths:
             try:
                 converted_pkgs.append(utils.convert(pkg, platform))
@@ -284,7 +284,7 @@ def build_recipes(
 
     convert_to_platforms : bool
         If True, try to convert packages to the additional platforms specified in
-        extra->convert->platforms. This will only work for pure Python packages.
+        extra->convert-to. This will only work for pure Python packages.
     """
     orig_config = config
     config = utils.load_config(config)
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -19,7 +19,7 @@ from . import linting
 logger = logging.getLogger(__name__)
 
 
-BuildResult = namedtuple("BuildResult", ["success", "mulled_images"])
+BuildResult = namedtuple("BuildResult", ["success", "mulled_images", "converted_pkgs"])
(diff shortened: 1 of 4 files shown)


bgruening:
@johanneskoester this can fail us as soon as one of the runtime requirements are not available on windows isn't it? So we really need to be careful with this feature.

melund:
@johanneskoester  I know you did this to make a windows package for snakemake.  But wouldn't it be a simple workaround to just let `snakemake` be a noarch package. 

```
build:
  noarch: python
```

johanneskoester:
@melund I don't think so, because noarch also implies independence of the python version as far as  I know. This is a bit unfortunate of course. Or do you know of a way to restrict to py 3k but use noarch?

@bgruening this is of course true. Maybe we can run a dependency check by faking a windows environment to conda.

johanneskoester:
@melund thanks a lot for that hint! I will give it another try. I thought that I tried that some time ago, but maybe it was a temporal bug that is gone now.

I have neither the right to restrict packages to bioconda nor is that my intention. In general, there is an agreement between Bioconda and conda-forge that bioinformatics stuff goes to Bioconda, as long as there is no particular need for windows support or other technical reasons to move stuff to conda-forge (this is why there is actually a lint wanring you on conda-forge when a package already exists in Bioconda). Moreover, in order to avoid confusion, moving a package to conda-forge entails that it should be removed from bioconda at the same time.

The Snakemake case is just a bit different though, because, as you might know, I am the author of Snakemake. In principle, I would not be against moving it over if there was a particular benefit. On the other hand, people expect Snakemake to be on bioconda, there is lots of documentation, blogposts, posters, etc. relying on this and so on. Hence, if this move would ever be made, it has to be well prepared and communicated. And since there are other ways to get windows-compatible builds (e.g., if now noarch works), I don't see currently the need. I am sorry if I was not so clear about that initially.

melund:
@johanneskoester I understand that people may expect that snakemake is available on bioconda. But since the install command include both channels `-c bioconda -c conda-forge`, moving to conda-forge would not break anything. You wouldn't even have to remove the bioconda packages, but just stop to build new ones. That way older versions would still be available.

melund:
If `norach` works that would be acceptable. Having to fall back to pip and manage dependencies our self is not. But the best would be not having to specify two channels when installing.

johanneskoester:
Ok, noarch seems to work indeed. Here is the PR: https://github.com/bioconda/bioconda-recipes/pull/12274

@melund that is only the case if people have the channel priority in the right order. Otherwise, they will get old versions of Snakemake without noticing. It is just hard to predict what confusion can occur. Since the technical reasons are now solved, I am happy to hear that we can agree on keeping things like they are for Snakemake.

In any case, thanks a lot for drawing my attention to this. It is indeed much better to have a decent windows installation method now!

@bgruening since noarch works (unlike I though) fine with pure-python but version-constrained packages, there is no need for the convert approach anymore, and this can be closed.

PR #359, opened 2018-11-28
Fix `noarch: False` triggers lint fail

epruesse:
Also add the name of the output causing the lint to the result (visible only in full report)


epruesse:
When using `outputs:`, conda will sometimes add `build: noarch: False` on its own, even if not specified in the raw recipe.

epruesse:
ping @daler - this ok?

daler:
Yes, thanks!

PR #365, opened 2018-12-07
Start supporting the BIOCONDA_LABEL env variable

dpryan79:
This will end up being used in `bulk` to add a gcc7 label to bioconductor packages.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -467,7 +467,7 @@ def build_recipes(
             for pkg in pkg_paths:
                 # upload build
                 if anaconda_upload:
-                    if not upload.anaconda_upload(pkg, label):
+                    if not upload.anaconda_upload(pkg, label=label):
                         failed_uploads.append(pkg)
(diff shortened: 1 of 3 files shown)


dpryan79:
I'm testing this now, so please don't merge yet (assuming it passes CI testing even)!

dpryan79:
This passes testing locally in that I can see it picking up the environment variable and adding it to the `anaconda upload` step. There are a few things about labels that should be kept in mind if others want to use them in the future:

 1. Obviously the channels in `bioconda-recipes/config.yml` need to be updated.
 2. The "see if a package needs to be built" function will default to looking in the first 2 channels in `config.yml`, so if you only want to build a subset of packages then you need to modify `bioconda-recipes/.circleci/config.yml` to use `--check-channels`. This obviously isn't necessary during a compiler migration (though if not everything needs to be rebuilt...).

In general this should be good to go. `bioconda-recipes` shouldn't be affected, since the environment variable isn't actually set there (yet).

PR #381, opened 2019-01-08
Autobump work

epruesse:
- allow checking only active updates
 - check CPAN and CRAN for dependency changes
 - add `get_raw` to `Recipe` class (access values before Jinja2 expansion)


epruesse:
closing in favor of #382 - no access to tokens coming from fork

PR #382, opened 2019-01-09
Autobump Nightly on CircleCI

epruesse:
- allow checking only active updates
 - check CPAN and CRAN for dependency changes
 - add `get_raw` to `Recipe` class (access values before Jinja2 expansion)
 - **Activate on CircleCI as nightly** and add as separate test
 - fix double spacing in our logs accidentally introduced by progress indicator


epruesse:
@bioconda/core This PR will make Autobump run on CircleCI nightly. Everyone OK with this?

bgruening:
That's so cool. The only objection I have is the time-interval. Is daily good? is weekly enough? Should there be a list with packages that should be updated, should there be a skip-list? I'm concerned about a lot of packages that no one is using ... with all bad implication as storage space and making it harder for the solver.

daler:
@bgruening we do still have the human intervention of reviewing auto-bumped PRs which could be used as a screening point (close PR to skip). I agree with worries about storage and solver, but for a skip list we'd have to choose some metric and threshold to determine whether a package should or should not be auto-bumped. I worry about being too strict on that list and alienating contributors.

As for time interval, to me weekly seems fine. But I don't really see a cost to daily frequency. Plus, a benefit of daily is that having updates the day after a release would go a long way in the perception of bioconda as an up-to-date source.

epruesse:
BTW: Here is a histogram of all 18k commit times in UTC:

out

It looks like least active hour is 4 AM UTC. That makes sense, given that it is 5 AM in Europe and 8 PM on the US west coast.

The 5 hours from 12 AM to 5 AM UTC are the least active period. So basically when the Pacific side of the US starts to go home (12 AM UTC = 4 PM PST) and before the Europeans start again (5 AM UTC = 6 AM CET).

epruesse:
(And by day of week, in case anyone else was curious)
out

bgruening:
Mh, I'm in a strange position here :) I actually don't want to argue against you, I mean I pushed this idea of this bot for so long time and this is awesome. Thanks a lot! But I also think we should discuss it and the implications of an automatic update of all packages every day ... what are the rules for merging, skipping etc ...

>  auto-bumped PRs which could be used as a screening point (close PR to skip)

Yes, but we do not have any model of who is in charge for this. We do not really have maintainers. What's currently happening is that the core team is merging them. Who can decide to skip a release?

> Going into the direction of what Ryan said, I was honestly even considering a more frequent scan (2x or 4x a day), so that authors would see the package published ASAP.

There is a value of slowing down processes/releases. People revoking releases, because of reasons (from version, number, readme not updated ...), certain versions are not meant for larger consumption ...
Democracy is for good reasons slow and maybe a super-fast release schedule is accumulating a lot of garbage?

> They do add to the length of the package index, which will be a problem eventually, but then that's something the conda guys will have to tackle - we can't do much there but postpone the problem a little.

Sure. But we do not need to make it unnecessary complex. How many packages will be accumulated that will not be used? Please be aware that we created an ecosystem that creates also containers, that are replicated, backup that people care about.

> The size of the packages that are updated often was pretty small in my observation. I've been downloading them over and over while running autobump, and it goes very quickly. The few large ones I haven't seen twice yet. We could watch this though if it becomes an issue.

We do also create containers, please don't forget that.

> Rather than have a skip list, I'd vote for moving more of the non-life-science packages to conda-forge (there are some active perl packages that don't really belong here and should be made available via CF).

I guess this is unquestionable and needs to happen either way. I have moved so many packages already, but it's time-consuming :(

With an auto-update list, we could ask the community about the packages that they care about. it's like asking for consent until we have a concept of a maintainer that can decide if this version is meaningful to be conserved for the eternity?

epruesse:
> Mh, I'm in a strange position here :) I actually don't want to argue against you, I mean I pushed this idea of this bot for so long time and this is awesome. Thanks a lot! But I also think we should discuss it and the implications of an automatic update of all packages every day ... what are the rules for merging, skipping etc ...

No worries - feedback is always welcome. 
 
> > auto-bumped PRs which could be used as a screening point (close PR to skip)
>
> Yes, but we do not have any model of who is in charge for this. We do not really have maintainers. What's currently happening is that the core team is merging them. Who can decide to skip a release?

I would generally be in favor of a bigger group in charge of merging. With 470 members we may need a more controlled group of senior members trusted to do the reviews and merge the PRs. 
 
> > Going into the direction of what Ryan said, I was honestly even considering a more frequent scan (2x or 4x a day), so that authors would see the package published ASAP.
> 
> There is a value of slowing down processes/releases. People revoking releases, because of reasons (from version, number, readme not updated ...), certain versions are not meant for larger consumption ...
> Democracy is for good reasons slow and maybe a super-fast release schedule is accumulating a lot of garbage?

Agreed. That was a factor in deciding on daily. I've had to mess with release tags I created myself, but that's usually done at the end of the day.

In theory it would be good to require that a release has a minimum age (say a day or two), but that would be complicated to implement across the various hosters, so it's on the "let's see if we need that" list.

Another way to deal with these things would be matching e.g. Debian's "Testing" distribution with a "testing" label in which new builds are quarantined for a few weeks. Packages labelled "testing" would only be available as "opt in" by users, and not guaranteed to be preserved forever.
 
> > They do add to the length of the package index, which will be a problem eventually, but then that's something the conda guys will have to tackle - we can't do much there but postpone the problem a little.
> 
> Sure. But we do not need to make it unnecessary complex. How many packages will be accumulated that will not be used? Please be aware that we created an ecosystem that creates also containers, that are replicated, backup that people care about.
> 
> > The size of the packages that are updated often was pretty small in my observation. I've been downloading them over and over while running autobump, and it goes very quickly. The few large ones I haven't seen twice yet. We could watch this though if it becomes an issue.
> 
> We do also create containers, please don't forget that.

I did. Not sure what implications they cause w.r.t. space constraints.

> 
> > Rather than have a skip list, I'd vote for moving more of the non-life-science packages to conda-forge (there are some active perl packages that don't really belong here and should be made available via CF).
> 
> I guess this is unquestionable and needs to happen either way. I have moved so many packages already, but it's time-consuming :(

Agreed. Maybe I can find some time to deal with more of that.

To start with, should we have a list of packages that need to be moved somewhere? Akin to the blacklist, those could be excluded from the updates easily.

> With an auto-update list, we could ask the community about the packages that they care about. it's like asking for consent until we have a concept of a maintainer that can decide if this version is meaningful to be conserved for the eternity?

I'm not sure how to populate a positive-list, in particular for new packages. Would a negative list (i.e. yet another blacklist) be an option in your opinion?

epruesse:
@bgruening I could add more information to the generated PRs. What do you think is relevant to making a decision? No sure how easy each item is, but I can think of:

  - total size of packages generated for last version - all archs, all outputs, one sum value
  - date of last update(s) - pub date on bioconda, should be available via anaconda api, could get the last two or so I think
  - size of container? - any idea whether quay has easily accessible data on this?

epruesse:
@bgruening Thinking about the containers - there are many packages that make no sense to have separately in a container - say `perl-gd` or `libarbdb`. So if the containers are an issue, should we perhaps instead of a white list for updates keep a white list for containers? We could probably populate it initially from usage counts, and then add `extra/bioconda/build_container: true` or something like that to them.

epruesse:
@bgruening - If you don't mind, I'll merge this once the tests check out to give it a test-flight. We can just disable the nightly runs after a few days. I just want to know that it works and check it of my list.

Also - this PR contains a fix for the TQDM issue.

PR #383, opened 2019-01-09
Let RepoData use config for determining which channels to retrieve.

johanneskoester:


johanneskoester:
cc @mikecormier

bioconda_utils/utils.py
```diff
@@ -1193,8 +1197,12 @@ class RepoData:
...
     #: Platforms loaded
     platforms = ['linux', 'osx', 'noarch']
-    #: Channels loaded
-    channels = ['conda-forge', 'bioconda', 'defaults']
+    # config object
+    config = None
+
+    @classmethod
+    def register_config(cls, config):
+        cls.config = config
```
epruesse: Should check that `__instance == None` to ensure the config is set before the singleton is instantiated.

bioconda_utils/utils.py
```diff
@@ -1213,6 +1221,13 @@ def set_cache(self, cache):
         else:
             self.cache_file = cache
 
+    @property
+    def channels(self):
```
epruesse: I'd do this test in `__init__()`, rather than adding things that "fail late" in a property.

bioconda_utils/utils.py
epruesse: Also consider `set_cache()`:

 - A change in channels invalidates the cache.
 - A single method setting both channels and cache is probably cleaner.
 - It may be easier to have a default for the channels than to set it everywhere.

johanneskoester:
@mikecormier this has been released as 0.15.4. Feel free to try with GGD.

PR #384, opened 2019-01-09
fixes https://github.com/bioconda/bioconda-recipes/issues/12988

bgruening:


epruesse:
This is also fixed by aac14fa7d3ffdd5bf3c5628fd4a23963a3f95438 in #382

epruesse:
There was an API change between 4.26 and 4.28 or so. The patch in #382 makes it work with either.

johanneskoester:
Since #382 is merged I guess this can be closed, right?

jfallmann:
Same for #12988 i guess then

bgruening:
I think this can be closed yes. @jfallmann please reopen if you still encounter issues.

Issue #385, opened 2019-01-14
Bioconda bot - playground issue


epruesse:

epruesse:
@bioconda-bot hello

epruesse:
Hello @epruesse!

epruesse:
@bioconda-bot hello

epruesse:
Hello @epruesse!

epruesse:
@bioconda-bot hello 1 2 3

epruesse:
Hello @epruesse!

daler:
@bioconda-bot hello!

epruesse:
Hmm. That’s taking too long.

epruesse:
@bioconda-bot hello

epruesse:
Ah - @daler: the exclamation mark threw it off

epruesse:
@bioconda-bot hello

epruesse:
@bioconda-bot hello

PR #386, opened 2019-01-14
Implement bot for Bioconda

epruesse:
@bioconda/core I've started building a bot for Bioconda. So far all it does is answer to `@bioconda-bot hello` in comments made on this repo.

Beyond "Conda-Forge has a bot, so we need one, too" , these are the use cases I see for the bot (some are just fancy ideas):
 
- Linting: 
  - Faster response to commits. It takes about 30 seconds to start (tried in #385) and is unaffected by CI backlog. That should take some of the frustration out of running against the linter.
  - Since the bot can have protected secrets, we can also make comments to all PRs, finally getting the messages in there.
  - With messages working, Lint warnings become useful.
- Merging: 
  - having the bot create merge commits should allow attributing everyone who helped complete a PR correctly
  - respond to a command "merge if currently running tests pass"
  - bulk pushing merges might save some CI time
- Rebasing: 
  - fix PRs that got out of date upon request

For now, I'll try to get it to lint recipes (upon request until it actually works).

## Help needed (Johannes?):
1. Who has control over @bioconda-bot? I'd need a personal access token to give to the bot so it doesn't sound like it's coming from me. I'd be happy to give access to the Heroku app so that whoever can do this can enter the token directly. (I've been racking my brain over a better name - something easily identified as the bot associated with bioconda but cute - without avail so far). 
2. Could someone create a webook in `bioconda/bioconda-recipes` pointing to `https://bioconda.herokuapp.com/_gh`? It's under Settings - I just don't have admin privileges to that repo. It would be more useful to do testing there, because the immediate use case is linting, and there isn't much to lint in this repo here...

## Changes to requirements.txt

- added gunicorn (webserver)
- added cachetools (caching github API requests)
- removed pydotplus (pulled in X11 and wasn't used as far as I could tell)

## Docker image
The app runs on Heroku using the container interface instead of the buildpack slugs to get around the size limit. Slugs can only be 500MB, which is hard to do with a conda installation. However, the existing `Dockerfile` builds a pretty large image (nearly 2GB) which didn't work on Heroku either. The `Dockerfile.web` uses multi-stage builds to create a leaner image:

1. It derives from `continuumio/miniconda3:4.5.11` which is only 422MB total, compared to the 1.06 GB of `conda-forge/linux-anvil`
2. After installing the bioconda-utils requirements, it removes a lot of stuff
   - all ELF binaries get stripped (cuts 40MB from pandas alone)
   - static libraries are deleted
   - package cache (`pkgs/`)is completely removed (patched binaries cannot be hardlinked, so this is another few hundred MB)
   - man files, docs, locales other than en_US are deleted
   - compiled python caches are deleted
   This layer only changes if the requirements change. It's adds 348MB for a total of 770MB. 
3. A temporary layer is used to create a wheel from bioconda utils. 
4. Only wheel is copied into the second layer where it is installed. This avoids adding all the files in this repo, and files that may be present locally (like a docs build) to the container. The layer is only 6MB (probably contains some compiled python), which is the only part that needs uploading to update the Heroku app. Much more reasonable this way.

## App

The app uses `gunicorn` as a webserver, `asyncio.web` to handle requests asynchronously and `gidgethub` to interact with github. So far it doesn't do much though. Try going over  to #385 and write `@bioconda-bot hello`. That's the extent of its capabilities for today.


johanneskoester:
@epruesse I have send you the token. I have also created the webhook. Currently, it tells me that the URL is not reachable, but I guess that is fine until you have set it up.

epruesse:
## Update

- The bot now uses the Github App API instead of acting as a user via an OAUTH token. As an app, the bot has an RSA key with which it signs a JWT to have temporary OAUTH tokens issued for the users and repositories it is installed in. Getting things to work with this required a little more infrastructure in the app, but now comments it posts will have the little bot symbol indicating it's status as a machine. More importantly, it now does not require a pseudo-account but has it's privileges assigned via normal permissions.
- Since some of the tasks will likely take a while to complete, the bot doesn't use `asyncio` to schedule tasks, but `celery`. The `celery` framework writes tasks into a message queue, in our case a RabbitMQ instance hosted by CloudAMPQ (largest free tier). A worker (running in the same dyno as the bot to save hours) then handles processing the tasks. The primary benefit here is that tasks can survive a restart of the app or its dyno/container.

It still doesn't do anything, except answer to
@bioconda-bot hello

epruesse:
a'right - I'm merging this so as to not diverge too much from master

PR #388, opened 2019-01-15
Bump tqdm version

yesimon:
Prevent this bug from occurring https://github.com/tqdm/tqdm/issues/457


epruesse:
This should already have been fixed (compatible with the broken state as well).

PR #389, opened 2019-01-16
Update cf pinning

johanneskoester:
For some reason, all old pinnings have been marked as broken.


bgruening:
> For some reason, all old pinnings have been marked as broken.

can you explain this. Please note that conda-forge switched the labels, so cf is offering the entire GCC7 stack since yesterday. We need to catch up as soon as possible. Not sure about site effects-

dpryan79:
The old pinnings are themselves marked with `broken` and no longer have a `main` label.

PR #390, opened 2019-01-16
update cf pinning

johanneskoester:
For some reason, all older pinnings have been marked as broken.


johanneskoester:
cc @mikecormier

johanneskoester:
@bioconda/core, does anybody of you have an insight into the reasoning behind that?

dpryan79:
@bgruening Is this related to the label change? That'd make some sense I guess.

dpryan79:
I think the downstream ramification from this if it's used in `master` is that everything will get built for gcc7/python 3.7 (as is already the case in `bulk`). If we want to forestall that a bit then adding `conda-forge/label/cf20190119` to the channels after conda-forge would be less drastic and not require bioconda dependencies in a PR to be rebuilt.

johanneskoester:
I won't be able to finish this in the next days (teaching and family duties). @dpryan79 your suggestion to temporarilty stick to the old label makes sense to me. If you or anybody else from core would do that, that would be really great! Sorry for not being able to do this myself.

dpryan79:
@johanneskoester No worries, I'll see if I can get this done tonight then.

dpryan79:
I'll close this since we have a temporary solution in place now.

PR #392, opened 2019-01-16
Partially revert pinning and start adding conda-forge/label/cf201901 to the channels

dpryan79:
A short(ish) synopsis of the problem: Many build are currently breaking because (A) the conda-forge-pinning version we use no longer has the `main` label and (B) the various `toolchain` packages also now lack this label. This means that typically bioconda-utils can't be installed and, that when it can (e.g., probably due to cache/CDN weirdness), anything that needs to be compiled will fail.

At the moment, everything in bioconda-recipes is using conda-forge-pinning==2018.09.20, so I've reverting to that release (I need to check what the differences are from 2018.11.27) so that nothing should be changing (yet) in master. In addition to the changes here, `conda-forge/label/cf201901` will need to be force-pushed to the master branch in bioconda-recipes (it should be after `conda-forge` in the list of channels).


dpryan79:
@bgruening Do you remember why you updated the pinning here https://github.com/bioconda/bioconda-utils/pull/366 ? Given the timing I wonder if it had to do with the bioconductor/R version changes.

dpryan79:
Ah, it was indeed the R pinning that changed between 2018.09.20 and 2018.11.27. But we've already set this to 3.5.1 in bioconda_utils-conda_build_config.yaml, so that shouldn't be an issue (if it were, we wouldn't have been building for R 3.5.1 only for so long).

bgruening:
As discussed during the call, I will merge this due to the lack of alternatives ...

Thanks @dpryan79!

PR #399, opened 2019-01-29
Fix broken conda FAQs link

erictleung:
Original broken link: http://conda.pydata.org/docs/faq.html
Changed to: https://bioconda.github.io/faqs.html


epruesse:
Thanks @erictleung!

docs/source/index.rst
```diff
@@ -70,7 +70,7 @@ package. The Python 3 version is recommended.
 .. seealso::
 
     * :ref:`conda-anaconda-minconda`
-    * The conda `FAQs <http://conda.pydata.org/docs/faq.html>`_ explain how
+    * The conda `FAQs <https://bioconda.github.io/faqs.html>`_ explain how
```
bgruening: I think the intention was really to link out to the conda documentation. The FAQ is not redirected to https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/index.html

Should we link to this page?

dpryan79: The tasks are probably about as close as it gets to an FAQ.

erictleung: Oh I didn't read around my change to notice it should be conda FAQs. I can replace it with the tasks link above if that's the consensus. Otherwise, maybe the getting started page might be a good alternative https://docs.conda.io/projects/conda/en/latest/user-guide/getting-started.html.

epruesse: There is no reason we need to keep calling it "FAQ"...

Also, since conda is now also using sphinx, the link could make use of `intersphinx` to determine the correct URL.

erictleung: I can remove the mention of "FAQ" to be more clear. @epruesse let me know if you've decided to use `intersphinx` instead and I can try to make the appropriate changes if they haven't already be done.

epruesse: Oh, I thought you could just add that :*)

epruesse:
Ok, moving on with intersphinx in #490

PR #401, opened 2019-01-31
docs: add r other sources guidelines

dlaehnemann:
This originates from reviewing the contribution of a non-CRAN, non-bioconductor R-package to bioconda, that threw up some questions for me. @jdblischak kindly provided some insights:
https://github.com/bioconda/bioconda-recipes/pull/13410#issuecomment-458959377

I hereby suggest to add these and some further infos to the docs, and would be grateful for further review from the other @bioconda/r people.

### DIFF
--- a/docs/source/guidelines.rst
+++ b/docs/source/guidelines.rst
@@ -227,6 +227,8 @@ in the meta.yaml`
 <http://conda.pydata.org/docs/building/meta-yaml.html#patches>`_.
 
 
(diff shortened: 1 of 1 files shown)


daler:
This looks good, thanks. I'll echo @jdblischak from that comment and say more docs are always better!

dlaehnemann:
Unless they get too detailed to quickly grasp / follow or confuse people... ;)

PR #406, opened 2019-02-01
Update pinning

dpryan79:
This (A) updates the conda-forge-pinnings to the (currently) most recent version and adds a `bioconda-utils update-pinnings` function that can be used to bump a given package and all dependencies as appropriate if and only if they are affected by the change in pinnings.

I also changed `bioconda-utils dag` so that it will show the DAG of all dependencies of the package(s) specified. The previous behavior was to filter the DAG for only the packages specified, which tended to leave a bunch of singletons and was probably now what people really wanted.

### DIFF
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -1,7 +1,7 @@
 anaconda-client=1.6.*
 argh=0.26.*
 beautifulsoup4=4.6.*
-conda=4.5.11
+conda=4.6.0
 conda-build=3.15.1
 galaxy-lib>=18.9.1
 jinja2=2.10.*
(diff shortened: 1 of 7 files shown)


dpryan79:
This PR now also:

 - Pins the docker version to a known functional docker container (before conda-forge added conda 4.6 to their containers)
 - Fixes the nightly builds such by adding bioconda back to the list of channels checked (so nightly doesn't keep trying to build and upload everything).

dpryan79:
@bioconda/core Can someone else have a look at this? This should resolve a number of issues that have popped up over the last couple days too (borked versions of conda in the docker container and nightly builds trying to build and upload everything).

bgruening:
Already on it ... testing the branch. From looking at the code it looks good.

bgruening:
Thanks @dpryan79! This fixes a lot of recent issues.

Issue #372, opened 2018-12-11
Show packages (outputs:) in recipe README


epruesse:
The docs currently assume that recipe == package. With `outputs`, a recipe can produce a bunch of packages. Those should be represented somehow as well.

epruesse:
packages.debian.org shows different pages for the "source package" and the "package", so one for the recipe and one for each output.

epruesse:
Fixed

Issue #332, opened 2018-09-10
Add link to cargo-port source to package docs


epruesse:
The documentation should add a generated link to the backed-up source code on `cargo-port`.

ping @bgruening Can you provide a bit of code that generates the link? You're more familiar with `cargo-port`...

bgruening:
With documentation you mean the tool card, like here? https://bioconda.github.io/recipes/abeona/README.html

epruesse:
@bgruening Yep

epruesse:
This would be best fixed by fixing #21

PR #408, opened 2019-02-02
Docs Update

epruesse:
See http://pruesse.net/tmp/bioconda/ for a demo-build.

### Test cases
- multiple outputs: [arb-bio](http://pruesse.net/tmp/bioconda/recipes/arb-bio/README.html)
- `license_family`: [vsearch](http://pruesse.net/tmp/bioconda/recipes/vsearch/README.html)
- `license_url`: [r-prestor](http://pruesse.net/tmp/bioconda/recipes/r-prestor/README.html)
- recipe-name != package name: [r-prestor](http://pruesse.net/tmp/bioconda/recipes/r-prestor/README.html)
- no primary meta.yaml: [tqdist](http://pruesse.net/tmp/bioconda/recipes/tqdist/README.html)

### Todo
- [x] Fix "Quick Search"
- [x] Fix Copyright
- [x] Keep sidebar always visible
- [x] Show `packages` separate from `recipe` (`outputs:`) #372
- [x] Replace `Available Packages` table
- [x] Add packages to search with high prio
- [x] Use latest build, not just latest package, for shown packages and depends
- [x] Show `required-by` in package card
- [x] Fix quay installation text
- [x] Show dependencies in package card
  - [x] Link to conda-forge feedstocks
     - [ ] fix case where package is output and doesn't have feedstock
  - [x] add icon showing source
- [ ] Show source package backup at cargo-port #332 
- [ ] Highlight most current version
- [ ] Show summary in italics
- [ ] Show if blacklisted
- [ ] Show linter errors?
- [ ] Show update status 
- [ ] Show detected upstream releases url
- [ ] Fix recipes not rendered due to DuplicateKey errors in the Json
- [ ] Fetch citation for DOI links

### RFC

- I'm generating the info for each package from the latest version in the channel. Technically, I could render all versions, possibly one-per-page or on an 'previous versions' page. I'm just not sure I want to create that many pages. Unless someone has a use case for this info?
- Linking to the right conda-forge feedstock won't be easy. I could take the data from the regro repo where they build the graph and have all the about.json's in one place to determine this. But that's the only idea I've had so far. Alternatively, I can just check if the repo exists and only link if that is the case. Right now it generates a lot of 404 links (just try a compiler link - they all point into nirvana).
- For incorporating autobump information I plan on passing that data as CSV from the autobump job to the docs-build job via CircleCI's workspace thingy.
- While parsing all the recipes, it might be interesting to create a machine-parseable output. Technically, `channeldata.json` is supposed to contain all that's needed, but then we have little control over that. Need someone with an actual use-case though.


docs/source/templates/readme.rst_t
```diff
@@ -37,6 +37,10 @@ and update with::
 
    conda update {{ name }}
 
+or use the docker container::
+
+   docker pull quay.io/repository/biocontainers/{{ name }}
```
bgruening: We do not have a :latest container, so a user needs to specify a version. 

what about `docker pull quay.io/repository/biocontainers/{{ name }}:<revision>` and maybe even link the revision to the `https://quay.io/repository/biocontainers/{{ name }}?tab=tags`

epruesse: In theory I could pull the latest revision via Quay's API ...

epruesse: But yeah - I guess leaving that to the user is better. You don't have the latest tag intentionally, right?

bgruening: Yes intentionally ...

epruesse:
There are semantic, practical and communication reasons for separating the recipe from the package(s) it produces. The two things are different. The recipe is (should be) equivalent to an upstream source, the packages are units in which the artifacts produced are distributed. We currently don't make much use of `outputs:` because it isn't very useful for Python or R packages. For the larger things it does make sense. E.g. `snakemake-minimal` and `snakemake` should be outputs of the same `snakemake` recipe.

Technically, the "package" doesn't even exist as a unit. Each version can be different. I'm just not sure how to display that properly.

The order is simply because the packages are part of the recipe. 

I was considering to merge the two if there is only one output and the two share names. But I can't easily get sphinx to do that. So it's separate for now.

From the index side, it should be easy to remove the packages that have a recipe of the same name.


>     * In fact, given the existing content, I'm not convinced that package and recipe need to be separate. However if more info were added to recipe (blacklist status; update status, linting errors) then that would justify a separate section.

That is the plan. :)

>     * While messing with these pages, the "2. Set up channels" can be changed to have better link text.
> 
>     * Will the "Recipe Index" eventually replace the "Available Packages" link at the top of the sidebar? I think we only need one place to look stuff up.

Yes. The "available packages" is empty - I just need to clean that out. 

> 
>     * For the Recipe Index, what will go where the "(extra)" and "(qualifiers)" text are?

Not sure. Sphinx offers those places to put extra information when I generate the index. osx/linux/noarch, latest-version, python/r/binary come to mind. I'm open to suggestions though. I'm guessing they disappear if I don't set anything at all.

Anyway - for today I got the back-references working. Each package now has a `required by` item showing which other (of our) recipes require it.

daler:
Got it, that convinces me that recipe/package should be separate and this is the correct order :)

> Sphinx offers those places to put extra information when I generate the index. osx/linux/noarch, latest-version, python/r/binary come to mind.

Agreed, those would be useful. Maybe py/r/binary/perl, if we're keeping Perl here rather than CF. Possibly license now that you're parsing that out, but that might get too cluttered.

Required-by is a nice touch!

epruesse:
Well ... I'm of course open to suggestions. I had originally thought to just have the `package` entry if the `package` and the `recipe` actually differ, but it just didn't fit well into the design of sphinx. We could have the package and the recipe on different pages, and just re-iterate the parts from the `about` section for example.

Incidentally - I wish the `outputs` section had an additional `about` section to add to the description/summary. Wonder if that works....

epruesse:
Ok to merge this? I'll move the open items at the top into a separate issue and work on them later.

Or do you guys have things on your wishlist that should be fixed first?

PR #417, opened 2019-02-09
Refactor update pinnings

epruesse:
@dpryan79 ... Working on cleaning this up to be integrated into the bot.

Do you happen to have test cases?


dpryan79:
I don't really have any test cases, I just used recipes that I knew needed updating (or not) and checked to see if they were getting bumped appropriately. The ideal test case would be to use a custom conda_build_config.yaml with some sort of odd numpy version and C compiler that don't exist I guess (combined with recipes that either use or don't use them).

bioconda_utils/cli.py
```diff
@@ -497,18 +526,18 @@ def dag(recipe_folder, config, packages="*", format='gml', hide_singletons=False
      help="""Bump package build numbers even if the only applicable pinning
      change is the python version. This is generally required unless you plan
      on building everything.""")
-def update_pinning(recipe_folder, config, packages="*", skip_if_in_channels=['bioconda', 'bioconda/label/gcc7'], bump_only_python=False):
-    """
-    Bump a package build number and all dependencies as required due to a change in pinnings
+@enable_logging()
+def update_pinning(recipe_folder, config, packages="*",
+                   skip_if_in_channels=['bioconda', 'bioconda/label/gcc7'],
```
dpryan79: Can you remove `bioconda/label/gcc7`? We can now exclude that from everywhere in the code.

epruesse: Removed both, actually. You can now only add to the channel list. We don't need to bump stuff that's got the right build number in another channel.

epruesse:
@dpryan79 - can you give this branch a spin and see if you like it?

dpryan79:
@epruesse Will do!

dpryan79:
For `--skip-additional-channels`, can we make the meaning a bit more explicit: "Skip updating/bumping packges that are already built with compatible pinnings in one of the given channels in addition to those listed in 'config'."?

bioconda_utils/update_pinnings.py
```diff
@@ -1,88 +1,125 @@
...
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
 buildNumberRegex = re.compile("^( +number: +)(\d+)(.*?)$")
 buildNumberJinja = re.compile("^({% +set build +=.+?)(\d+)(.*?%})")
 
-def already_bumped(r, name, version, buildNumber):
-    rsub = r.loc[(r.name == name) &
-             (r.platform != "osx") &
-             (r.version == version)]
```
epruesse: @dpryan79 Using `RepoData().get_package_data()` is actually a lot faster than using `RepoData.df.loc`. The `df` property should really be private, actually. Part of the point of the class was to keep the logic working on the DF in one place, so that if its contents need to change, only one place needs to be considered.

bioconda_utils/update_pinnings.py
```diff
@@ -1,88 +1,125 @@
...
-        return True
-    return False
 
+def will_build_variant(meta):
+    """Check if the recipe variant will be built as currently rendered"""
+    build_numbers = RepoData().get_package_data(
+        'build_number',
+        name=meta.name(), version=meta.version(),
+        platform=['linux', 'noarch'],
+    )
```
epruesse: This says "get me `'build_number"` for packages where `name=meta.name()` AND ...".

bioconda_utils/update_pinnings.py
```diff
@@ -1,88 +1,125 @@
...
-        return True
-    return False
 
+def have_variant(meta):
+    """Checks if we have an exact match to name/version/buildstring
+    """
+    res = RepoData().get_package_data(
+        name=meta.name(), version=meta.version(), build=meta.build_id(),
+        platform=['linux', 'noarch']
+    )
```
epruesse: without any `key` set, the result is `exists`. The key could also be a list (`['version', 'build'], name='htslib'`), in which case the pandas tuple iterator is returned.

bioconda_utils/update_pinnings.py
```diff
@@ -1,88 +1,125 @@
...
-    (bumped) due to a recent change in pinnings.
+    def strip_py(build):
+        if build.startswith("py"):
+            return build[4:]
+        return build
+
+    builds = RepoData().get_package_data(
+        'build',
+        name=meta.name(), version=meta.version(),
+        platform=['linux', 'noarch']
```
epruesse: @dpryan79 - Why did you exclude OSX from the consideration?

dpryan79:
Also, we absolutely need a way to set the number of forks this thing uses and the default should be 1. I expect I'm not alone in routinely working on systems with 100+ cores and multiple other users, many of whom will be quite ticked if I start suddenly using all of the cores.

bioconda_utils/utils.py
```diff
@@ -50,11 +55,24 @@ def emit(self, record):
...
+
+    Args:
+      loglevel: logging loglevel (the number, so logging.INFO)
+      logger: local logger (in case it has different effective log level)
+    """
+    term_ok = (sys.stderr.isatty()
+               and os.environ.get("TERM", "") != "dumb"
+               and os.environ.get("CIRCLECI", "") != "true")
+    loglevel_ok = (kwargs.get('logger', logger).getEffectiveLevel()
+                   <= kwargs.get('loglevel', logging.INFO))
```
epruesse: @dpryan79 - This makes the progress monitor now silent if you pass `--loglevel warn`.

bioconda_utils/cli.py
```diff
@@ -20,11 +21,49 @@
...
+    >>> @enable_logging('warn')
+    >>> def command():
+    >>>   pass
+
+    Args:
+      default_loglevel: loglevel used when --loglevel is not passed
+    """
+    def decorator(func):
+        @arg('--loglevel', help="Set logging level (debug, info, warning, error, critical)")
+        @arg('-P', '--pdb', help="Drop into debugger on exception")
```
epruesse: Probably should be added via it's own decorator, but I really wanted the "if you crash, drop into PDB" feature.

bioconda_utils/cli.py
```diff
@@ -489,58 +516,83 @@ def dag(recipe_folder, config, packages="*", format='gml', hide_singletons=False
...
+    config = utils.load_config(config)
+    if skip_additional_channels:
+        config['channels'] += skip_additional_channels
+
+    if cache:
+        utils.RepoData().set_cache(cache)
+    utils.RepoData().df  # trigger load
+
     build_config = utils.load_conda_build_config()
-    build_config.trim_skip = True
```
epruesse: This is actually the default

bioconda_utils/cli.py
```diff
@@ -489,58 +516,83 @@ def dag(recipe_folder, config, packages="*", format='gml', hide_singletons=False
...
     # Print some information
     print("Packages requiring the following:")
-    print("  No build number change needed: {}".format(updated[0]))
-    print("  A rebuild for a new python version: {}".format(updated[1]))
-    print("  A build number increment: {}".format(updated[2]))
-    if updated[3]:
-        print("{} packages produced an error in conda-build: {}".format(updated[3], list(hadErrors)))
-    if len(bumpErrors):
-        print("The build numbers in the following recipes could not be incremented: {}".format(list(bumpErrors)))
+    print(stats)
```
epruesse: Yeah - stats aren't as pretty as yours anymore. Might fix that later.

bioconda_utils/graph.py
```diff
@@ -0,0 +1,141 @@
...
+                get_inner_deps(get_deps(meta, "build")),
+                get_inner_deps(get_deps(meta, "host")),
+                get_inner_deps(get_deps(meta, "run")),
+            ))
+        )
+
+    return dag, name2recipe
+
+
+def build_from_recipes(recipes):
```
epruesse: This DAG is different from the original one in that it is strictly built on recipes. Before there was some confusion between the two, which becomes wrong in the presence of `outputs:`. It also just uses the `Recipe` objects as graph nodes, so no need to have a side-car dictionary to go with the graph.

bioconda_utils/update_pinnings.py
dpryan79: In practice everything in OSX is in Linux (with few if any exceptions) and the pinnings are almost entirely identical, so I figured I'd save the time. We can add OSX in, it'll just increase the run time slightly. I guess if someone were to run this on OSX it would currently incorrectly think that almost everything should be updated, so I suppose that's a compelling argument for adding `osx` to that list of platforms.

epruesse:
> For `--skip-additional-channels`, can we make the meaning a bit more explicit: "Skip updating/bumping packges that are already built with compatible pinnings in one of the given channels in addition to those listed in 'config'."?

Do you mean the help or the command line option? Simple enough ... just tell me which name. ;)

Also - might want to change the default and name on `bump-python-only`. It should be on by default I think and it should not sound like it's only going to consider python packages...

> Also, we absolutely need a way to set the number of forks this thing uses and the default should be 1. I expect I'm not alone in routinely working on systems with 100+ cores and multiple other users, many of whom will be quite ticked if I start suddenly using all of the cores.

You'll be in the minority I think. It also does obey cgroups settings, so on a typical cluster you'd only run on the CPUs you reserved. Still, a max threads would be useful, as would a "no threads" (for debugging). 

> Other than that I'm quite liking your changes so far (I'll keep playing around with it tomorrow morning my time)!

epruesse:
> In practice everything in OSX is in Linux (with few if any exceptions) and the pinnings are almost entirely identical, so I figured I'd save the time. We can add OSX in, it'll just increase the run time slightly. I guess if someone were to run this on OSX it would currently incorrectly think that almost everything should be updated, so I suppose that's a compelling argument for adding `osx` to that list of platforms.

I'll take that out then. It doesn't make a difference for the test cases I have, and performance isn't an issue - `api.render` takes almost a minute per package. It feels more `correct`.

epruesse:
What definitely needs fixing is the sub-recipes. It should only bump subrecipes if they are actually required from somewhere. Or, as first fix, ignore all but the most recent one as we do in the docs.

bgruening:
>  Or, as first fix, ignore all but the most recent one as we do in the docs.

:+1:

epruesse:
ok - max threads has been added

dpryan79:
> Do you mean the help or the command line option? Simple enough ... just tell me which name. ;)

Just the help. Feel free to redo `--bump-only-python` however you like, we could just get rid of it completely.

Thanks for adding a threads option!

epruesse:
This was enough for one PR.

PR #421, opened 2019-02-13
Refactor bioconductor test

dpryan79:
This should speed up the bioconductor tests by ~33% (at least it does in testing). These still rely heavily upon internet sites, which are the slowest components (most of the time is spent waiting for sites to respond).


dpryan79:
Ping @epruesse

epruesse:
Thanks!

PR #427, opened 2019-02-26
Remove conflicting information about conda-build 2

sndrtj:
The guidelines for python still mentioned the use of conda-build 2.x over conda-build 3. Bioconda now uses conda-build 3. This PR rectifies the docs to point users to the correct version of conda-build for use with bioconda. This should fix https://github.com/bioconda/bioconda-recipes/issues/10311

It also fixes a line about noarch packages. Per the linter, noarch packages are required for pure python packages. I'm not sure in what other circumstances noarch packages are allowed / required. If someone could chime in on there, that would be great :-).

### DIFF
--- a/docs/source/cb3.rst
+++ b/docs/source/cb3.rst
@@ -1,3 +1,5 @@
+.. _cb3-main:
+
 Conda build v3
 --------------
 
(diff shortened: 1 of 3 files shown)


dpryan79:
Thanks @sndrtj! I took the opportunity to make a few more updates to the documentation.

PR #429, opened 2019-02-27
Fix broken conda.io links

stain:
Not sure why this was broken at conda.io/docs as the new links seem more fragile - see conda/conda-docs#666


epruesse:
using intersphinx to keep those ok now

PR #430, opened 2019-03-04
Use curl instead of wget for data packages.

johanneskoester:
This fixes issue bioconda/bioconda-recipes#13846.


johanneskoester:
The linux test passed, but circle ci was unable to trigger a status update because of github rate limiting. I'll merge this manually.

PR #433, opened 2019-03-07
Provide a subcommand to commit subdags.

johanneskoester:
Inside a modified branch, you can now run e.g.

```
bioconda-utils commit-subdags recipes/ config.yml -k 1 -n 10 --msg "Fixing bioconductor datapackages"
```

Bioconda-Utils will then create a commit with the k-th of n subdags of changed recipes, properly considering dependencies.
@bioconda/core, this could be a potential replacement for the bulk branch in the future. Let me know your thoughts.
I used this to create PR bioconda/bioconda-recipes#13946.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -319,54 +319,11 @@ def build_recipes(
     failed = []
     skip_dependent = defaultdict(list)
 
-    # Get connected subdags and sort by nodes
-    if testonly:
-        # use each node as a subdag (they are grouped into equal sizes below)
-        subdags = sorted([[n] for n in nx.nodes(dag)])
-    else:
-        # take connected components as subdags, remove cycles
-        subdags = []
-        for cc_nodes in nx.connected_components(dag.to_undirected()):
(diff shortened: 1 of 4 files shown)


daler:
@johanneskoester, revisting this after a while . . . 

Nowadays the [bulk branch builds separate subdags](https://github.com/bioconda/bioconda-recipes/blob/master/.github/workflows/Bulk.yml#L52) which, if I understand the rationale for this PR correctly, satisfies the same immediate need as this PR.

In terms of replacing bulk altogether, I kind of like that idea just because it's one less moving part to document and worry about. However an advantage of bulk as it stands now is that it builds on GitHub Actions while everything else is on Azure, so it's on a different queue and quota. It would be good for a `bulk` replacement to be able to run on a different queue.

That could be handled here by adding special commit strings to signal that it should be built elsewhere -- and then Azure would have to not pick it up and GitHub Actions would need to pick it up. But I feel like that would introduce more complexity than the bulk solution we currently have which simply depends on branch name.

PR #451, opened 2019-03-14
Add README.md to bioconda.github.io

epruesse:
We don't see issues posted over there, or people creating PRs against the computed `rst`s

Unless anyone objects, I will also disable issues and PRs on that repo once I've cleaned the existing items out.


dpryan79:
:+1: We should mention on the bioconda-recipes to submit issues over here then.

PR #499, opened 2019-04-13
Just use the most recent docker image

dpryan79:
Ping @bgruening @epruesse 

We can then tag a new release.


dpryan79:
Looks like we need to update the conda version in the container. I'm not sure what ramifications that will have in terms of API changes.

dpryan79:
Hmm, some of the docker-related tests are failing. I'm unlikely to have a chance to look at that until later tonight.

dpryan79:
I suspect prebuilding a new `bioconda-utils-build-env:latest` would get around most of these errors, since that could have a compatible `conda` version inside.

bgruening:
@dpryan79 latest should be up-to-date imho. Its 8h old.

https://hub.docker.com/r/bioconda/bioconda-utils-build-env/tags

dpryan79:
It's up to date with an old conda version and fails CI testing if we try to use it, yes. My changes effectively bootstrap things such that we test with the container that will eventually be uploaded.

dpryan79:
I won't pretend to understand why some of these error are so intermittent when they don't appear to rely on networking. Anyway, all green now.

bgruening:
Ok, then let's try this and see what this yields in bioconda-recipes :)

PR #502, opened 2019-04-16
Autobump: add update pinnings (fixes #423)

### LINKED ISSUE #423: Combine update-pinings with autobump
I'm picturing about this:
 - iterate over recipes in topo order, starting at ancestors
 - check if they need version-bump OR just pinning-rebuild
 - skip descendants of recipes for which PRs exist
   - mark the recipes blocking other bumps

So if I see this right, I need to change the way recipes are iterated. Instead of random shuffle, I'll go by the graph, which means pre-loading all recipes. This creates some questions re handling of recipes with updates in progress. I think using the putative, new state of the recipes is the right way, although I don't see actually possible situations in which it would make a difference.

Do we stick to the bulk-approach for "doing migrations" (I think that's CF lingo for updating pinnings), or do we come up with an iterative way of doing this with individual builds and a special label to keep packages in until the update is complete?

epruesse:
Issue: #423

### Changelog:
- renamed module async to aiopipe (fixes py 3.7 bug)
- renamed module update to autobump (for consistency)
- ignore run dependencies for graph building (shouldn't need those for building/pinning)
- allow turning off shuffle (`--no-shuffle`)
- allow turning off version update (`--no-version-update`)
- process recipes in graph order (disable with `--no-follow-graph`)
- check recipes for pinning updates (disable with `--no-check-pinnings`)
- skip recipes with pending updates on their deps (disable with `--no-check-pending-deps`)
- add machine parsable info to PRs via HTML comment
- handle Ctrl-C better with running subprocesses

- [x] supply separate PR message for pinning updates


epruesse:
Ok, so some research RTFS on conda-build from [metadata.py](https://github.com/conda/conda-build/blob/master/conda_build/metadata.py)

- The MetaData class has a function `hash_dependencies()` that returns the hash for the package file name. 
- The function `get_hash_contents()` returns the raw data that is hashed.
- This data is a dict mapping the dependencies that are in `build` somewhere in the recipe and are pinned in `conda_build_config` to a specific value (not a range) to that value.

That means that `run_exports` and `pin_compatible` settings are ignored. It also means that it is almost moot for us to consider the graph at all. We pin only four packages: htslib, libdeflate, r_base and bamtools.

epruesse:
It might also explain why Conda-Forge's regro-auto-tick bot uses very manual looking "migrators" to do package bumps:

https://github.com/regro/cf-scripts/blob/7144ed74dd3af3dd7e5369104bddcd676de42726/conda_forge_tick/auto_tick.xsh#L223-L224

Every time they have a package that requires downstream rebuilds, they add a piece of code that walks the graph and adds the packages.

Issue #503, opened 2019-04-17
circleci fails and scripts do not run on macOS, Ubuntu


crashfrog:
Overall the consensus seems to be that 

https://github.com/bioconda/bioconda-recipes/issues/13932

actually belongs over here:

```
$ circleci build
Docker image digest: sha256:65b2102646d5658f892e0ad8253b7912c676126c857c87c8c12460f0aa4f5aa1
====>> Spin up Environment
Build-agent version 1.0.8563-43047892 (2019-03-06T15:11:54+0000)
Starting container bioconda/bioconda-utils-build-env
  using image bioconda/bioconda-utils-build-env@sha256:1f62a95df52ff1f900f14e391d2925f400d97463c2071f840b0d649a7a5af088

Using build environment variables
  BASH_ENV=/tmp/.bash_env-localbuild-1551890040
  CI=true
  CIRCLECI=true
  CIRCLE_BRANCH=pyani
  CIRCLE_BUILD_NUM=
  CIRCLE_JOB=build
  CIRCLE_NODE_INDEX=0
  CIRCLE_NODE_TOTAL=1
  CIRCLE_REPOSITORY_URL=https://github.com/widdowquinn/bioconda-recipes.git
  CIRCLE_SHA1=4e42d36a20e979535651040bd28505bc39adb20d
  CIRCLE_SHELL_ENV=/tmp/.bash_env-localbuild-1551890040
  CIRCLE_WORKING_DIRECTORY=~/project

====>> Checkout code
  #!/bin/bash -eo pipefail
mkdir -p /root/project && cd /tmp/_circleci_local_build_repo && git ls-files -z | xargs -0 tar -c | tar -x -C /root/project && cp -a /tmp/_circleci_local_build_repo/.git /root/project
/bin/bash: git: command not found
```

I can confirm this problem in both Mac OS and Ubuntu.

epruesse:
The most recent SHA is `359113e96c2c6b85583c04c0b4f0b084019677ba2a1c9d48bc15e467ba2f9964`. Can you do `docker pull bioconda/bioconda-utils-build-env:lastest` and try again?

crashfrog:
Yeah, that's the one I'd pulled:

Digest: sha256:359113e96c2c6b85583c04c0b4f0b084019677ba2a1c9d48bc15e467ba2f9964

Did CircleCI run with a different image? I hadn't noticed at the time, but looks like:

```
[...]

PR #504, opened 2019-04-17
Update `circleci build` docs

epruesse:
See #503


epruesse:
Hmm. This part of the docs seems to be somewhat redundant. Anyone able to update this? It would be great if it was someone for whom figuring those things out isn't too far in the past. Easier to write helpful docs if you haven't forgotten the gotchas that get you when you start with conda.

PR #512, opened 2019-04-29
Create our own container for mulled testing.

dpryan79:
This won't pass testing until the container is actually uploaded of course. `MulledDockerfile` is an updated version of `continuumio/miniconda3` with a new conda inside. This container fixes issues like https://github.com/bioconda/bioconda-recipes/pull/14472 locally for me at least (I can reproduce those issues locally with the default `continuumio/miniconda3` container).

### DIFF
--- a/MulledDockerfile
+++ b/MulledDockerfile
@@ -1,0 +1,27 @@
+FROM debian:latest
+
+#  $ docker build -f MulledDockerfile -t bioconda/mulled_container:latest
+#  $ docker run --rm -it bioconda/mulled_container:latest /bin/bash
+#  $ docker push bioconda/mulled_container:latest
+
+ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
+ENV PATH /opt/conda/bin:$PATH
+
+RUN apt-get update --fix-missing && \
+    apt-get install -y wget bzip2 ca-certificates curl git && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-4.6.14-Linux-x86_64.sh -O ~/miniconda.sh && \
+    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
+    rm ~/miniconda.sh && \
+    ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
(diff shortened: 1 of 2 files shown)


dpryan79:
I switched this to use a copy of the image that I've uploaded into my own account. That obviously should be changed before merging!

dpryan79:
OK, this appears to be passing testing now. Obviously the container should be moved still. @bgruening @mbargull thoughts?

bgruening:
Looks good to me for the time being. I'm also fine to use your container. We assume that they offer an official container soon, isn't it?

dpryan79:
It doesn't sound like we should be holding our breath about them updating the containers without changing other things in incompatible ways. Do you want to instead upload that container to the bioconda channel? I guess it doesn't much matter, it just looks more proper :P

dpryan79:
Otherwise I'll merge and update stuff tonight after dinner.

bgruening:
Merge, if I find time I will configure a new project in Dockerhub to autobulid this one ... but for the time being we can just use yours.

Thanks Devon!

dpryan79:
OK, I'll wait to tag a release on the resolution to #514

mbargull:
Just saw this now :O. Yeah, totally fine by me to use your image for now!
I think it makes sense for use to create a `bioconda/container-images` GitHub repository where we can transfer `bioconda/extended-base-image` and this one to. I'll see (this/next week) if I can scrap together/cleanup some of the local container image build files I use myself and put them up on GitHub. Some small-ish containers and for running `conda`.
I can't right now, but I'll checkout/review your image tomorrow, probably. Depending on how the image is used (whether we run `bash` or something else that does not source the shell init script), we might need to alter the entrypoint a bit to accommodate for `conda >=4.6` activation stuff. (Also, it probably makes sense to use something like `debian:9-slim` instead of `debian:latest` -- but either one is of course fine for now!)

mbargull:
> I'll checkout/review your image tomorrow

Okay, I made sure where/how we use that image. AFAICT, everything should work properly, so nothing crucial we have to change right away!
We only use the image via `mulled` -> `involucro` at https://github.com/galaxyproject/galaxy-lib/blob/19.5.2/galaxy/tools/deps/mulled/invfile.lua#L89. Hence, we run it with `sh -c ' ... conda install ...'` which means the activation scripts are not `source`d, but due to `ENV PATH /opt/conda/bin:$PATH` and the fact that we don't run anything else apart from `conda install`, that is fine.

PR #514, opened 2019-04-30
Update bamtools pinning to a version that's been rebuilt

dpryan79:
2.4.1 was built with the old compilers, so everything that links against it can't get updated. The most recent version of bamtools we've built is 2.5.1, so I suggest pinning to that instead.

### DIFF
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -13,7 +13,7 @@ pin_run_as_build:
 htslib:
   - 1.9
 bamtools:
-  - 2.4.1
+  - 2.5.1
 libdeflate:
   - 1.0
 r_base:


dpryan79:
We'll need to push a new release after merging this.

dpryan79:
2.5.1 isn't currently packaging the shared library.

bgruening:
Is rebuilding the old version an option?

epruesse:
What changed?

dpryan79:
In the bamtools release notes it's mentioned that proper support for C++11 and C++14 was added in version 2.4.2 and support for long reads (those with >64k CIGAR ops) was added in 2.5.0. Given that we'd probably need to move to 2.4.2 to get the thing to compile.

epruesse:
Might as well go to 2.5.1 then. We build on C++17 now, so we'd want all fixes, lest we'd have to do a lengthy backport. Easier to rebuild things.

dpryan79:
OK, I'll expect to spend part of tomorrow getting things that depend on bamtools updated. I'll also move it to the build section on a lot of packages, since most of them are using libbamtools.a.

PR #522, opened 2019-05-06
Add realtime command log output (with max) and log file

epruesse:
- Adds three options to all commands that support `--loglevel`:
  - `--logfile`: If given, output is also logged to this file
  - `--logfile-level`: Specify logging level for logfile. Default is DEBUG
  - `--log-command-max-lines`: If specified, *console* logging output from `utils.run` will be    truncated after *n* consecutive lines.
- Logs stdout/err live as `(ERR) msg` and `(OUT)` 
- Disables `tmp-bioconda-builder` unless explicitly requested via `--build-image`

This replaces #167

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -367,7 +367,7 @@ def build(
     if setup:
         logger.debug("Running setup: %s" % setup)
         for cmd in setup:
-            utils.run(shlex.split(cmd))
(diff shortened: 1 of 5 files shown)


epruesse:
Scenarios:
 - at users desk: default, no logging limit
 - PR builds: `--log-file /tmp/artifacts/log.txt --log-command-max-lines 5000`
 - bulk with `--log-command-max-lines 0`

PR #527, opened 2019-05-07
Properly sort semantic versions of bioconductor and change cargoport interaction

### LINKED ISSUE #525: bioconductor-skeleton still picking up 3.8 rather than 3.9
For some reason, `bioconda-utils bioconductor-skeleton --force ... update-all-packages` is still picking up bioconductor 3.8 as the most recent release. 3.9 can be forced, but that shouldn't be required.

dpryan79:
This should close #525. This issue has only hit us since we in to double digit minor release versions of bioconductor now, but '3.10' should be sorted before '3.9' when we get the list of releases.

### DIFF
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -345,7 +345,7 @@ class BioCProjectPage(object):
         it exists.
         """
         url = cargoport_url(self.package, self.version, self.bioc_version)
-        response = requests.get(url)
+        response = requests.head(url)
         if response.status_code == 404:
             # This is expected if this is a new package or an updated version.
             # Cargo Port will archive a working URL upon merging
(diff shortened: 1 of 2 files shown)


dpryan79:
@bgruening When `bioconda-utils bioconductor-skeleton` checks for the existence of packages on depot.galaxyproject.org it's currently using a `get` request, causing the whole package to be downloaded. For other sources, it's doing a simple `head` request. Since it can take 20 minutes to to download larger (e.g., BSgenome) files, this seriously slows things down. Is there a reason for this on the cargoport side of things or can I safely change it to a `head` request?

dpryan79:
Ah, the failed test is due to `affydata` requiring an `affy` version that obviously can't exist yet. We should just pick a different experiment data test.

dpryan79:
N.B., this modified version was used in https://github.com/bioconda/bioconda-recipes/pull/15160

epruesse:
Question on the side - do we need to use actual recipes and data in the tests, or can we use mocks? Actually building and downloading and rendering seems to be what makes the tests so slow.

dpryan79:
The benefit to using real packages is we catch things when/if bioconductor changes formats.

bgruening:
@dpryan79 I think this is a left-over from the early days where we checked if the package is actually >0b. head should be fine imho.

dpryan79:
OK, thanks, I'm happy that fixes things since it was taking > 1 day to create the package recipes otherwise.

PR #531, opened 2019-05-08
Point contributors to conda-forge for CRAN packages (for CRAN packages without Bioconductor Dependencies)

gwaygenomics:
noted in https://github.com/bioconda/bioconda-recipes/pull/15189#issuecomment-490566204 all CRAN packages should be deposited in conda-forge. This will save time for future contributors.

### DIFF
--- a/docs/source/guidelines.rst
+++ b/docs/source/guidelines.rst
@@ -227,49 +227,49 @@ R (CRAN)
    Bioconda. If the CRAN package does not have a Bioconductor package dependency, it
    belongs in Conda-Forge.
 
- .. note::
+.. note::
 
-     Using the ``conda skeleton cran`` method results in a recipe intended to be
-     built for Windows as well, with lines like::
+   Using the ``conda skeleton cran`` method results in a recipe intended to be
+   built for Windows as well, with lines like::
 
-          {% set posix = 'm2-' if win else '' %}
-          {% set native = 'm2w64-' if win else '' %}
+        {% set posix = 'm2-' if win else '' %}
+        {% set native = 'm2w64-' if win else '' %}
 
-     and
+   and
 
-     .. code-block:: yaml
+   .. code-block:: yaml
 
-         test:
-           commands:
-             - $R -e "library('RNeXML')"  # [not win]
-             - "\"%R%\" -e \"library('RNeXML')\""  # [win]
+       test:
+         commands:
+           - $R -e "library('RNeXML')"  # [not win]
+           - "\"%R%\" -e \"library('RNeXML')\""  # [win]
 
-     The bioconda channel does not build for Windows. To keep recipes
-     streamlined, please remove the "set posix" and "set native" lines described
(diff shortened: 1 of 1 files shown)


gwaygenomics:
cc @bgruening

epruesse:
I still maintain that *all packages* goes too far. There are legit life science packages on CRAN, CPAN and PyPi. Whether you host on Github or on PyPI or on CRAN does not change the primary criterion - life science specificity.

epruesse:
Otherwise we might as well shut this down and move everything to Conda-Forge. Because what applies to CRAN should apply to PyPi as well, and to C/C++ packages hosted on SF or GH or ...

epruesse:
@bioconda/core - this one didn't want to build. "Unauthorized User", see https://circleci.com/gh/bioconda/workflows/bioconda-utils/tree/pull%2F531

gwaygenomics:
> There are legit life science packages on CRAN, CPAN and PyPi. Whether you host on Github or on PyPI or on CRAN does not change the primary criterion - life science specificity.

Definitely agree. My goal with this PR is to reduce the amount of time for contributors to contribute a recipe. It is not an insignificant amount of time to get a package on conda. This is especially true if the contributor has to file two PRs! (e.g. bioconda/bioconda-recipes#15189 and conda-forge/staged-recipes#8309)

Often "life science specificity" is a hazy and subjective boundary. For R users, this boundary can be divided without subjectivity (bioconductor vs. CRAN)

epruesse:
I'll leave it to the R experts. We just had a package from CRAN I merged because it was so unequivocally life-science specific (sequencing stuff) that there was really no good argument to push it to CF.

dpryan79:
As I recall, the main reason to point people to conda-forge was that inevitably stuff from CRAN ends up over there anyway, even if it is bio-related. Perhaps this is less of an issue now that staged-recipes is checking for package existence in bioconda and we have a bot that can check for upgrades. Thought @bgruening ? I guess I'm not opposed to allowing (relevant, obviously) CRAN packages given these changes.

bgruening:
I think we should be clear in our communication. Most cran packages are on conda-forge and they have some good people maintaining them. Sooner or later all cran will end up there and they need to duplicate those, as they do not depend on bioconda.

So I think the cleanest solution is to communicate all pure cran-packages, that means no-dependencies to bioconductor should move to conda-forge.

@gwaygenomics could you please add this to your PR. If a package depends on a bioconductor package it needs to be in bioconda, even it is a cran package.

epruesse:
CF wants to package all of CRAN? That's ... like 14k packages? They cracked the 10k 2 years ago and are in an exponential growth phase. I guess it's less than CPAN (40k) or PyPi (180k) though.

gwaygenomics:
> If a package depends on a bioconductor package it needs to be in bioconda, even it is a cran package.

Yes - I agree with this! I will update the PR. Thanks @bgruening

PR #548, opened 2019-05-20
Rewrite Linting

### LINKED ISSUE #112: Lint: handle malformed yaml
xref: https://github.com/bioconda/bioconda-recipes/issues/4723

### LINKED ISSUE #280: Fail with lint if blacklisted recipe is tried to be updated.
If a recipe is blacklisted and somebody updates it, all tests silently pass because the recipe is ignored. It would be better if linting failed in such a case, with an instruction to remove the recipe from the blacklist.
@mbargull please put this on your todos for the next weeks.

### LINKED ISSUE #367: Lint: require license_file if license==gpl
All versions of GPL and LGPL require that a copy of the license be distributed with the source and binaries.

Are there other license families that require this?

### LINKED ISSUE #414: Linter: check for duplicate keys in yaml
A few of the recipes have had multiple `build:` sections and this breaks `api.render()`.

### LINKED ISSUE #415: Linter: Linting should fail if no build number
There was at least one recipe that had:

    build:
        number:

That broke `api.render()`.

### LINKED ISSUE #515: Lint: recipe folder name and top level package name must be equal
It's just convention, but we expect it and we should enforce it for clarity. If necessary, can always be overridden.

### LINKED ISSUE #540: Lint: allow warnings
In addition to "error/fail", we need at least a "warn" category. Perhaps "info" as well.

epruesse:
This is another bigger one, refactoring the entire linting setup.

New features: check severities, check dependencies

Details:
- Checks are now classes, not functions
- Docstrings serve as message to the user
- Severity can be set as class property
- Several "check_xyz" methods in check are visited with prepared data:
   - `check_recipe` get's the entire recipe
   - `check_deps` get a dependency-to-section-list mapping
   - `check_source` gets the, or each, source
- Messages to user are issued via `self.message()`, which constructs appropriate
   object from test and passed data (e.g. line number, file name, etc).
- A `requires` property allows checks to depend on one another. Excluding a check will
  exclude all checks depending on it. Failing a check will skip all tests depending on it.
- Registration is handled by a metaclass
- No more calls to `conda-build.api.render`, so a few minutes to lint all of Bioconda

In terms of tests, this:
- Checks only the build number in `build_number_needs_bump`, not whether the 
   recipe would build anything new because of a build_string change. The latter was
   the **only** reason to do the slow render of the recipe. We're good without.
- Otherwise reproduces all existing tests
- Changes a few names, splits a few tests

This is how individual checks can be implemented now:

```python
class must_have_about_home(LintCheck):
    """This will be the title of a message presented to user

    And this text goes into the body of a notification
    """
    severity = ERROR  # could also just be a WARNING 
    def check_recipe(self, recipe):
        # get the about/home section and check if it's empty or absent:
        if not recipe.get('about/home', '')
             # issue a message to the user, using class docstring and the
             # line number for the `about` section
             self.message(section='about')
    # we assumed `about` exists, let require that that check has passed
    requires = [must_have_about]

class no_setuptools_in_run:
    def check_deps(self, deps):  # check pre-parsed depends
        # deps is a dict like e.g. 
        # { 'python': ['requirements/host', 'test/requirements', 'outputs/0/run'] }
        # indicating where a dependency is found in the recipe
        for location in deps.get('setuptools', [])
            if 'run' in location:
                self.message(section=location)
       
class no_fn_in_soure:
    def check_source(self, source, section):  # check source section(s)
        if 'fn' in source:
            self.message(section=section)
```

# TODO

- [x] fix severities in build (don't fail if no ERROR checks failed)
- [x] fix severities in cli
- [x] fix severities in bot
- [x] update documentation page
- [x] show new messages in github check annotations

# FIXES

- fixes #540 (allow warnings)
- fixes #515 (check package name = recipe folder name)
- fixes #415 (check no build number)
- [x] fixes #414 (duplicate keys)
- fixes #367 (require license file if gpl)
- fixes #280 (fail in lint if blacklisted)
- [x] fixes #112 (handle malformed yaml)


epruesse:
@dpryan79 - Just look at the `checks.py` file. That's the one where you'd be putting new checks.

dpryan79:
OK, that part I went through the most :) Can you document at the top that the returned message is split between a title and body, which are pulled from the docstring of each check function? In general I think it'd be good to give a brief synopsis of the documentation from `lint/__init__.py` in `lint/checks.py`. Almost anyone who will actually want to add/modify tests will be looking primarily in the latter.

epruesse:
I put a note to look in __init__.py into checks.py. A reason to have the whole submodule is to allow more files containing checks, to have a bit more structure should that list continue to grow.

bioconda_utils/lint/checks.py
```diff
@@ -0,0 +1,552 @@
...
+
+    def check_recipe(self, recipe):
+        if 'setuptools' in recipe.get_deps('run'):
+            self.message()
+
+
+class has_windows_bat_file(LintCheck):
+    """The recipe directory contains a ``bat`` file.
+
+    Bioconda does not and will never build Windows packages.
```
daler: "and never will" is a bit harsh . . . I'd prefer "does not plan to"

bioconda_utils/lint/__init__.py
```diff
@@ -0,0 +1,446 @@
+"""Recipe Linter
+
```
daler: Very helpful. Reading this before even looking at the code gave me a lot of insight and I felt comfortable that I could at least figure things out. Then when reading the code everything made a lot of sense.

bioconda_utils/lint/__init__.py
```diff
@@ -0,0 +1,446 @@
...
+  The ``check_sources`` is called for each source listed in the
+  recipe, whether ``source:`` is a dict or a source, eliminating the
+  need to repeat handling of these two cases in each check. The value
+  ``source`` passed is a dictionary of the source, the value
+  ``section`` a path to the source section.
+
+- When a check encounters an issue, it should use ``self.message()`` to
+  record the issue. The message is commonly modified with a path in
+  the meta data to point to the part of the ``meta.yaml`` that lead to
+  the error. This will be used to position annotations on github.
```
daler: ...and if no message is created, the check is assumed to have passed.

bioconda_utils/lint/checks.py
```diff
@@ -0,0 +1,552 @@
...
+    def __init__(self, linter):
+        super().__init__(linter)
+        self.blacklist = linter.get_blacklist()
+
+    def check_recipe(self, recipe):
+        if recipe.name in self.blacklist:
+            self.message(section='package/name')
+
+
+class folder_and_package_name_must_match(LintCheck):
```
daler: Is this a new one? Good idea.

bioconda_utils/lint/checks.py
```diff
@@ -0,0 +1,552 @@
...
+    in and the name of the toplevel package should match.
+    """
+    def check_recipe(self, recipe):
+        recipe_base_folder, _, _ = recipe.reldir.partition('/')
+        if recipe.name !=  recipe_base_folder:
+            self.message(section='package/name')
+        print(recipe.name, recipe.reldir, recipe_base_folder)
+
+
+class gpl_requires_license_distributed(LintCheck):
```
daler: Also good idea. Maybe convert to WARNING?

bioconda_utils/lint/checks.py
```diff
@@ -0,0 +1,552 @@
...
+    """
+    def check_deps(self, deps):
+        for dep in deps:
+            if dep.startswith('compiler_'):
+                for location in deps[dep]:
+                    if 'run' in location or 'host' in location:
+                        self.message(section=location)
+
+
+class recipe_is_blacklisted(LintCheck):
```
daler: I don't know if this is new or if it's been around and I missed it, either way, super useful.

bioconda_utils/lint/__init__.py
```diff
@@ -0,0 +1,446 @@
...
+        self.recipe: Recipe = None
+
+    def __str__(self):
+        return self.__class__.__name__
+
+    def run(self, recipe: Recipe) -> List[LintMessage]:
+        """Run the check on a recipe. Called by Linter"""
+        self.messages: List[LintMessage] = []
+        self.recipe: Recipe = recipe
+
```
daler: It's not clear what happens to the whole linting system when a check raises an exception. The current checks are pretty simple (hopefully that means robust), but it would be good to gracefully handle exceptions. Maybe wrap each `self.check_*()` with try/except and if exception then a message gets added saying the process of checking failed?

daler:
Oh and the annotations is really nice. I guess we'd need a mirror copy of GitHub to test that though? Or do you have an example PR where it's in use?

bioconda_utils/lint/checks.py
epruesse: It was just my honest estimation. It's just too complicated. Editing as requested, though, wasn't meant to sound harsh.

bioconda_utils/lint/__init__.py
epruesse: done

bioconda_utils/lint/checks.py
epruesse: Thanks, yes. Had a ticket, and it was easy to do now.

bioconda_utils/lint/checks.py
epruesse: I had been pondering. It's easy enough to disable lints, that's why I ended up at ERROR. Your choice.

bioconda_utils/lint/checks.py
epruesse: It's sort of new. The old way, the linter was not considering blacklisted recipes, just issuing warnings to the command line. Now it's a lint itself, so I can make a fail from it.

epruesse:
https://github.com/epruesse/bioconda-recipes/pull/12

epruesse:
(Don't mind that there are a few copies of the annotations on that PR's latest check_run. I can only add, not replace, annotations, so every time I try they add up. The bottom one is the relevant one).

bioconda_utils/lint/checks.py
epruesse: Fixed in e7a10f9

bioconda_utils/lint/__init__.py
epruesse: Great! :)

bioconda_utils/lint/__init__.py
epruesse: Fixed in dc5c91b

bioconda_utils/lint/checks.py
epruesse: changed to warning in 3b04808

bioconda_utils/lint/__init__.py
epruesse: Also done.

epruesse:
I will merge this once the tests pass and update the bot. Since we run dual lint checks anyway at the moment, we can just have the new linter run in the bot and the old on the CI.

If they agree, we can then tag and release an updated version to use in the CI as well.

There are likely some kinks, even though I tried to reproduce everything the old code did, I may have missed corner cases.

PR #574, opened 2019-06-14
Conda skeleton pypi clarification

sndrtj:
This PR adds a little note as to which python packages can be automatically created using `conda skeleton pypi`. 

I got bit by this myself ;-).


epruesse:
Thanks @sndrtj!

PR #582, opened 2019-07-04
Dockerfile: set auto_update_conda=False to allow conda=4.7.5 to downgrade itself

mbargull:
cf. https://gitter.im/bioconda/Lobby?at=5d1bb2d06933ed1a622a334a ff.

> That `Collecting package metadata (repodata.json)` looks like a change in `conda >=4.7.0` https://github.com/conda/conda/commit/1f2a1864199f35aefaade1ed1b8fe6333a73c8b5#diff-127bca0abbd39d90577af817c26af378R234 . But aren't we get conda `4.6.12` here, c.f., https://github.com/bioconda/bioconda-utils/blob/5ca450f4bc9462ca2472389aac000d96846e773b/bioconda_utils/bioconda_utils-requirements.txt#L3 ??

> Oh dear...
> ```
> >  docker pull bioconda/bioconda-utils-build-env &&  docker run --rm -it bioconda/bioconda-utils-build-env conda --version
> Using default tag: latest
> latest: Pulling from bioconda/bioconda-utils-build-env
> Digest: sha256:1bc2f91ac938dfb50283b77512ece61d4faaa9ef85a809c53095097e7c2f3b0a
> Status: Image is up to date for bioconda/bioconda-utils-build-env:latest
> conda 4.7.5
> ```
> https://circleci.com/gh/bioconda/bioconda-utils/5171?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link :
> ```
> ...
> conda install --file /tmp/repo/bioconda_utils/bioconda_utils-requirements.txt &&     conda clean -y --all
> ...
>   added / updated specs:
> ...
>     - conda==4.6.12
> ...
> The following packages will be DOWNGRADED:
> ...
>   conda                                        4.7.5-py37_0 --> 4.7.5-py36_0
> ...
> ```
> https://github.com/conda/conda/pull/8853
> https://github.com/conda/conda/issues/8824
> 
> > After updating conda to 4.7.5, conda can no longer be downgraded to a previous version using a version spec such as `conda install conda=4.6`. This behavior can be changed if the `auto_update_conda` configuration parameter is set to false.
> 
> So we could try setting `auto_update_conda` then.

PR #583, opened 2019-07-04
Add .dockerignore

mbargull:
Same as gh-582 with added `.dockerignore`. (I didn't add this to gh-582 since this is less important to include.)

We don't have to include `.git/` in the container image and even less so the local Miniconda installation created by `.circleci/setup.sh`.

If this does not exclude enough, we could also create a cleaner build context via:
```sh
mkdir ./docker-build-context
git archive HEAD | tar -xC ./docker-build-context
docker build -t bioconda-utils-build-env:latest ./docker-build-context
rm -rf ./docker-build-context
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+miniconda


mbargull:
CI shows this works as intended:
```
93268cd8268b        2 seconds ago      /bin/sh -c #(nop) COPY dir:2a8fce3e91b1ec7...   4.71MB
```
instead of
```
8bef5e983fd6        20 seconds ago      /bin/sh -c #(nop) COPY dir:2650bd09ba59d2a...   668MB
```
A couple hundreds useless MB gone :tada:

mbargull:
Oh, NB: This does not affect the images on Docker Hub, of course. This is just so that the images during testing more closely resemble those on Docker Hub. That way it is easier to asses where an image got bloated by looking at the CI's `docker history` output.

PR #587, opened 2019-07-13
Update conda-forge-pinning

dpryan79:
This will begin the R 3.6 migration


bgruening:
Please note this will now also build against 3.5. I guess we need to overwrite it for Bioconductor?

dpryan79:
Oh yes, thanks for the reminder.

PR #590, opened 2019-07-15
Update conda and conda-build

dpryan79:
These are the newest versions. I'm curious if this will pass testing even.

### DIFF
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -1,7 +1,7 @@
 # basics
 python>=3.6
-conda=4.6.12
-conda-build=3.17.8
+conda=4.7.5
+conda-build=3.18.7
 conda-verify=3.1.*
 argh=0.26.*          # CLI
 colorlog=3.1.*       # Logging
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -1,6 +1,6 @@
 # basics
 python>=3.6
-conda=4.7.5
+conda=4.6.14
 conda-build=3.18.7
 conda-verify=3.1.*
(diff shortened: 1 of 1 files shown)


mbargull:
Fails due to the same `noarch: python`-related bug we worked around in gh-582. This is fixed (supposedly, I didn't check) in `conda >=4.7.6`. Apparently there is `4.7.7` already, not yet on the `conda-forge` channel, though: https://github.com/conda-forge/conda-feedstock/pull/100 .

dpryan79:
4.7.5 is also the most recent on the anaconda channel :(

mbargull:
Ah, okay, `conda >=4.7.6` is still on `conda-canary`, i.e., not blessed as production ready:
https://anaconda.org/anaconda/conda/files
https://anaconda.org/anaconda/conda-canary/files
Hence, we shouldn't use those.

And it looks like like the old packages from the `defaults` channel, i.e. `pkgs/free`, are "effectively removed" per default (can be overridden by `restore_free_channel=True`): https://github.com/conda/conda/blame/4.7.7/CHANGELOG.md#L219
IDK, whether we have packages/recipes that depend on some of those of old packages. Ideally we shouldn't, of course. The old outdated `gcc` package is on `pkgs/free`, for example.

Let's just update `conda-build` and wait on a stable `conda`.

dpryan79:
Anyone know why conda-forge had to mark their builds with conda-build 3.18 as broken?

mbargull:
Not completely sure.
I know I had to work some file access issues on Windows/Azure in https://github.com/conda-forge/r-base-feedstock/pull/84/commits/e59a2427bb55c98a0244ff993c0b8733ba23701a .
Scrolling through the `conda-forge/core` Gitter, it looks like https://github.com/conda/conda-build/pull/3590 and https://github.com/conda/conda-build/pull/3596 were some of the changes to make `3.18.7` work.

Issue #325, opened 2018-08-08
Make `bioconda_utils-conda_build_config.yaml` an `exclusive-config-file`


mbargull:
We need to take advantage of https://github.com/conda/conda-build/pull/3022 and specify our `bioconda_utils-conda_build_config.yaml` also as an `exclusive-config-file`. Reason:
> The use case is Bioconda which is essentially a conda-forge downstream repository now. As such we want to use conda-forge-pinning's conda_build_config.yaml but also add our own on top of that. Using --variant-files won't work, due to the config file processing order, as soon as we would start to add per-recipe configuration files.

epruesse:
@mbargull What's the difference between the config files? Will you be able to fix this?

mbargull:
Sorry, forgot about this one. The fix is trivial and in gh-591.
The difference between the "variant" and "exclusive" config files is their precedence.
We have:
variant config files
*take precedence over*
`recipe/conda_build_config.yaml` which
*takes precedence over*
exclusive config files.
Meaning, you can't override the variant config files, i.e., our `bioconda_utils-conda_build_config.yaml`, for a single recipe, e.g., won't be able to prevent building against `htslib: 1.9` (via the recipe's config file -- in `meta.yaml` still possible, of course).

PR #591, opened 2019-07-16
conda_build_config: clear CONDA_BUILD_SYSROOT and use exclusive_config_files

mbargull:
Fixes https://github.com/bioconda/bioconda-recipes/pull/16402#issuecomment-511840128

```
    # conda-build >=3.18.0 sets CONDA_BUILD_SYSROOT to a hard-coded default path.
    # We clear its value in our bioconda_utils-conda_build_config.yaml.
    # With CONDA_BUILD_SYSROOT being empty, the activation script of clang_osx-64
    # can set it to a valid path.
```

Also fixes https://github.com/bioconda/bioconda-utils/issues/325 .

### DIFF
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -2,6 +2,10 @@
 # in
 # https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml
 
+# clear hard-coded default value for CONDA_BUILD_SYSROOT
+CONDA_BUILD_SYSROOT:
+  - ''
+
 pin_run_as_build:
   htslib:
     max_pin: x.x
--- a/bioconda_utils/utils.py
+++ b/bioconda_utils/utils.py
@@ -456,17 +456,14 @@ def load_conda_build_config(platform=None, trim_skip=True):
     # get environment root
     env_root = PurePath(shutil.which("bioconda-utils")).parents[1]
     # set path to pinnings from conda forge package
-    config.exclusive_config_file = os.path.join(env_root,
-                                                "conda_build_config.yaml")
-    config.variant_config_files = [
+    config.exclusive_config_files = [
+        os.path.join(env_root, "conda_build_config.yaml"),
         os.path.join(
             os.path.dirname(__file__),
-            'bioconda_utils-conda_build_config.yaml')
+            'bioconda_utils-conda_build_config.yaml'),
     ]
-    for cfg in config.variant_config_files:
+    for cfg in chain(config.exclusive_config_files, config.variant_config_files or []):
         assert os.path.exists(cfg), ('error: {0} does not exist'.format(cfg))
-    assert os.path.exists(config.exclusive_config_file), (
-        "error: conda_build_config.yaml not found in environment root")
     if platform:
         config.platform = platform
     config.trim_skip = trim_skip
@@ -483,7 +480,7 @@ def get_conda_build_config_files(config=None):
     if config is None:
         config = load_conda_build_config()
     # TODO: open PR upstream for conda-build to support multiple exclusive_config_files
-    for file_path in ([config.exclusive_config_file] if config.exclusive_config_file else []):
+    for file_path in (config.exclusive_config_files or []):
         yield CondaBuildConfigFile('-e', file_path)
     for file_path in (config.variant_config_files or []):
         yield CondaBuildConfigFile('-m', file_path)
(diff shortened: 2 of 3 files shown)


mbargull:
BTW: `conda-forge-ci-setup` has this:
https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/69c04af2ba90cce0165d6874eb0e197b3cda104e/recipe/run_conda_forge_build_setup_osx#L18-L35
```sh
export CONDA_BUILD_SYSROOT="$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk"

if [[ ! -d ${CONDA_BUILD_SYSROOT} || "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]]; then
    echo "Downloading ${MACOSX_DEPLOYMENT_TARGET} sdk"
    curl -L -O https://github.com/phracker/MacOSX-SDKs/releases/download/10.13/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk.tar.xz
    tar -xf MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk.tar.xz -C "$(dirname "$CONDA_BUILD_SYSROOT")"
    # set minimum sdk version to our target
    plutil -replace MinimumSDKVersion -string ${MACOSX_DEPLOYMENT_TARGET} $(xcode-select -p)/Platforms/MacOSX.platform/Info.plist
    plutil -replace DTSDKName -string macosx${MACOSX_DEPLOYMENT_TARGET}internal $(xcode-select -p)/Platforms/MacOSX.platform/Info.plist
fi

if [ -d "${CONDA_BUILD_SYSROOT}" ]
then
    echo "Found CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
else
    echo "Missing CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
    exit 1
fi
```
Which means conda-forge was not hit by the `conda-build` change from https://github.com/conda/conda-build/commit/8a326413e1f7fc2441a9668027977ecb2e2fdfe0#diff-593f4997b372b8d4a4fe630de1e497adR36 as `conda-forge-ci-setup` always sets `CONDA_BUILD_SYSROOT` to a valid path beforehand, overriding the default value.

Should we do the same in https://github.com/bioconda/bioconda-recipes/blob/master/.circleci/setup.sh ?

bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -2,6 +2,10 @@
 # in
 # https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml
 
+# clear hard-coded default value for CONDA_BUILD_SYSROOT
+CONDA_BUILD_SYSROOT:
+  - ''
```
mbargull: Setting this to an empty string works because of https://github.com/AnacondaRecipes/aggregate/blob/afa864187423917054a307ac4d9f04d3d2463fcc/clang/activate-clang.sh#L122 :
```sh
 "CONDA_BUILD_SYSROOT,${CONDA_BUILD_SYSROOT:-$(xcrun --show-sdk-path)}"
```
which sets resets the variable when it is set but empty.

bioconda_utils/utils.py
```diff
@@ -456,17 +456,14 @@ def load_conda_build_config(platform=None, trim_skip=True):
...
         os.path.join(
             os.path.dirname(__file__),
-            'bioconda_utils-conda_build_config.yaml')
+            'bioconda_utils-conda_build_config.yaml'),
     ]
-    for cfg in config.variant_config_files:
+    for cfg in chain(config.exclusive_config_files, config.variant_config_files or []):
         assert os.path.exists(cfg), ('error: {0} does not exist'.format(cfg))
-    assert os.path.exists(config.exclusive_config_file), (
-        "error: conda_build_config.yaml not found in environment root")
```
mbargull: With this changeset, the more readable `not found in environment root` message is gone, but I think that is fine as the error message shows the path anyway.

mbargull:
Pull request for `conda-build` to remove the default value: https://github.com/conda/conda-build/pull/3606

dpryan79:
I worry that if we just patch around this in .circleci/setup.sh that people will still run into issues running bioconda-utils locally. If the changes to bioconda-utils prevents this in OSX then :+1: from me.

mbargull:
Correct, the changes to `.circleci/setup.sh` in `bioconda-recipes` would be additional/orthogonal to these changes, such that `bioconda-utils` can handle any of those configurations, not just the `bioconda-recipes` one.

bioconda_utils/bioconda_utils-conda_build_config.yaml
epruesse: @mbargull Is that documented somewhere? Or is that undocumented feature? Should we have that noted in the yaml as a comment?

PR #600, opened 2019-08-27
Fix incorrect updating URL

dpryan79:


dpryan79:
It's annoying that the libgfortran-ng from the defaults channel keeps timing out.

epruesse:
Agreed. This can be merged anyway though

PR #605, opened 2019-09-07
Fix #599, at least for vmatch

### LINKED ISSUE #599: duplicate_key_in_meta_yaml doesn't respect selectors
This is understandable given how it's done, but when `yaml.load` hits a duplicate key it throws an error that's passed on as a linter error that can't be skipped. That's often going to be correct, but it doesn't then respect selectors, such as in https://github.com/bioconda/bioconda-recipes/pull/16968. We should really come up with a way around this, as a number of recipes can not longer be "built" because of it. The normal cause of this is that binaries are being redistributed, which while not ideal is sometimes the only option.

dpryan79:
This simply disables the current "convert to list" work-around in the case of a space in a possible selector. In effect this lowers the false rate caused by using comments with a space after the `#`.

### DIFF
--- a/bioconda_utils/recipe.py
+++ b/bioconda_utils/recipe.py
@@ -272,8 +272,11 @@ class Recipe():
             return None
         if not variants:
             return None
-        if any(" " in v for v in variants):
-            # can't handle "[py2k or osx]" style things
+        for k in list(variants.keys()):
+            if " " in k:
+                # can't handle "[py2k or osx]" style things
+                del variants[k]
+        if not variants:
             return None
 
         new_lines = []


johanneskoester:
Ok. I trust you here. 
In the long run I would think that we could maybe preprocess the yaml string with the selector logic from conda, and then yaml.load the selected lines only. This would make the issue go away wouldn't it?

dpryan79:
I'm with you on using something from the conda API instead, it's just a matter of finding the right function and then linting all of the resulting "final" meta.yaml files.

johanneskoester:
This one is it:
https://github.com/conda/conda-build/blob/master/conda_build/metadata.py#L187
In conda-build, it is used [here](https://github.com/conda/conda-build/blob/master/conda_build/metadata.py#L332) to do the selection before YAML is used to parse the remaining content.

For the linting, we should run it in a loop over all possible (combinations of) selector values, which are defined here: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors

johanneskoester:
@epruesse the additional lint you propose is indeed a reasonable workaround for me.

epruesse:
Then we're agreed. Now someone just needs to write that. :)

Since it's pre-parsing, it would probably "just" be a different message for the "duplicatekey" exception.

dpryan79:
That seems like a nicer proposal, closing this

PR #610, opened 2019-09-16
try out some actions, update pinnings and fix some docs

dpryan79:
This implements github actions for most of our CI testing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: dpryan79/github-actions/bioconda_utils_setup_conda/@master
+    - name: Setup
+      uses: dpryan79/github-actions/bioconda_utils_setup_conda/@master
     - name: Install bioconda-utils
       run: |
         python setup.py install
@@ -27,7 +28,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: dpryan79/github-actions/bioconda_utils_setup_conda/@master
+    - name: Setup
+      uses: dpryan79/github-actions/bioconda_utils_setup_conda/@master
     - name: Install bioconda-utils
       run: |
         python setup.py install
@@ -48,7 +50,8 @@ jobs:
     runs-on: macOS-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: dpryan79/github-actions/bioconda_utils_setup_conda/@master
+    - name: Setup
+      uses: dpryan79/github-actions/bioconda_utils_setup_conda/@master
     - name: Install bioconda-utils
       run: |
         python setup.py install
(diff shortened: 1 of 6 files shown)


dpryan79:
@bioconda/core This is a partial proof of concept for using github actions rather than circleci. I was getting some odd conda errors on OSX, so I had to temporarily tweak some things to get around that. It's not immediately apparent, but code reuse in github actions works by writing steps using nodejs in a (typically) separate repository. For the moment you can find that here: https://github.com/dpryan79/github-actions/tree/master/bioconda_utils_setup_conda

If we do end up migrating to github actions then we'll need to have our own repository for stuff like this (the current repository is being shared with the code I use to test deepTools, which I used to try and wrap my head around github actions). We would then need to write some instructions so we don't have to look up how to rebuild the javascript files every time we try to update something.

I haven't implemented the cron job or documentation rebuilding yet.

dpryan79:
The autobump test is now implemented, so it's just the cron jobs and the documentation building that hasn't been migrated. There's no rush on any of that, so I would call this "ready for review".

.github/workflows/test.yml
```diff
@@ -0,0 +1,78 @@
+name: Test
```
bgruening: something more meaningful?

.github/workflows/test.yml
```diff
@@ -0,0 +1,78 @@
...
+on: [push]
+jobs:
+  test-linux:
+    name: Shorter linux tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Setup
+      uses: dpryan79/github-actions/bioconda_utils_setup_conda/@master
```
bgruening: should this be migrated to bioconda?

.github/workflows/test.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: Install bioconda-utils
+      run: |
+        python setup.py install
+    - name: Build docker container
+      run: |
+        docker build -t bioconda-utils-build-env:latest ./
+        docker history bioconda-utils-build-env:latest
+        docker run --rm -t bioconda-utils-build-env:latest sh -lec 'type -t conda && conda info -a && conda list'
+    - name: Run longer tests
```
bgruening: I'm not familiar with pipeline yet, but is there no way to avoid the repetitive definitions here?

.github/workflows/test.yml
dpryan79: Yeah, I should rename the file too.

.github/workflows/test.yml
dpryan79: Yes, if people agree that we should continue on with github actions then I'll create the repos.

.github/workflows/test.yml
dpryan79: Unfortunately not, though at least I got it to about the same length as the circleci tests by off-loading stuff to a different repo. I have to say I prefer the circleci way of writing tests more than how github actions does things, but the latter will likely evolve with time.

dpryan79: I should clarify, we can off-load all of this to a github action repository. We should do that for the docker stuff and we could do it for py.test as well, though since that stuff is so short I figured it wasn't worth the hassle of yet another repo.

dpryan79:
We should update to the very latest conda-forge-pinning before merging this too.

.github/workflows/test.yml
dpryan79: We now have a bioconda-actions repository for things like that.

dpryan79:
This is now using the newly created bioconda-actions repository.

dpryan79:
If we need to migrate things back to circleci we can just uncomment a few lines in it's config.yml.

PR #611, opened 2019-09-18
add debian link

bgruening:


dpryan79:
What would be the real benefit?

epruesse:
Linking packages with those in Debian. E.g. arb-bio (Bioconda) <-> arb (Debian)

dpryan79:
> Linking packages with those in Debian. E.g. arb-bio (Bioconda) <-> arb (Debian)

Do we actually care about that?

bgruening:
> Do we actually care about that?

Debian cares and the some people in ELIXIR. I think it's a good thing and we try to strengthen our collaboration with Debian and others. More updates in the next meeting :)

Also see here: https://blends.debian.org/med/tasks/bio (search bioconda)

PR #613, opened 2019-09-24
Add --keep-old-work option to keep work directory

nathanweeks:
Currently, when `bioconda-utils build` does not complete successfully, the work directory created in ${CONDA_PREFIX}/conda-bld is still removed, which can hinder debugging (e.g., any config.log gets removed). 

It would be helpful to have an option like `conda build --keep-old-work`; this PR (I think) roughly implements similar behavior for `bioconda-utils build` (but if there are differences in behavior, it would be OK to rename the option avoid confusion).

### DIFF
--- a/.github/workflows/GithubActionTests.yml
+++ b/.github/workflows/GithubActionTests.yml
@@ -1,5 +1,5 @@
 name: Github Action Tests
-on: [push]
+on: [push, pull_request]
 jobs:
   test-linux:
     name: Shorter linux tests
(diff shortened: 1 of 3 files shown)


nathanweeks:
The Linux tests failing due to this error seems unrelated to my PR:
```
ERROR conda.core.link:_execute(568): An error occurred while installing package 'defaults::libgfortran-ng-7.3.0-hdf63c60_0'.
```
Could someone force the tests to rerun?

dpryan79:
That's an issue affecting all PRs at the moment. Oddly, it only happens on circleci (the testing works using github actions).

PR #615, opened 2019-10-01
Parallel builds

dpryan79:
This will eventually enable having more than one `bioconda-utils build` instance running at the same time. The basic idea is to partition the DAG on the root nodes into non-overlapping chunks. Workers can then process every Nth partition. The primary (sole?) conceived use of this is on the bulk branch.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -290,15 +290,17 @@ def build_recipes(recipe_folder: str, config_path: str, recipes: List[str],
         logger.info("Nothing to be done.")
         return True
 
-    logger.info("Building and testing %s recipes in total", len(dag))
-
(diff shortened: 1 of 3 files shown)


dpryan79:
I'm still getting this to work, BTW, so it's a bit of a mess at the moment.

dpryan79:
This now at the very least works locally. Interestingly, there is always one group of ~2000 recipes. I'm not sure if that's an artifact of the naive way in which I'm dividing this up or whether one recipes (maybe htslib?) really is the single root node needed by a third of all recipes.

dpryan79:
@bioconda/core This is ready for review. The changes are only intended for use in `bulk`, where in the future we could use multiple bulk workers at the same time to speed things up.

Issue #588, opened 2019-07-15
Bioconductor skeleton needs updating


dpryan79:
Currently the skeleton doesn't handle licenses or descriptions. I should update that before bioconductor 3.10.

dpryan79:
This also needs to be updated to correctly handle packages using r-rgl. bioconductor-mapkl is a good example of this.

PR #617, opened 2019-10-08
Update the bioconductor skeleton

dpryan79:
This should do the following (mostly layed out in #588 ):

 - [x] Include a `license_file` if possible
 - [x] Recipes finally pass linting
 - [x] Detect as much as possible if packages require X
 - [x] Convert any SystemRequirements to actual conda requirements, if possible

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -682,7 +682,6 @@ def bioconductor_skeleton(
             bioc_version = _bioconductor_skeleton.latest_bioconductor_release_version()
         packages = _bioconductor_skeleton.fetchPackages(bioc_version)
         needs_x = _bioconductor_skeleton.packagesNeedingX(packages)
-        print("need X: {}".format(needs_x))
         problems = []
         for k, v in packages.items():
             try:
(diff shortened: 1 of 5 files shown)


dpryan79:
It would of course be much nicer to just parse the system requirements like we do imports/depends/etc., but it turns out that this is completely unstructured. Given that, since we have to special case quite a few packages, we might as well just do all of them.

dpryan79:
The only failing lints are occasional "long summary" message, which isn't a big deal.

bioconda_utils/cli.py
```diff
@@ -681,13 +681,15 @@ def bioconductor_skeleton(
         if not bioc_version:
             bioc_version = _bioconductor_skeleton.latest_bioconductor_release_version()
         packages = _bioconductor_skeleton.fetchPackages(bioc_version)
+        needs_x = _bioconductor_skeleton.packagesNeedingX(packages)
+        print("need X: {}".format(needs_x))
```
bgruening: remove?

dpryan79: Oops, yes, old debug code

Issue #505, opened 2019-04-18
git not found in local circleci build


skurscheid:
I am seeing the same issues described in #13932 ...

Local testing with circleci (installed via snap) fails. For testing purposed I have setup a clean Ubuntu 16.04 VM, installed git and docker (via snap), forked bioconda-recipes and created a branch in my fork

Error message from circleci build:
```
skurscheid@biocondaDev:~/bioconda-recipes$ circleci build
Docker image digest: sha256:58c0be919018ee5ef1a34509a25fcf0b2c57ea1322ab2167632456d316c7b937
====>> Spin up Environment
Build-agent version 1.0.10572-3ce00c85 (2019-04-15T22:09:28+0000)
Docker Engine Version: 18.06.1-ce
Kernel Version: Linux 27d0087d5c23 4.15.0-1041-azure #45-Ubuntu SMP Fri Mar 15 14:41:00 UTC 2019 x86_64 Linux
Starting container bioconda/bioconda-utils-build-env
  using image bioconda/bioconda-utils-build-env@sha256:359113e96c2c6b85583c04c0b4f0b084019677ba2a1c9d48bc15e467ba2f9964

Using build environment variables
  BASH_ENV=/tmp/.bash_env-localbuild-1555382943
  CI=true
  CIRCLECI=true
  CIRCLE_BRANCH=genrich
  CIRCLE_BUILD_NUM=
  CIRCLE_JOB=build
  CIRCLE_NODE_INDEX=0
  CIRCLE_NODE_TOTAL=1
  CIRCLE_REPOSITORY_URL=git@github.com:skurscheid/bioconda-recipes.git
  CIRCLE_SHA1=46c2ea8a9634dface074a4fbfccaf9dec3b7b334
  CIRCLE_SHELL_ENV=/tmp/.bash_env-localbuild-1555382943
  CIRCLE_WORKING_DIRECTORY=~/project

====>> Checkout code
  #!/bin/bash -eo pipefail
mkdir -p /root/project && cd /tmp/_circleci_local_build_repo && git ls-files -z | xargs -0 tar -c | tar -x -C /root/project && cp -a /tmp/_circleci_local_build_repo/.git /root/project
/bin/bash: git: command not found
tar: Cowardly refusing to create an empty archive
Try `tar --help' or `tar --usage' for more information.
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
Error: Exited with code 2
Step failed
Error: runner failed (exited with 101)
[...]

PR #618, opened 2019-10-11
add conda base environment to path in dockerfile

leej3:
I am encountering issues that are detailed in #505. I believe this change fixes that.

Currently git is not on the path inside the container because Conda's base environment isn't activated when the container runs.


dpryan79:
There's at least one other PR open to fix this, but I like this solution the best. I expect we'll have a new bioconda-utils version released within the next couple weeks, so it'll be good to finally have this issue resolved for everyone.

Issue #598, opened 2019-08-13
should_be_noarch_source is overly sensitive


dpryan79:
The should_be_noarch_source lint doesn't appear to check whether the recipe is actually noarch.

dpryan79:
The same issue crops up with `should_be_noarch_source`, e.g., https://github.com/bioconda/bioconda-recipes/pull/16937

PR #625, opened 2019-11-26
lint: fix: should_not_be_noarch_source triggered for noarch:False

mbargull:
Fixes gh-598.

- [x] TODO: add test

### DIFF
--- a/test/lint_cases.yaml
+++ b/test/lint_cases.yaml
@@ -219,7 +219,7 @@ tests:
         sha256: 123
       - url: https://elsewhere  # [osx]
         sha256: 123
-- name: selector_in_source
+- name: selector_in_source_with_compiler
   remove:
     - source
     - build/noarch
@@ -229,6 +229,7 @@ tests:
         sha256: 123
       - url: https://elsewhere  # [osx]
         sha256: 123
+    requirements: { build: ['{{compiler("c")}}'] }
 - name: noarch_java_with_python_wrapper
   add:
     build: { noarch: generic }
--- a/test/lint_cases.yaml
+++ b/test/lint_cases.yaml
@@ -219,6 +219,16 @@ tests:
         sha256: 123
       - url: https://elsewhere  # [osx]
         sha256: 123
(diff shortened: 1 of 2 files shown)


test/lint_cases.yaml
```diff
@@ -219,6 +219,17 @@ tests:
...
+  remove:
+    - source
+    - build/noarch
+  add:
+    source:
+      - url: https://somewhere  # [linux]
+        sha256: 123
+      - url: https://elsewhere  # [osx]
+        sha256: 123
+    requirements: { build: ['{{compiler("c")}}'] }
```
mbargull: If we have a recipe that doesn't use `{{ compiler(...) }}` but has selectors in `source`, `should_be_noarch_generic` is still triggered. We likely wouldn't want such recipe to be `noarch`, but I'm not too concerned about that since that would be rather uncommon => I'd say let's just fix/handle that if we encounter it :man_shrugging:.

dpryan79: Shouldn't the `sha`s have selectors too? Otherwise you have duplicate keys after rendering.

mbargull: Yep, I just mindlessly copy-pasted that part.

mbargull: -> gh-627

nsoranzo:
@mvdbeek and I are seeing the issue fixed by this PR in https://github.com/bioconda/bioconda-recipes/pull/19114 , is this live yet?

dpryan79:
No, we haven't tagged a release and switched to it yet. We're rather slow at doing that, can you just use the skip-lints for now?

marcelm:
I saw this false positive just now in bioconda/bioconda-recipes#15994. Is it correct this fix hasn’t been deployed, yet?

mbargull:
Yep, sorry for being slow on this. I'll tag a release in a few minutes and update the rest so this gets deployed.

marcelm:
Cool, thanks a lot!

PR #626, opened 2019-11-26
Split off more long running tests

mbargull:
Just to make the tests finish a bit faster.
Not very clever, I know.


mbargull:
Works as expected: Still slow, but now doesn't take over half an hour if run in parallel.
(The selection of required status checks has to be adjusted, obviously.)

PR #629, opened 2019-12-20
build_recipes: enable container tests for conda-dependent recipes

mbargull:
From conda>=4.4 on, conda is not restricted to root/base env anymore.

xref: https://github.com/bioconda/bioconda-recipes/pull/19402#issuecomment-567912780

### DIFF
--- a/bioconda_utils/cran_skeleton.py
+++ b/bioconda_utils/cran_skeleton.py
@@ -26,14 +26,14 @@ INVALID_NAME_MAP = {
 gpl2_short = r"  license_family: GPL2"
 gpl2_long = r"""
   license_family: GPL2
-  license_file: '{{ environ["PREFIX"] }}\/lib\/R\/share\/licenses\/GPL-2'  # [unix]
+  license_file: '{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-2'  # [unix]
   license_file: '{{ environ["PREFIX"] }}\\R\\share\\licenses\\GPL-2'  # [win]
 """.strip('\n')
 
 gpl3_short = r"  license_family: GPL3"
 gpl3_long = r"""
   license_family: GPL3
-  license_file: '{{ environ["PREFIX"] }}\/lib\/R\/share\/licenses\/GPL-3'  # [unix]
+  license_file: '{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3'  # [unix]
   license_file: '{{ environ["PREFIX"] }}\\R\\share\\licenses\\GPL-3'  # [win]
 """.strip('\n')
 
--- a/bioconda_utils/cran_skeleton.py
+++ b/bioconda_utils/cran_skeleton.py
@@ -24,20 +24,19 @@ INVALID_NAME_MAP = {
 # Raw strings needed to support the awkward backslashes needed when adding the
 # command to yaml
 gpl2_short = r"  license_family: GPL2"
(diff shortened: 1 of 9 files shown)


test/test_utils.py
```diff
@@ -317,7 +317,7 @@ def test_conda_as_dep(config_fixture, mulled_test):
         one:
           meta.yaml: |
             package:
-              name: one
```
mbargull: Test failed because some other `one==0.1=0` package is being picked up from the local package cache, i.e., some other test is not cleaned up properly.
Obviously, we should fix that, but it's not integral to this PR and I'm not motivated enough to go on that goose chase.

bioconda_utils/cran_skeleton.py
```diff
@@ -24,20 +24,19 @@
...
-gpl3_long = (
-    "  license_family: GPL3\n  license_file: '{{ environ[\"PREFIX\"] }}"
-    "\/lib\/R\/share\/licenses\/GPL-3'  # [unix]\n  "
-    "license_file: '{{ environ[\"PREFIX\"] }}"
-    "\\\R\\\share\\\licenses\\\GPL-3'  # [win]"
-)
+gpl3_long = r"""
+  license_family: GPL3
+  license_file: '{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3'  # [unix]
+  license_file: '{{ environ["PREFIX"] }}\\R\\share\\licenses\\GPL-3'  # [win]
```
mbargull: These strings are later used in `filter_lines_regex` as the `substitute`s for `re.sub(regex, substitute, line)` which means we need the double `\` because `re.sub` processes backslash-escapes in those strings.

mbargull:
This ended up doing a couple of different things:
- ` .github/workflows/GithubActionTests.yml ` and `.circleci/config.yml` and `Dockerfile.test`:
  - Create a local `bioconda/bioconda-utils-test-env:latest` container image that gets used for the tests.
    Previously that was `bioconda-utils-test-env:latest` missing the `bioconda/` part, which means that local image might've never been used, actually, because the code looks for `bioconda/...`.
  - Add `--log-level=DEBUG` to the `pytest` runs so we get more information in case something does wrong.
- `bioconda_utils/bioconductor_skeleton.py`:
  - Suppress some warnings by using `yaml.safe_load` instead of `yaml.load` and switch from HTTP to HTTPS URLs.
- `bioconda_utils/build.py`:
  - Remove the `has_conda_in_build(recipe)` test because it's not needed anymore (main reason for this PR).
- `bioconda_utils/cran_skeleton.py` and ` bioconda_utils/docker_utils.py`:
  - Fix warnings (escapes of characters that don't need escaping) and cleanup string formatting.
- `test/test_utils.py`:
  - Use the `bioconda/bioconda-utils-test-env:latest` container image mentioned above.
  - Adjust/fix `test_conda_as_dep` for the new behavior.

bioconda_utils/cran_skeleton.py
mbargull: Ah, the docs say https://docs.python.org/3/library/re.html#re.sub:
>  If the pattern isn’t found, string is returned unchanged. repl can be a string or a function; if it is a string, any backslash escapes in it are processed.

So, we could also make this
```diff
-gpl3_long = r"""
+def gpl3_long():
+    return r"""
```
if we would want to avoid that strange subtlety.

mbargull:
Thanks for the review!

PR #636, opened 2020-01-31
Update email for the bot

dpryan79:


dpryan79:
Let's see if the new key even works first, since at the moment it's being skipped.

PR #637, opened 2020-02-20
requirements: update gitpython>=3.0.8,3.0.*

mbargull:
Older `gitpython` builds are broken when a newer `gitdb2` is installed.
See https://github.com/conda-forge/staged-recipes/issues/10874 and the linked issues therein for details.


mbargull:
Failed macOS test is only due to an intermittent HTTP error. Merging anyway...

Issue #643, opened 2020-03-20
Installation for MacOS: use curl instead of wget


dariober:
In the installation instruction of (bio)conda for MacOS (https://bioconda.github.io/user/install.html) I suggest to replace:

```
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
```

with:

```
curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
```

Since MacOS by default doesn't ship with `wget`. (It's a bit offputting for a new user to face a problem of installation for a system that is supposed to make installations easy!)

epruesse:
Just `conda install wget`! ;-)

Seriously, good point. Can you click on the "edit on github" link on the page and make a PR? I'll merge ASAP.

epruesse:
Fixed by #644 - thanks @dariober

PR #644, opened 2020-03-20
 Installation for MacOS: use curl instead of wget

dariober:
See issue #643 - For consistency, change also the instructions for Linux.


epruesse:
Thanks @dariober!

PR #650, opened 2020-05-03
Update bioconductor skeleton and pinnings

dpryan79:
The skeleton updates should add blas and lapack as host dependencies for all compiled packages. This should prevent a repeat of the last "can't link to libopenblasp" errors that kept popping up on OSX.


dpryan79:
Even if the tests pass this will need to be discussed before merging. At the moment it adds support for python 3.8 and drops support for 2.7 (since conda-forge has dropped it).

PR #652, opened 2020-05-08
Gah, need to actually ADD the new requirements...

dpryan79:
I really should have tested this in my last PR...


dpryan79:
I'm going to test build BiocGenerics and Biobase locally to double check this (the latter is affected by this) and generate all of the recipes locally before merging.

PR #658, opened 2020-06-04
docs: rename recipe index to package index + minor layout fixes

mbargull:
I'll post screenshots for comparison in a bit (building docs locally right now..).

### DIFF
--- a/.circleci/build-docs.sh
+++ b/.circleci/build-docs.sh
@@ -58,6 +58,9 @@ rm -r *
 
 # build docs and copy over to tmpdir
 cd ${DOCSOURCE}
+# NOTE: With more build jobs (e.g., "-j8") the "Required By:" entries in the
+#       package index do not work. DO NOT change "-j1" unless that gets fixed!
+#       ref: https://github.com/bioconda/bioconda-utils/pull/658#issuecomment-639399930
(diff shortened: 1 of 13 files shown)


mbargull:
before:
image
image
image

mbargull:
after:
image
image
image
image

dpryan79:
Nice!

mbargull:
Oh, I just noticed that the `Required By:` doesn't work anymore... Maybe I have to do something different than just adding `:noindex:` to the recipes.

mbargull:
> Oh, I just noticed that the `Required By:` doesn't work anymore... 

Nvm. I built the docs locally with `-j8` and that didn't yield the `Required By:` entries. But with `-j1` it works. I'll just add a note in `build-docs.sh`. 

----
I also added a commit to put versions/constraints in `<code>` blocks: IMHO, that makes things slightly more readable and it also fixed reST issues with dependencies like `python_abi 3.6.* *cp36m` that have asterisks in them (that commit also does `.replace(" ", "\xa0")` for those entries which might also fixed things -- still, code blocks make them more readable, anyway).

before:
image
after:
image

image

mbargull:
I'll merge once https://github.com/bioconda/bioconda-recipes/pull/22526 is in to check if the newly built package index also has the `pmid`/`rrid` identifiers linked properly (works locally, but just wanting to be sure).

mbargull:
Thanks for the reviews!

PR #662, opened 2020-06-18
requirements: update to conda=4.8.3 + conda-build issue 3974 workaround

mbargull:
This updates our requirement to `conda=4.8.3`... and removes `current_repodata.json` from the config where possible.

We also have to adjust https://github.com/bioconda/bioconda-recipes/blob/8b019e7be9dbc87b3419f93b5a2afcf6a1411466/.circleci/setup.sh#L77 then. (And things for `bulk`.)

### DIFF
--- a/MulledDockerfile
+++ b/MulledDockerfile
@@ -18,11 +18,6 @@ RUN apt-get update -qq --fix-missing && \
 RUN curl -sL https://repo.anaconda.com/miniconda/Miniconda3-py37_4.8.3-Linux-x86_64.sh -o ~/miniconda.sh && \
     /bin/bash ~/miniconda.sh -b -p /opt/conda && \
     rm ~/miniconda.sh && \
-    ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
-    echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && \
-    echo "conda activate base" >> ~/.bashrc && \
-    { conda config --remove repodata_fns current_repodata.json || true ; } && \
-    conda config --prepend repodata_fns repodata.json && \
     CONDA_AGGRESSIVE_UPDATE_PACKAGES= CONDA_AUTO_UPDATE_CONDA=0 \
         conda install -y conda-forge::tini && \
     conda clean -ya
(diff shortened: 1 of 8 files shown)


mbargull:
Also adds a hack/workaround for https://github.com/conda/conda-build/issues/3974.

MulledDockerfile
```diff
@@ -1,27 +1,37 @@
-FROM debian:latest
+FROM debian:10-slim as base
 
 #  $ docker build -f MulledDockerfile -t bioconda/mulled_container:latest
```
mbargull, line 3: @dpryan79, could you build and upload this to `quay.io/dpryan79/mulled_container:latest`?
In the next few days I'll create another repo that'll hold this and uploads to `bioconda` on Docker Hub, but in the meantime we can just update yours :).

dpryan79: Sure, I'll try to get that done today

dpryan79: Not sure if this is an issue, since `conda env list` still works:

```
$ docker run --rm -it dpryan79/mulled_container:latest /bin/bash

CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
To initialize your shell, run

    $ conda init <SHELL_NAME>

Currently supported shells are:
  - bash
  - fish
  - tcsh
  - xonsh
  - zsh
  - powershell

See 'conda init --help' for more information and options.

IMPORTANT: You may need to close and restart your shell after running 'conda init'.
```

dpryan79: Weird, adding `conda init bash` doesn't change it, which surprises me a bit.

dpryan79: I'll just push this then.

mbargull: Ah, balls. Hold on, I'll fix it right now.

dpryan79: It's live

mbargull: Shouldn't be much of an issue, but can you rebuild from https://github.com/bioconda/bioconda-utils/pull/662/commits/8e6c9aa56684d54955d597f9f07683d66e31ac0a ?
Sorry for not catching this before!

mbargull: I'll have to reread https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html  yet again, I guess -.-

mbargull: Well, the `/bin/bash -l` in the entrypoint was correct. But there were other very dump oversights I fixed in https://github.com/bioconda/bioconda-utils/pull/662/commits/7ef9e1ac84c09b2042ed83e841e77c9d968e647f .
Can you build for that commit again when you find the time for it? Thanks and sorry again.

PR #664, opened 2020-06-26
Add Docs: Only support python v3.6 and v3.7

LiterallyUniqueLogin:
I added a package to Bioconda, and then flailed for quite a while not understanding why it wouldn't build against python=3.5 or python=3.8. If these are not supported, somewhere in the docs should say so. With this change, that's now true, and there's now a place to list similar information for other important software (maybe R?).

### DIFF
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -1,3 +1,7 @@
+# Some of Bioconda's supported versions are described
+# in ../docs/source/user/versions.rst
+# so please keep that documentation up to date as they change
+
 # Additional bioconda-specific pinnings to use in addition to those specified
 # in
 # https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml
--- a/docs/source/user/versions.rst
+++ b/docs/source/user/versions.rst
@@ -14,6 +14,12 @@ Python
 Bioconda only supports python 2.7, 3.6 and 3.7. The community is currently working on adding support for
 python 3.8.
 
+(The exception to this is Bioconda packages which declare `noarch: python` and only depend on
+such packages - those packages can be installed in an environment with any version of python
+they say they can support.
+However many python packages in Bioconda depend on other Bioconda packages with architecture specific
+builds, such as `pysam`, and so do not meet this criteria.)
+
 Unsupported versions
 --------------------
 If the version of a package you wish to use is not supported because it is too old, please upgrade to a newer
--- a/docs/source/user/versions.rst
+++ b/docs/source/user/versions.rst
@@ -11,7 +11,7 @@ Bioconda only supports 64-bit Linux and Mac OS
 
 Python
 ------
-Bioconda only supports python 3.6 and 3.7. The community is currently working on adding support for
+Bioconda only supports python 2.7, 3.6 and 3.7. The community is currently working on adding support for
 python 3.8.
 
 Unsupported versions
--- a/docs/source/user/versions.rst
+++ b/docs/source/user/versions.rst
@@ -1,0 +1,29 @@
+
+Versions Supported in Bioconda
+==============================
+Bioconda only supports some versions of conda packages.
+Here is an incomplete list of the dependencies that might be relevant to your
+conda environment.
+
+Operating Systems
+-----------------
+Bioconda only supports 64-bit Linux and Mac OS
+
+Python
+------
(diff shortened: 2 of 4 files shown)


LiterallyUniqueLogin:
@mbargull @dpryan79 you two helped me through this, so would you review this contribution?

docs/source/user/versions.rst
```diff
@@ -0,0 +1,29 @@
...
+Here is an incomplete list of the dependencies that might be relevant to your
+conda environment.
+
+Operating Systems
+-----------------
+Bioconda only supports 64-bit Linux and Mac OS
+
+Python
+------
+Bioconda only supports python 3.6 and 3.7. The community is currently working on adding support for
```
dpryan79, line 14: And 2.7. Can you mention that this is only relevant for architecture-specific builds? `noarch: python` packages are unaffected by these restrictions.

LiterallyUniqueLogin: Good catch r.e. 2.7. But [my recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/trtools/meta.yaml) is `noarch: python` but I'm still having these issues. Any idea why? Is it because of some of my dependencies?

dpryan79: That must be due to some dependencies. For `noarch` packages the pinnings (what underlies the python versions we build for) don't play a role.

LiterallyUniqueLogin: I added a note with those details. Is it accurate and reasonable comprehensible?

dpryan79: Yeah, that's a fine note, thanks!

dpryan79:
Not sure what's wrong with the OSX test. That's unrelated to this PR so I'll merge

dpryan79:
Thanks @LiterallyUniqueLogin !

LiterallyUniqueLogin:
Happy to! This is the first time I've contributed to an open source project, it feels good! Thank you for all your help figuring this out.

mbargull:
> This is the first time I've contributed to an open source project, it feels good!

:tada: :congratulations: !
Thanks for this, I'm always grateful when someone helps out with the docs!

docs/source/user/versions.rst
```diff
@@ -0,0 +1,35 @@
...
+Bioconda only supports python 2.7, 3.6 and 3.7. The community is currently working on adding support for
+python 3.8.
+
+(The exception to this is Bioconda packages which declare `noarch: python` and only depend on
+such packages - those packages can be installed in an environment with any version of python
+they say they can support.
+However many python packages in Bioconda depend on other Bioconda packages with architecture specific
+builds, such as `pysam`, and so do not meet this criteria.)
+
+Unsupported versions
```
mbargull, line 23: @LiterallyUniqueLogin, can you help me understand this section?
Why do we need to distinguish between smaller or bigger software? Whether it should be added to Bioconda shouldn't be a matter of size usually but can be decided in the usual contribution process, I think.

I kind of feel we don't need this "Unsupported versions" section -- but I maybe/probably just didn't get the gist of what you were getting at.

bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -1,3 +1,7 @@
+# Some of Bioconda's supported versions are described
```
mbargull, line 1: (Thanks for including those comments! :heart:)

LiterallyUniqueLogin: Happy to! =)

docs/source/user/versions.rst
LiterallyUniqueLogin: I think there should be some link to the contributing page and the gitter so that someone who wants a new version supported has a path forward instead of feeling stuck. Is that all I should have done? I can make that change.

Once we provide a link, I was worried that you might get people who fundamentally misunderstand the amount of work required by and likelihood of such a change which could be frustrating (e.g. people who post a PR to upload a single package with a python 3.5 dependency without understanding the issues with their dependencies, or someone saying "Hey, I'd like Bioconda to support Windows. Can you guys make that happen?"). So I tried to anticipate some of those issues. But maybe instead of trying to anticipate that here we could just let people go to the gitter and get feedback there.

Thoughts?

mbargull: I see, now I understand. Thanks for the explanation!

> Thoughts?

Not yet ^^. I'll have to think about how to phrase that. Probably not something I'll do too too soon, but I've made a note to remind me!

PR #666, opened 2020-06-29
Fixed links to use correct rst syntax

nilshoffmann:


dpryan79:
We are very appreciative of these fixes! Thanks much @nilshoffmann!

PR #667, opened 2020-07-02
update-pinnings fixes

mbargull:
Some fixes for `update-pinnings`.

IMPORTANT: The change from https://github.com/bioconda/bioconda-utils/pull/662/files#diff-2aed786d397cbb062fb58d1ff86a6950R26 caused different build strings for `r-base`-dependent packages because it triggered the logic from https://github.com/conda/conda-build/pull/3895. I hadn't thought of that before unfortunately.
Thankfully the last Bioconductor `bulk` work happened before that change. That means it should not affect terribly many recipes.

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -541,6 +541,9 @@ def dag(recipe_folder, config, packages="*", format='gml', hide_singletons=False
      help="""Bump package build numbers even if the only applicable pinning
      change is the python version. This is generally required unless you plan
      on building everything.""")
+@arg('--skip-variants',
+     nargs='*',
+     help='Skip packages that use one of the given variant keys.')
 @arg('--cache', help='''To speed up debugging, use repodata cached locally in
      the provided filename. If the file does not exist, it will be created the
      first time.''')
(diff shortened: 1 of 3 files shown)


bioconda_utils/cli.py
```diff
@@ -541,6 +541,9 @@ def dag(recipe_folder, config, packages="*", format='gml', hide_singletons=False
      help="""Bump package build numbers even if the only applicable pinning
      change is the python version. This is generally required unless you plan
      on building everything.""")
+@arg('--skip-variants',
```
mbargull, line 544: I added this as a convenience things so we can update pinnings for Python/R packages separately.

mbargull: However, those skips happen after the dependency graph is created, ergo packages that are dependencies of others may not be bumped.

bioconda_utils/cli.py
```diff
@@ -578,11 +583,13 @@ def update_pinning(recipe_folder, config, packages="*",
...
 
-    needs_bump = partial(update_pinnings.check, build_config=build_config)
+    needs_bump = partial(
+        update_pinnings.check, build_config=build_config, skip_variant_keys=skip_variants,
+    )
 
     State = update_pinnings.State
 
-    for status, recip in zip(utils.parallel_iter(needs_bump, dag, "Processing..."), dag):
+    for status, recip in utils.parallel_iter(needs_bump, dag, "Processing..."):
```
mbargull, line 592: The order of `utils.parallel_iter(...)` and `iter(dag)` is different, apparently, which means we got "random" misassigned build bumps.

mbargull:
I'm going ahead with the merge and will tag yet another release so we get the `r_base: [4.0.*]` change live.
If there are any issues with the `update-pinnings`, we can address them later since it is just a "on-demand" command but does not affect daily business.

PR #668, opened 2020-07-04
lint: add trigger for missing_meta_yaml

mbargull:
Fixes issue reported in https://github.com/bioconda/bioconda-recipes/pull/22992, i.e.,
recipe file ending with `.yml` instead of `.yaml` but not being detected by the linter.


dpryan79:
Ah, that was the issue with that PR, I miss that sort of thing 99% of the time :)

PR #672, opened 2020-07-30
Add explanations on how to handle tests with dependencies

LiterallyUniqueLogin:
This is what @dpryan79 explained to me while I was trying to upload the TRTools package. Figure it would be useful to include in the docs for future contributors. I'm no expert on this, so please correct the wording where I'm misrepresenting some detail.

### DIFF
--- a/docs/source/contributor/guidelines.rst
+++ b/docs/source/contributor/guidelines.rst
@@ -155,15 +155,14 @@ Test Dependencies
 ~~~~~~~~~~~~~~~~~
 
 Bioconda's stringent ``mulled-build`` test that runs during a package build
-will run the tests listed in the `tests` section of the recipe's `meta.yaml`
-file in an environment which does not have the dependecies listed in the
-`tests.requires` section installed. Because of this, all the tests in that section
+will run the tests listed in the ``test`` section of the recipe's ``meta.yaml``
+file in an environment which does not have the dependencies listed in the
+``test: requires:`` section installed. Because of this, all tests in that section
 must only rely on runtime dependencies or the build will fail.
-Tests that rely on the testtime dependencies listed in `test.requires` should be put in
-the `run_test.sh` file in the same directory as the `meta.yaml` file. This file is
+Tests that rely on the test-time dependencies listed in ``test: requires:`` should be put in
+the ``run_test.sh`` file in the same directory as the ``meta.yaml`` file. This file is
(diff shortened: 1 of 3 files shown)


docs/source/contributor/guidelines.rst
```diff
@@ -148,6 +148,24 @@ include them in the PR. One shortcut is to use ``anaconda search -t conda
...
+Bioconda's stringent ``mulled-build`` test that runs during a package build
+will run the tests listed in the `tests` section of the recipe's `meta.yaml`
+file in an environment which does not have the dependecies listed in the
+`tests.requires` section installed. Because of this, all the tests in that section
+must only rely on runtime dependencies or the build will fail.
+Tests that rely on the testtime dependencies listed in `test.requires` should be put in
+the `run_test.sh` file in the same directory as the `meta.yaml` file. This file is
+picked up and run by the standard build that occurs before the ``mulled-build``, but is
+not installed with the package and so is not picked up and run by the ``mulled-build``.
+
```
mbargull, line 166:
```suggestion
Bioconda's stringent ``mulled-build`` test that runs during a package build
will run the tests listed in the ``test`` section of the recipe's ``meta.yaml``
file in an environment which does not have the dependencies listed in the
``test: requires:`` section installed. Because of this, all tests in that section
must only rely on runtime dependencies or the build will fail.
Tests that rely on the test-time dependencies listed in ``test: requires:`` should be put in
the ``run_test.sh`` file in the same directory as the ``meta.yaml`` file. This file is
picked up and run by the standard build that occurs before the ``mulled-build``, but is
not passed on to and run by the ``mulled-build``.
```

mbargull:
I hope we will change how these things are handled in the future.
But it also was a big oversight that we didn't note that deficiency in the docs.
Thank you for fixing that!

LiterallyUniqueLogin:
Apologies, I had failed to commit and include the last round of fixes I had on my computer into this PR; I've now done so. It's just a few minor wording changes and fixing a broken cross-section link

docs/source/contributor/guidelines.rst
LiterallyUniqueLogin: Thanks for finding this! I had changed the single backticks to double backticks on my computer alongside a few other changes but failed to get that last commit into the PR, and didn't notice till I saw your fix here. I've added that last commit on top of your fix below.

mbargull:
All good, thanks!

PR #675, opened 2020-07-30
utils.get_recipes: consider exclude in meta_yaml_found handling

mbargull:
follow up fix for gh-668 that let's the `autobump` not process `bioconductor-*` recipes.


mbargull:
Merging right away to stop `autobump` from touching `bioconductor-*`

PR #680, opened 2020-08-30
Update pinning: py3.8, conda-forge-pinning=2020.08.29.21.09.48

mbargull:
Fixes/updates to `bioconda-utils update-pinning` and pinning updates to include Python 3.8 in our build matrix.

### DIFF
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -29,6 +29,28 @@ bamtools:
 r_base:
   - 4.0.*
 python:
-  - 2.7.*  *_cpython
-  - 3.6.*  *_cpython
+  - 3.8.*  *_cpython
   - 3.7.*  *_cpython
+  - 3.6.*  *_cpython
+  - 2.7.*  *_cpython
+
+# conda-forge-pinning>=2020.07.13.05.23.11 uses clang=10.
+# We still pin clang=9 here and will update to clang=10 when we update to gcc=9.
+# (Note that with bioconda-utils update-pinning we don't look at osx changes.
+#  Hence, if we update clang alongside gcc, we get more consistent behavior.)
+c_compiler:
+  - gcc                        # [linux]
+  - clang                      # [osx]
+c_compiler_version:            # [unix]
+  - 9                          # [osx]
+  - 7                          # [linux64 or aarch64]
+cxx_compiler:
+  - gxx                        # [linux]
+  - clangxx                    # [osx]
+cxx_compiler_version:          # [unix]
+  - 9                          # [osx]
+  - 7                          # [linux64 or aarch64]
+fortran_compiler:              # [unix or win64]
+  - gfortran                   # [(linux64 or osx)]
+fortran_compiler_version:      # [unix or win64]
+  - 7                          # [linux64 or osx or aarch64]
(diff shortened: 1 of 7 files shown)


bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -29,6 +29,28 @@ bamtools:
...
   - 4.0.*
 python:
-  - 2.7.*  *_cpython
-  - 3.6.*  *_cpython
+  - 3.8.*  *_cpython
   - 3.7.*  *_cpython
+  - 3.6.*  *_cpython
+  - 2.7.*  *_cpython
+
+# conda-forge-pinning>=2020.07.13.05.23.11 uses clang=10.
```
mbargull, line 37: My initial plan was to just update to `gcc=9` ahead of `conda-forge-pinning` here.
But I got really tired of the rebuilds/recipe fixes lately and am not willing to handle the nearly 2k recipes that would be rebuild by the compiler changes right now.
Postponing that until `conda-forge-pinning` gets updates GCC itself. There's no concrete schedule for that, yet, probably this fall for `gcc=9` (GCC 9 was suggested over GCC 10 due to compatibility concerns the last time this has been discussed on the conda-forge/core meeting).

bioconda_utils/cli.py
```diff
@@ -537,13 +537,13 @@ def dag(recipe_folder, config, packages="*", format='gml', hide_singletons=False
...
-     help="""Bump package build numbers even if the only applicable pinning
-     change is the python version. This is generally required unless you plan
-     on building everything.""")
 @arg('--skip-variants',
      nargs='*',
      help='Skip packages that use one of the given variant keys.')
+@arg('--max-bumps', type=int,
+     help='Maximum number of recipes that will be updated.')
+@arg('--no-leaves',
+     help='Only update recipes with dependent packages.')
```
mbargull, line 546: The idea here is to update packages in certain batches on `bulk`.
With `--no-leaves` we leave (no pun intended!) out packages that are not needed by others and initially only bump the "inner nodes" of the dependency graph.
With `--max-bumps` we can limit the number of recipes that will get bumped. That way we can get issue smaller PRs from `bulk` to `master` and as such synchronize both branches more quickly. This will be particularly helpful in big rebuilds like for the compiler updates.

bioconda_utils/recipe.py
```diff
@@ -206,6 +210,32 @@ def load_from_string(self, data) -> "Recipe":
...
+        # Cache contents of conda_build_config.yaml for conda_render.
+        path = Path(self.dir, 'conda_build_config.yaml')
+        if path.is_file():
+            self.conda_build_config = path.read_text()
+        else:
+            self.conda_build_config = ''
+
+    def read_build_scripts(self):
+        # Cache contents of build scripts for conda_render since conda-build
+        # inspects build scripts for used variant variables.
```
mbargull, line 223: This keeps the build scripts (and `conda_build_config.yaml` if present) in memory, increasing the memory usage of `bioconda-utils update-pinning` even more. I haven't checked by how much, though. `bioconda-utils update-pinning` in general chews through a good amount of memory. Just beware that you should run `bioconda-utils update-pinning` on a machine with more than a couple of gigs of RAM ;).

bioconda_utils/cli.py
```diff
@@ -537,13 +537,13 @@ def dag(recipe_folder, config, packages="*", format='gml', hide_singletons=False
      help="""Skip updating/bumping packges that are already built with
      compatible pinnings in one of the given channels in addition to those
      listed in 'config'.""")
-@arg('--bump-only-python',
```
mbargull, line 540: @dpryan79, I hope the added `noarch` checks are enough to offset the need for this. Let me know if I didn't catch some use cases for this.

PR #692, opened 2021-01-06
Add shellcheck linter

### LINKED ISSUE #691: [Feature request] Add Lint for shellcheck
# Problem

Many recipes have shell scripts (e.g. `build.sh`, `post-link.sh`, `pre-unlink.sh`). Some of those scripts have portability issues, such as:

- do not have a shebang. [example](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/r-empiricalfdr.deseq2/build.sh)
- do not quote directories [example](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/a5-miseq/build.sh) (conda already warns about avoiding spaces, but anyway better safe than sorry)
- Have bashisms even if they are declared as /bin/sh scripts [example](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/porfast/build.sh)
- ...
 
This makes those scripts fragile, or lead to unwanted effects.

For instance, if a user installs a package on a conda environment with spaces, like `/path/with spaces` a post installation script like:

```bash
#!/bin/sh
DIR="/path/with spaces"
# whatever work done here
rm -rf $DIR
```
will remove "/path/with" and "spaces" instead of "/path/with spaces"

More dangerously could be a path with a trailing space:

```bash
#!/bin/sh

#DIR="/path/with-a-trailing-space /"
#rm -rf $DIR
```
Which I have commented out because you do not want to remove `/` by accident.

# Solution: shellcheck

shellcheck is a tool that checks shell scripts and looks for potential and common errors. See more bad-code examples here: https://github.com/koalaman/shellcheck/blob/master/README.md#gallery-of-bad-code

It is widely available (as a debian package, installable through brew on osx and on conda-forge). 

Linting bioconda scripts with shellcheck could prevent many of those issues.

    shellcheck recipes/bioconductor-dapardata/*.sh

imatge

It would be great to try to have an integrated linter in bioconda to detect those issues and suggest corrections.

zeehio:
This PR Closes #691.

I need to test this properly, but I am not very familiar with bioconda-utils. Therefore I will mark this PR as Draft. I'll try to find time to setup a proper test environment (help from others is always welcome :smiley: ).

### DIFF
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -26,6 +26,7 @@ anaconda-client=1.6.*  # anaconda_upload
 involucro=1.1.*        # mulled test and container build
 skopeo=0.1.35          # docker upload
 git=2.14.*             # well - git
+shellcheck=0.7.*       # shellcheck linter
 
 # hosters - special regex not supported by RE
 regex=2018.08.29
--- a/bioconda_utils/lint/__init__.py
+++ b/bioconda_utils/lint/__init__.py
@@ -87,6 +87,7 @@ Module Autodocs
    check_noarch
    check_policy
    check_repo
+   check_shellcheck
    check_syntax
 
 """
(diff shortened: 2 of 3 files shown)


epruesse:
Keep going, I think this would be useful

PR #698, opened 2021-03-13
Update conda-forge-pinnings, conda-build and bioconda-specific compilers

dpryan79:
This adds python 3.9 as well and basically updates everything else. We could try to time this to correspond to the htslib 1.12 release instead, but given the relative urgency of the TBB stuff perhaps waiting is a bad option.

### DIFF
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -44,11 +44,11 @@ python_impl:
   - cpython
   - cpython
 numpy:
-  - 1.19
-  - 1.17
-  - 1.17
-  - 1.17
-  - 1.16
+  - 1.19.*
+  - 1.17.*
+  - 1.17.*
+  - 1.17.*
+  - 1.16.*
 
 channel_sources:
   - conda-forge,bioconda,defaults
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -20,7 +20,9 @@ pin_run_as_build:
     min_pin: x.x
 
 htslib:
-  - 1.10
+  - 1.12
+libdeflate:
+  - 1.7
 bamtools:
   - 2.5.1
 
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -29,25 +29,22 @@ bamtools:
 r_base:
   - 4.0.*
 python:
+  - 3.9.*  *_cpython
   - 3.8.*  *_cpython
   - 3.7.*  *_cpython
   - 3.6.*  *_cpython
   - 2.7.*  *_cpython
-
-c_compiler:
-  - gcc                        # [linux]
-  - clang                      # [osx]
-c_compiler_version:            # [unix]
-  - 9                          # [osx]
-  - 7                          # [linux64 or aarch64]
-cxx_compiler:
-  - gxx                        # [linux]
-  - clangxx                    # [osx]
-cxx_compiler_version:          # [unix]
-  - 9                          # [osx]
-  - 7                          # [linux64 or aarch64]
-fortran_compiler:              # [unix or win64]
-  - gfortran                   # [(linux64 or osx)]
-fortran_compiler_version:      # [unix or win64]
-  - 7                          # [linux64 or osx or aarch64]
+# conda-forge only has 3 python versions, so we need to modify the length of the other zip_keys in its group
+python_impl:
+  - cpython
+  - cpython
+  - cpython
+  - cpython
+  - cpython
+numpy:
+  - 1.16
+  - 1.16
+  - 1.16
+  - 1.16
+  - 1.16
 
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -45,10 +45,10 @@ python_impl:
   - cpython
 numpy:
   - 1.19
-  - 1.16
-  - 1.16
-  - 1.16
-  - 1.16
+  - 1.17
+  - 1.17
+  - 1.17
+  - 1.17
 
 channel_sources:
   - conda-forge,bioconda,defaults
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -42,7 +42,7 @@ python_impl:
   - cpython
   - cpython
 numpy:
-  - 1.16
+  - 1.19
   - 1.16
   - 1.16
   - 1.16
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -29,28 +29,8 @@ bamtools:
 r_base:
   - 4.0.*
 python:
+  - 3.9.*  *_cpython
   - 3.8.*  *_cpython
   - 3.7.*  *_cpython
   - 3.6.*  *_cpython
   - 2.7.*  *_cpython
-
-# conda-forge-pinning>=2020.07.13.05.23.11 uses clang=10.
-# We still pin clang=9 here and will update to clang=10 when we update to gcc=9.
-# (Note that with bioconda-utils update-pinning we don't look at osx changes.
-#  Hence, if we update clang alongside gcc, we get more consistent behavior.)
-c_compiler:
-  - gcc                        # [linux]
-  - clang                      # [osx]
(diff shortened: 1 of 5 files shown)


dpryan79:
I don't know why pandas isn't working, it's fine locally.

dpryan79:
@mbargull Am I correct in assuming that the Linux failure is due to the hashing change?

dpryan79:
Indeed, these are the fields that are added to the hash:

```
INFO :: The inputs making up the hashes for the built packages are as follows:
{
  "one-0.1-py36ha770c72_0": {
    "recipe": {
      "channel_targets": "conda-forge main",
      "target_platform": "linux-64"
    }
  }
}
```

I'll change the channel_targets to `bioconda main`

bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -48,3 +48,8 @@ numpy:
   - 1.16
   - 1.16
 
+channel_sources:
+  - conda-forge,bioconda,defaults
+
+channel_targets:
```
mbargull, line 54: Ouch... Yep, checked conda-build's source... This is now hard-coded to be included... I can't say I'm a big fan of including hashes for something which should be purely output-related :/.
(^-- Just a rant about conda-build's, not your changes.)
(My best guess they included it for their convenience since some recipes on conda-forge/AnacondaRecipes use `channel_targets` in some conditional Jinja control code so they don't have to manually sync the recipes.)

bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -29,28 +29,27 @@ bamtools:
...
+  - cpython
+  - cpython
+  - cpython
+  - cpython
+numpy:
+  - 1.19
+  - 1.16
+  - 1.16
+  - 1.16
+  - 1.16
```
mbargull, line 49: I wonder if we have to do a similar dance as with `python`/`r_base` above regarding https://github.com/conda/conda-build/issues/3974 ..

mbargull:
> We could try to time this to correspond to the htslib 1.12 release instead,

Do we have any insights on for when that release is planned?

> but given the relative urgency of the TBB stuff perhaps waiting is a bad option.

`tbb`/`tbb-devel` only affect few recipes *directly*\*, right? We could also just do a minor pinning change with just the TBB constraint beforehand and do the mass-rebuild afterwards.
(\* of course, indirectly for workflows/other packages, I guess it has much more impact, seeing `bowtie2` et al. being affected.)

bioconda_utils/bioconda_utils-conda_build_config.yaml
mbargull: Hm.. I find things around merged config + zip_keys too confusing (code is not without bugs and I don't understand all intentions).
How about we avoid all of that and do things manually? I.e., not use multiple "exclusive config files" for now, but just copy conda-forge's, comment out what we want to override and pre- or append our pinnings in the same file. Not ideal, but probably much safer...

dpryan79:
@mbargull Should we pin libdeflate here? That's caused some issues on OSX in particular.

mbargull:
> @mbargull Should we pin libdeflate here? That's caused some issues on OSX in particular.

I guess that makes sense, yes, adding it here might help update-pinnings-run to pick up recipes that depend on it.

bioconda_utils/bioconda_utils-conda_build_config.yaml
mbargull: What do you think about this one? I'm rather concerned that `conda-build` will manage to collapse these list again :/.
IDK if it will work because we only duplicate an entry but not add a different one here (adding a different one made `conda-build` barf before).

mbargull: Oh, and some hours ago conda-forge updated to `numpy=1.17` -- so, probably best to update the conda-forge-pinnings version again ;).

dpryan79: Good to know, we might as well update then. The main issue with copying over their pinnings and modifying them is that it's easier to lose track of what our differences are. If you're really concerned maybe we merge and tag and then manually run a couple packages for testing. Since it's numpy and python we can use pybigwig as a test (it uses both with C so it needs the pinning).

dpryan79: I'll install this locally in a new environment and try this there.

dpryan79: It's good I started looking into this, we can't go above numpy 1.16 with python 2.7

dpryan79: Hmm, this doesn't want to build against python 3.9

dpryan79: This is definitely a problem. If we keep only python 2.7 and 3.9 then the zip_keys work properly, 2.7 builds use numpy 1.16 and 3.9 builds get numpy 1.19. However, once we add in python 3.6, 3.7, or 3.8, then only those python builds get used and the others ignored. They use the right numpy versions, but that's not terribly helpful.

mbargull: Ah, bummer. Yep, that sounds like the exact same "`conda-build` takes the intersection of values from `conda_build_config.yaml` and `bioconda-conda_build_config.yaml`"-issue.
So, for `numpy` itself we could go with the "append `.*`" workaround to get distinct sets.
As for `python_impl`: That one is used as a verbatim string (e.g., with `blabla  # [python_impl == "cpython"]` in conda-forge recipes). We don't use `python_impl` in our recipes currently, so it should generally not be important yet. And, more significantly, we only have a single `cpython` value in the configs for now, so the intersection remains just `cpython`.

TL;DR We could use `numpy: [ 1.16.*, 1.17.*, ... ]` for now. I'll add that change.

dpryan79: Good call, I'll run that locally to see if it resolves the issue.

dpryan79: Yup, looks like that solved it locally

dpryan79:
@mbargull This is ready to go from my side. Is there anything else from your end?

jmarshall:
Can we talk some more about libdeflate?

IMHO the real problem with libdeflate is that most things that depend on it shouldn't. As of a week ago:

```
Has host and run dependency on libdeflate (and also htslib):

recipes/arriba/meta.yaml
recipes/bcftools-gtc2vcf-plugin/1.9/meta.yaml [bundles htslib somehow]
recipes/cancerit-allelecount/meta.yaml [libdeflate on host only]
recipes/demuxlet/meta.yaml [also bundles htslib somehow]
recipes/hap.py/meta.yaml [libdeflate on host only]
recipes/ivar/meta.yaml
recipes/megadepth/meta.yaml [libdeflate on host only]
recipes/pheniqs/meta.yaml
recipes/pysam/0.15.2/meta.yaml [run htslib implicit due to samtools/bcftools]
~~recipes/wiggletools/meta.yaml~~ [no longer depends on libdeflate since PR bioconda-recipes#27401]

Has host and run dependency on libdeflate (but not htslib):

recipes/crispritz/meta.yaml [run only]
recipes/htslib/meta.yaml
recipes/nanopolish/meta.yaml [run only]
recipes/pysam/meta.yaml
recipes/staden_io_lib/meta.yaml
recipes/vt/meta.yaml
```

Most likely all of those in the first category do not themselves use libdeflate, but only get it via htslib. So culling their lists of dependencies will insulate them from libdeflate changes.

In the second category, htslib and staden_io_lib truly do use libdeflate. Pysam currently does but in the near future I plan to raise a PR to revert it to using the shared library from bioconda's htslib package instead — so it would move to the first category. I am unsure of the exact status of crispritz, nanopolish, and vt.

So when bumping those in the first category to rebuild them against a pinned htslib 1.12, I would propose to check whether they have a non-htslib-based libdeflate dependence and simultaneously remove their libdeflate dependency if not.

I expect this would largely take care of the libdeflate problem!

dpryan79:
@mbargull Yeah, we might be able to do without a libdeflate pinning, most of them predate the migration to conda-forge and fix of the run-time restrictions that it exports (or maybe it was an issue of the shared-library name change on OSX from .so to .dylib).

I think vt builds htslib internally, the others are using pysam (internal htslib build) or htslib, so we can remove them there (the libdeflate addition is to work around an old OSX issue I think).

dpryan79:
Gah, that should have been @jmarshall! Sorry

dpryan79:
So TLDR, I like what @jmarshall proposes.

dpryan79:
I'll tag this tonight and run the pinning update on bulk today or tomorrow. I'll temporarily blacklist the packages mentioned by @jmarshall (except htslib, staden_io, and pysam) and handle them separately after the migration.

PR #700, opened 2021-03-20
Ensure bioconda-utils uses the same version build container

dpryan79:
This now no longer uses :latest, but rather whatever its version is when pulling the docker build container.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,10 +41,10 @@ variables:
     run:
       name: Build the updated docker container
       command: |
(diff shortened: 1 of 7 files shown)


dpryan79:
I tested this locally and it seems to work. For releases we'll get tags like `0.16.22` whereas for intermittent commits it's of the form `0.16.22+0.gd5e0d9a.dirty`.

dpryan79:
@mbargull I very much appreciate how much more thorough you were than me :)

mbargull:
I appreciate about anything you do :heart: :)

PR #702, opened 2021-03-22
Update to linux-anvil-cos7-x86_64 build base and new base container images

mbargull:
1. Use condaforge/linux-anvil-cos7-x86_64 as build base

   - Update the build base system to Centos 7.
   - Use the "conda" user from the conda-forge base image during build.
     Write access to /opt/conda is given to "conda" via a group "lucky".
     -> Make sure to `chgrp` (change group) and `chmod g=u` (copy (root)
     users' permission to group) when changing /opt/conda in the image!
   - Add back libGL.so.1 to the build container.
     So it can be used for linking as well as non-container-based tests.

2. Use new base images and env creation container

   Base images are:
   - standard: quay.io/bioconda/base-glibc-busybox-bash:2.0.0 ~~1.1.0~~
   - extended: quay.io/bioconda/base-glibc-debian-bash:2.0.0 ~~1.1.0~~

   Container for environment creation via mulled:
   - quay.io/bioconda/create-env
     In conjunction with the entrypoints from the base images above and a
     postinstall step we inject, new containers will run in activated envs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,10 +41,10 @@ variables:
     run:
       name: Build the updated docker container
       command: |
-        docker build -t quay.io/bioconda/bioconda-utils-build-env:latest ./
-        docker history quay.io/bioconda/bioconda-utils-build-env:latest
-        docker run --rm -t quay.io/bioconda/bioconda-utils-build-env:latest sh -lec 'type -t conda && conda info -a && conda list'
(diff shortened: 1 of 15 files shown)


dpryan79 (approved):

mbargull:
(Merged without "all green" since OSX test failure was just a HTTP-related temporary thing and unrelated.)

PR #705, opened 2021-03-26
Fix ssh not being available in build container

mbargull:
Also split the `yum install` into two commands since a sole non-existing package will cause a failure then atleast.


dpryan79 (approved):

dpryan79:
@mbargull Can you tag a new release once you have this merged in? We can switch over bulk then.

PR #708, opened 2021-04-14
Fix licenses copying for Biocontainers

mbargull:
Small fix regarding license copying into the containers.
`bioconda/create-env:2.0.0` used `cp -l` for licenses which does not work for bind mounted volumes like we use via mulled.
`bioconda/create-env:2.1.0` drops the `-l` and just copies the license files.
Plus, with `2.0.0` errors during those copying were not caught properly, i.e., copying failed because of the hard-linking issue but the build only logged this, but did not fail as intended.


dpryan79:
Ah, that explains some of the weird "can't copy" messages I've seen in outputs :)

PR #713, opened 2021-05-31
Filter out skipped variants before env check

mbargull:
Addresses a failure encountered at https://github.com/bioconda/bioconda-recipes/pull/28721 where we skip the Python 2.7 build but `conda_build.api.render` fails because of dependencies incompatible with Python 2.7.


dpryan79 (approved):

dpryan79:
Sorry, didn't notice this until now. Looks good to me.

PR #738, opened 2022-02-10
fix: allow pkg_dir to be empty by creating a real conda channel there

jpfeuffer:
Additionally reindex after build. Copying is not enough.
As discussed on gitter.


dpryan79:
Obviously the failures have nothing to do with this PR. Note that this doesn't seem to be required on the CI nodes, as evidenced by a PR I'm working on at the moment:

image

Of course it won't hurt anything.

jpfeuffer:
Maybe it's the conda version? This requirement was introduced in 4.3.

Issue #736, opened 2022-02-08
Wrong quay image requested when attemping test locally due to wrong version.py generated by versioneer


DrYak:
Cross posting [this issue from bioconda-recipes](https://github.com/bioconda/bioconda-recipes/issues/32833).

I posted it initially over-there as it was a result of following the "contributing to bioconda" procedure, but given that this is a bug when invoking `bioconda-utils build`, it might actually be more relevant to post about it here.

DrYak:
I suspect that [this line of code is the culprit](https://github.com/bioconda/bioconda-utils/blob/ae0128e13b06a031cc0e1b74226faf6dd2e1cfae/bioconda_utils/docker_utils.py#L174) given that every call to this constructor leaves the parameter to its default (except in the tests).

It should be better to filter the `__version__` string to only the subpart that corresponds to the parts used for quay.io image tags.

dpryan79:
How did you install bioconda-utils? This should normally only happen if you install it from source rather than via conda.

DrYak:
On this last attempt I've run the `bootstrap.py`executable. (While doing the previous 0.3.5 release of smallgenomeutilities, I remember having similarish issue, both after attempting to update the build environment (`conda update --all`) and a doing a clean re-install with bootstrap).

So the `bioconda-utils` package is definitely _**not**_ self-compiled, but fetched from bioconda.

DrYak:
Interestingly, install the package myself, doesn't reproduce this:

```console
$ mamba create -n biocu-test bioconda-utils
…
$ conda activate biocu-test
$ ls ${CONDA_PREFIX}/lib/python*/site-packages/bioconda_utils/_version.py
/home/dryak/miniconda/envs/biocu-test/lib/python3.7/site-packages/bioconda_utils/_version.py
$ grep 'version' ${CONDA_PREFIX}/lib/python*/site-packages/bioconda_utils/_version.py
…
 "version": "0.17.10"
```

DrYak:
I would still suggest using a different default strategy:
```python
docker_base_image='quay.io/bioconda/bioconda-utils-build-env-cos7:{}'.format(__version__.split("+")[0])
```
As, I presume, such longer tags aren't expected anyway on quay.io

DrYak:
The steps used yesterday according to my history are:

```console
$ ./bootstrap.py /tmp/bioconda-build
$ source ~/.config/bioconda/activate
$ bioconda-utils lint --git-range master
[...]

PR #740, opened 2022-02-17
feat: update conda pinnings and build docs (xref #736)

dpryan79:
As mentioned by @DrYak in #736 , we need to update the local build instructions.

We could also add `conda mambabuild` instructions, which are handy for when you need to debug environment resolution issues.


dpryan79:
I'll use this as an excuse to fix the build here.

johanneskoester (approved):

PR #741, opened 2022-02-18
fix: avoid latest markupsafe if using old jinja

peterjc:
As per pallets/markupsafe#282 / pallets/markupsafe#283 markupsafe v2.1 breaks Jinja2 < 3.0, which causes problems like https://github.com/bioconda/bioconda-recipes/pull/33263

### DIFF
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -48,6 +48,7 @@ pyjwt=1.7.*               # githubhandler (JWT signing)
 beautifulsoup4=4.6.*
 galaxy-lib>=18.9.1
 jinja2>=2.10.1,<2.11.a0
+markupsafe<2.1           # markupsafe 2.1 breaks jinja2
 
 # docs
 sphinx>=4.1


peterjc:
This failure puzzles me, but hopefully makes sense to the bioconda team:

```
UnsatisfiableError: The following specifications were found
to be incompatible with the existing python installation in your environment:

Specifications:

  - conda=4.8.4 -> python[version='>=2.7,<2.8.0a0|>=3.6,<3.7.0a0|>=3.8,<3.9.0a0|>=3.7,<3.8.0a0']

Your python: python=3.9
```

dpryan79:
The error is indeed unrelated and due to the version of python in the base docker image having been upgrade. I'm fixing that in a separate PR.

dpryan79:
I'm pushing out a new bioconda-utils version this weekend, which will contain this fix. Thanks @peterjc !

marcelm:
New PRs still use bioconda-utils 0.17.10 (e.g. https://github.com/bioconda/bioconda-recipes/pull/33291). It appears that the BIOCONDA_UTILS_TAG needs to be updated as well. (I would have sent a PR, but couldn’t find where it is defined.)

PR #751, opened 2022-02-23
fix: allow finding dependencies on subpackages in dag

jpfeuffer:


jpfeuffer:
Are there tests for the graph "module"?

PR #755, opened 2022-03-16
feat: update dependencies and switch to conda boa/mambabuild

dpryan79:


dpryan79:
No clue if conda-build can use the new solver, but it'd certainly be nice! Otherwise we should use boa.

To do: update networkx

dpryan79:
OK that doesn't work. Time to try boa so `conda mambabuild` can be used.

dpryan79:
I haven't done a full build yet but this appears to do the rendering and everything now using boa's API. As that's likely fluid we should pin to boa versions.

PR #762, opened 2022-03-30
fix: https prompts to password, ssh to the rescue

afshinfard:
Since GitHub has removed password authentication, using https will prompt users to use their password and get errors. I'd suggest switching to ssh to avoid this confusion and a headache for users to figure things out; this change will automatically solve the issue for users with ssh authentication already setup.

### DIFF
--- a/docs/source/contributor/setup.rst
+++ b/docs/source/contributor/setup.rst
@@ -38,13 +38,21 @@ installed, and then run::
 
   git clone git@github.com:<USERNAME>/bioconda-recipes.git
 
+or::
+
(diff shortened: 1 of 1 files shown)


dpryan79:
Can you leave both in there and put an `or` in between them? For example, I don't use ssh keys because I need strictly separate keys in different places, but can use the same github token everywhere for use with https.

afshinfard:
> Can you leave both in there and put an `or` in between them? For example, I don't use ssh keys because I need strictly separate keys in different places, but can use the same github token everywhere for use with https.

Thanks for pointing out, just made the change;

dpryan79:
Awesome, thanks!

PR #795, opened 2022-08-03
feat: linter for missing space in dependency version contraints

### LINKED ISSUE #789: lint: missing space in dependency version pins
This example causes a cryptic error in the tests:
```
host:
  - python>3
```

This is because a space was missing!
```
host:
  - python >3
```

It might be nice if the linter catches missing spaces in dependencies (in the `build`, `host` and `run` sections), and returns a descriptive error for it.

siebrenf:
Identify missing whitespace between a package and its version constraints, before it causes a cryptic error in the build phase.
Also implements the `fix` method, and adds lint test cases.

notes: `check_recipe` was used instread of `check_deps` because the latter removes version constraints.

resolves #789

### DIFF
--- a/bioconda_utils/lint/check_syntax.py
+++ b/bioconda_utils/lint/check_syntax.py
@@ -4,8 +4,49 @@ These checks verify syntax (schema), in particular for the ``extra``
 section that is otherwise free-form.
 
 """
+import re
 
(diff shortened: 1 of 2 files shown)


daler (approved):
Thank you!

PR #803, opened 2022-09-12
fix: autobump uses correct version from common.sh

daler:
I had previously noticed some packages confusingly being built for py310 even though the only place we *should* have been building for py310 was on bulk. Figured something was pulling in conda-forge-pinnings maybe as part of an install process. But I just noticed https://github.com/bioconda/bioconda-recipes/pull/36889, which is the autobump bot picking up that we had updated pinnings...something that only happened recently in the master branch of bioconda-utils.

Turns out the autobump bot, which runs on CircleCI, was installing bioconda-utils straight from the master branch. This PR fixes it to use the method used everywhere else, which is to install the version specified in bioconda-common/common.sh.

This affected autobumped recipes where the autobump was successful and the PR was approved and merged with no changes. In such a case, the autobump PR built artifacts using the master branch of bioconda-utils which was then directly uploaded as part of the bioconda-bot merge process.

In practice, this effectively built new packages using the same versions and pinning as the bulk branch, which I guess saved some time on the bulk builds. But this now makes building of autobump PRs consistent with the building of regular PRs to hopefully prevent any issues in the future.

**Edit:** while looking through this CircleCI config, there's a lot of old cruft laying around. E.g., we're not doing anything with osx any more; docs have been moved to GitHub Actions; we're not building any containers here; we're not doing any testing. So now this PR also includes a round of cleanup and simplification of the CircleCI config to hopefully make it easier to grok and maintain.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,198 +1,79 @@
+# The autobump bot runs on CircleCI (rather than run on Azure Pipelines or
+# GitHub Actions) so that we avoid additional workload on those other services
+# which currently take care of PRs and bulk branch, respectively.
+
 version: 2
 
-variables:
-  restore_cache: &restore_cache
-    restore_cache:
-      keys:
-        - bioconda-utils-{{
-            checksum ".circleci/common.sh" }}-{{
-            checksum ".circleci/setup.sh" }}-{{
-            checksum ".circleci/config.yml" }}-{{
-            checksum "bioconda_utils/bioconda_utils-requirements.txt" }}-{{
-            arch }}
-  save_cache: &save_cache
-    save_cache:
-      key: bioconda-utils-{{
+jobs:
+  autobump:
+    machine:
+      image: ubuntu-2004:202201-02
+    steps:
+      - add_ssh_keys:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -180,17 +181,17 @@ jobs:
...
-  #      - schedule:
-  #          cron: "0 0,6 * * *"
-  #          filters:
-  #            branches:
-  #              only:
-  #                - master
-  #    jobs:
-  #      - build-docs:
-  #          context: org-global
+  bioconda-utils-build-docs:
```
bgruening, line 184: Is this change intended to be in this commit?

bgruening (commented):
Awesome detective work @daler!

bgruening (approved):

.circleci/config.yml
daler: not intended! decided to simplify and refactor though in [1631bf3](https://github.com/bioconda/bioconda-utils/pull/803/commits/1631bf3ea34dc71599f7f95d2dfff21984414f82)

PR #807, opened 2022-09-13
docs: reflect Jillians leave of the core team in the docs

johanneskoester:


johanneskoester:
> Should we add a "Core Alumni" section and move the former members there? It'd be good to continue to acknowledge their former involvement :)

Done!

dpryan79:
Excellent!

PR #822, opened 2022-10-23
feat: switch data packages to bioconductor-data-packages

dpryan79:
Xref: https://github.com/bioconda/bioconda-recipes/pull/37578

### DIFF
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -11,6 +11,7 @@ import os
 import re
 from collections import OrderedDict
 import logging
+import json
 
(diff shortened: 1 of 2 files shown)


dpryan79:
I think the idea was to have a way to update locations after the fact still, since sometimes the URLs get changed. Allowing the json file to get updated simplifies that.

mbargull:
Ah, you mean like adding a new mirror and/or when URLs change to some `foo/archive/bar`? Hm.. makes sense.
I'm still a bit unsure if we really want a separate package for that. I'm always a little wary when putting different things into a single package (and esp. single file).
I'll have to think about your point...

PR #832, opened 2022-12-09
docs: support anchors in <details>

daler:
This builds on ideas in @marcelm's #831 by keeping the front page streamlined while still supporting anchors that can be used as permalinks.

First, the `details_ext.py` Sphinx extension was updated to support adding an optional `:anchor:` argument. Using this option causes an anchor to be created at the top of the `<details>` element, which repeats the summary text followed by a `¶` with the permalink.

Then there's additional JS in the `docs/source/templates/layout.html` that, upon visiting an anchor, looks for the closest `<details>` to the target and makes it visible.

So once built, the link https://bioconda.github.io/index.html#do-not-modify-condarc will link to, and open if folded, the details section about how `-c bioconda` is not sufficient (the original inspiration for #831).

### DIFF
--- a/docs/source/_ext/details_ext.py
+++ b/docs/source/_ext/details_ext.py
@@ -23,19 +23,24 @@ The resulting HTML has the structure:
     </div>
 </details>
 
-That is, you can style the part that expanded with CSS like:
+That is, you can style the part that expanded with CSS like::
 
-.. code-block:: css
+    details div { }
 
-    details div {
-    }
+And the label (when collapsed) with::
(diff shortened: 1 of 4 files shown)


daler:
See https://github.com/bioconda/bioconda-utils/suites/9780753813/artifacts/470242736 for latest built docs

daler:
Ah, thanks for pointing that out. I'll look into a firefox fix, and worst-case pull these \<details> tags out into a separate FAQ page.

PR #835, opened 2023-01-25
chore: add CI to changevisibility of private containers

rpetit3:
This PR adds a script to set private repos on Biocontainers to public. Using the Quay API, it will get all repos under `biocontainers`, and if private it will submit a `changevisibilty` API request for that repo.

A few files (info about each repo, and any repos that were set public) are kept as artifacts and deleted after 7 days.

An example usage is available here: https://gist.github.com/rpetit3/c1962790dee7fa513722eb575c348622?permalink_comment_id=4448741#gistcomment-4448741

Notes:
- Quay doesn't have a way to only get private repos, so each run pulls down all available repos (public and private)
- If needed we could integrate the script into `bioconda_utils`.

### DIFF
--- a/.github/quay-namespace-info.py
+++ b/.github/quay-namespace-info.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python
+"""
+A simple script to query API to get public/private status of all repos under a namespace.
+
+Caveat being, without an API key, you will only ever see the public repos.
+"""
+PROGRAM = 'quay-namespace-info'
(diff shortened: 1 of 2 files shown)


dpryan79 (approved):

dpryan79:
The web UI on quay.io is painful, I hope the API is more useful there. :+1: from me, this is worth a shot.

rpetit3:
Thank you!

PR #837, opened 2023-02-23
build: update htslib version to 1.17

maarten-k:
As needed by:
- https://github.com/bioconda/bioconda-recipes/pull/39501 
-  https://github.com/bioconda/bioconda-recipes/pull/39499

### DIFF
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -27,7 +27,7 @@ glpk:
   - 5
 
 htslib:
-  - 1.16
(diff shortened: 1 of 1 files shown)


dpryan79:
Hmm, we have an issue with our OSX testing. That's irrelevant here though, thanks for the timely PR :) I'll merge this in though I'm not sure exactly when we should start the migration. We'll have to discuss that within the core team.

jmarshall:
Another one for the core team's agenda: bumping conda-forge-pinning to migrate from OpenSSL 1.1.1 to 3. (Htslib is probably one of bioconda's dominant users of OpenSSL.)

PR #838, opened 2023-02-24
fix: success check does not consider pkg_dir #836

### LINKED ISSUE #836: Non-zero exit code but package get's build and uploaded successfully
Hi, our bioconda build always returns with exit code != 0 because it doesn't find the tar.gz file 

https://github.com/OpenMS/OpenMS/issues/6514#issuecomment-1420670234

We disabled upload during build and the perform the upload  separately - which works:
```
Using Anaconda API: https://api.anaconda.org
[DEBUG] Using token from command line args
Using "openms" as upload username
Processing '/home/builder/jenkins-ws/ws/openms/ntly/conda/output/linux-64/libopenms-3.0.0dev-h135471a_0.tar.bz2'
Detecting file type...
[DEBUG] Testing if conda package ..
[DEBUG] This is a conda package
File type is "conda"
Extracting conda package attributes for upload
Creating package "libopenms"
Creating release "3.0.0dev"
Uploading file "openms/libopenms/3.0.0dev/linux-64/libopenms-3.0.0dev-h135471a_0.tar.bz2"
[WARNING] Distribution "linux-64/libopenms-3.0.0dev-h135471a_0.tar.bz2" already exists. Removing.
Upload complete
```
Could this be an error in build?

jpfeuffer:
completely untested but it is an obvious bug.
If you want to change the package dirs earlier in the code somewhere, please suggest changes


jpfeuffer:
@bgruening @daler Ready to review. Somehow I cannot request it from the normal UI.

daler:
@jpfeuffer just to clarify, is OpenMS is using bioconda-utils for internal builds?

daler:
Ah, ok. Can you point to where (and how) you are calling it then? I bet if it's not *exactly* how we call it in bioconda-recipes then that might introduce behavior that we don't otherwise test for.

jpfeuffer:
Sure! Yes, you might not be using the `--docker`-specific `--pkg_dir` option to specify a directory to be mounted into the docker container. We use it, so we can reuse already-built packages for the next run and generally have access to the finished packages after the build. I did not check if we can change the upload behavior otherwise that would have been an alternative.

```bash
# output must be a valid conda "channel" (otherwise I think I encountered problems)
mkdir -p output/noarch
touch output/noarch/repodata.json
conda activate bioconda
mamba update -y bioconda-utils
conda index output

git clone github.com/bioconda/bioconda-recipes && cd bioconda-recipes
bioconda-utils build --docker --pkg_dir $WORKSPACE/output --packages openms-meta
bioconda-utils build --docker --pkg_dir $WORKSPACE/output --packages pyopenms
```

Note: you cannot run them together because of #751

daler:
So the default behavior is for `--pkg_dir` to not be set, which in turn defaults to the identified `conda-build` directory on the host, which is where the built recipes will be copied to and mounted read-only into the container. So the default behavior does, I think, what you are aiming to do.

Does it work if you remove the `--pkg_dir` arg?

jpfeuffer:
Yes, I think it works without `--pkg_dir` but somehow we ended up using a subfolder in the workspace (persistency, permission reasons I guess). We might be able to reevaluate the thing but I think it would be great if the build also works with `--pkg_dir`.

daler:
I _think_ it would be ok to merge as-is since we don't use `--pkg_dir` anywhere in production, but merging without tests always makes me nervous. And there are currently bigger issues to resolve that affect the entire channel (and very limited time to do so). 

Can you write a test for this in the `test` dir, that fails in master branch but passes with your addition?

test/test_utils.py
```diff
@@ -126,6 +126,31 @@ def single_build(request, recipes_fixture):
     yield recipes_fixture.pkgs['one']
     for pkg in recipes_fixture.pkgs['one']:
         ensure_missing(pkg)
+        
+
+@pytest.fixture(scope='module', ids=IDS)
```
jpfeuffer, line 131: @daler I have no idea how to see if this actually was tested and what its output was.

test/test_utils.py
```diff
@@ -126,6 +126,31 @@ def single_build(request, recipes_fixture):
...
+    logger.error("Fixture: Building 'one' within docker with pkg_dir")
+    build.build(
+        recipe=recipes_fixture.recipe_dirs['one'],
+        pkg_paths=recipes_fixture.pkgs['one'],
+        docker_builder=docker_builder,
+        mulled_test=mulled_test,
+    )
+    logger.error("Fixture: Building 'one' within docker and pkg_dir -- DONE")
+    yield recipes_fixture.pkgs['one']
+    for pkg in recipes_fixture.pkgs['one']:
```
jpfeuffer, line 152: I think instead of using this "ensure_missing" one could just check the BuildResult returned by build.

jpfeuffer:
@daler Ready. I am 90% sure it will fail on master.

jpfeuffer:
@daler Do you need anything else?

timosachsenberg:
Hi, has there been any progress with this PR?

bioconda_utils/build.py
```diff
@@ -116,6 +117,16 @@ def build(recipe: str, pkg_paths: List[str] = None,
                                         env=whitelisted_env,
                                         noarch=is_noarch)
             # Use presence of expected packages to check for success
+            if (docker_builder.pkg_dir is not None):
+                    platform = os.environ.get('OSTYPE', sys.platform)
+                    if platform.startswith("darwin"):
+                        platform = 'osx'
+                    elif platform == "linux-gnu":
+                        platform = "linux"
```
johanneskoester, line 125: The platform can be inferred from `RepoData().native_platform()`.

johanneskoester: Also, make sure that you indent only with 4 spaces.

johanneskoester (changes requested):

poshul:
`FAILED test/test_utils.py::test_single_build_pkg_dir - AssertionError: bug: ensure to load config before instantiating RepoData.
ERROR test/test_utils.py::test_single_build_only[with docker] - AssertionError: bug: ensure to load config before instantiating RepoData.`

jpfeuffer:
Sorry, I don't know what that means. I merely added the review suggestion. I guess this error is also why the suggested code was not used in the place where I copied it from 🤷.
I guess we have to wait for further suggestions..

timosachsenberg:
any update here?

poshul:
@johanneskoester @daler any updates?

timosachsenberg:
any news?

bioconda_utils/build.py
```diff
@@ -138,6 +139,12 @@ def build(recipe: str, pkg_paths: List[str] = None,
                                         env=whitelisted_env,
                                         noarch=is_noarch)
             # Use presence of expected packages to check for success
+            if (docker_builder.pkg_dir is not None):
```
bgruening, line 142:
```suggestion
            if docker_builder.pkg_dir is not None:
```

bgruening (approved):
Thanks everyone!

PR #840, opened 2023-02-26
ci: fix macOS tests

daler:
Current tests are failing, seems to be an issue with using the wrong Python version. Using this PR to troubleshoot the CI builds.

Edit: honestly unclear how this was actually working correctly in the first place...maybe something changed with the macOS runners? Anyway, this now sets up macOS similar to how it's done in bioconda-recipes. But rather than copy/paste that setup/install code and have to maintain it in two places, I pulled that out into [a separate script in bioconda-common](https://github.com/bioconda/bioconda-common/pull/31) which is intended to be `curl|bash`ed in GitHub Action yaml files (and which is done here). A near-term next step is to replace the bioconda-recipes macOS setup with calling that same script.

### DIFF
--- a/.github/workflows/GithubActionTests.yml
+++ b/.github/workflows/GithubActionTests.yml
@@ -68,6 +68,8 @@ jobs:
         which conda
         echo $PATH
         python setup.py install
+        conda install -y -f test-requirements.txt
+
     - name: Run tests
       run: |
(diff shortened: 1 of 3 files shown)


daler:
Forcing merge without review since it's known to be broken now anyway!

PR #843, opened 2023-02-28
fix: Change logic for checking if there are missing commits in a branch vs…

dpryan79:
… master

This should prevent the bot from closing PRs prematurely, since it's now checking whether all commits to a recipe in master are synced with a PR. If not, the branch is deleted and recreated.


daler:
oh wow I hope this does it!

Do we have a good way of testing this in the test suite?

dpryan79:
@daler I don't have a good way to test this. The best idea I can come up with is to install this locally and have it try to update a specific recipe PR. Doing that before and after touching something like the blacklist should presumably work for a test.

daler (approved):

PR #853, opened 2023-03-09
fix: load config before instantiating repodata in artifact upload

johanneskoester:


johanneskoester:
Skipping tests as the change is currently untested anyway.

PR #872, opened 2023-04-28
perf: upgrade to latest conda, conda-build, and boa versions

johanneskoester:


Dockerfile
```diff
@@ -21,9 +21,9 @@ RUN yum install -y mesa-libGL-devel \
 # This changes root's .condarc which ENTRYPOINT copies to /home/conda/.condarc later.
 RUN . /opt/conda/etc/profile.d/conda.sh && \
     conda config \
-      --prepend channels defaults \
```
daler, line 24: Can we use [install-and-set-up-conda.sh](https://github.com/bioconda/bioconda-common/blob/master/install-and-set-up-conda.sh) to install in this dockerfile? Otherwise this risks getting out-of-sync with how it's being installed everywhere else.

johanneskoester: Yes, we should do that. But let us do that in a separate PR. There are some steps in this dockerfile that I do not fully understand, such that it would need more time. For now, I just want to upgrade the conda and mamba versions as quickly as possible.

daler: Sounds good. The base linux-anvil image used here has a lot of moving parts as well; I have a local branch that completely rewrites the dockerfile for simplicity but that can be for later.

PR #878, opened 2023-05-07
perf: update to latest conda-forge pinnings and ditch building for Python 3.7 and 3.6

johanneskoester:


bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -21,44 +21,35 @@ pin_run_as_build:
...
-  - 4.2.*
+  - "4.2.*"
 
 python:
   - 3.10.*  *_cpython
   - 3.9.*  *_cpython
   - 3.8.*  *_cpython
-  - 3.7.*  *_cpython
-  - 3.6.*  *_cpython
   - 2.7.*  *_cpython
```
bgruening, line 39: It's their a reason to keep python2? People could add local pinning files if they really need to.

bgruening (commented):
LGTM, I have just the usual python2 question ❓ :-)

bioconda_utils/bioconda_utils-conda_build_config.yaml
johanneskoester: Good point. You are completely right I think. I will remove it.

bgruening (approved):
We should create an issue and inform the community before we merge. Maybe wait 1-2 days?

johanneskoester:
> We should create an issue and inform the community before we merge. Maybe wait 1-2 days?

It would first be used only in bulk, right? Couldn't we do that already now?

johanneskoester:
Likely the bulk update will need a couple of days, so that there is enough time to inform the community while bulk is running?

dpryan79:
Seems reasonable to merge now, since this won't affect master immediately. I also kind of doubt there will be much pushback regarding python 3.6 and 3.7

dpryan79:
Kind of curious what the feedback will be to dropping 2.7. It's probably time to finally do it.

dpryan79 (commented):
Should we add 3.11?

bgruening:
I like the idea of adding 3.11. but I expect some pain as well ;-)

johanneskoester:
Regarding 3.11 we should not do that since conda-forge has not yet done it in their pinnings.

jamestwebber:
> Regarding 3.11 we should not do that since conda-forge has not yet done it in their pinnings.

Is this still an issue? I can install most packages from `conda-forge` for 3.11, but nothing from bioconda.

corneliusroemer:
It sounds like conda-forge is now using 3.11 by default. This was announced on September 25 2023:

> At the same time, we have stopped the Python 3.11 migration and added it to the list of default Python versions on conda-forge.
> https://conda-forge.org/docs/user/announcements.html#python-3-12-migration-and-python-3-11-by-default

PR #890, opened 2023-05-21
feat: add subcommand to skiplist a given recipe using the new recipe specific mechanism; in addition, some fixes for the new skiplisting approach, and renaming blacklist into skiplist

johanneskoester:


bgruening:
Thanks @johanneskoester! Great work. How much work is it to call it skiplist or something similar? We should avoid the term blacklisting going forward imho.

johanneskoester:
Oh, very good point! I did not think about that. You are right, I will refactor it ASAP.

PR #894, opened 2023-05-23
fix: various little fixes for build failure records and automatic skiplisting

johanneskoester:


johanneskoester:
There is a seemingly temporal problem with the https certificate of bioconductor that makes the test fail, which is definitely unrelated to this PR. Hence, I will merge this now.

bioconda_utils/utils.py
```diff
@@ -561,8 +570,8 @@ def temp_os(platform):
         sys.platform = original
 
 
-def run(cmds: List[str], env: Dict[str, str]=None, mask: List[str]=None, live: bool=True,
-        mylogger: logging.Logger=logger, loglevel: int=logging.INFO,
+def run(cmds: List[str], env: Dict[str, str]=None, mask: List[str]=None, mask_envvars: bool=False, live: bool=False,
```
aliciaaevans, line 573: @johanneskoester The log messages from inside docker are not getting to stdout. What was the reasoning for this change of default to `live=False`? Would there be an issue to pass in `live=True` for the docker build?  https://github.com/bioconda/bioconda-utils/blob/88680cf4340ca0eeaa38cb6da6eb0fc1fc126543/bioconda_utils/docker_utils.py#L477

Issue #895, opened 2023-05-25
GithubRelease assets are not found via autobump


holtjma:
Hello,

I've having issues getting autobump to correctly identify a version change for this recipe: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/hiphase/meta.yaml.  It _appears_ to be an issue between the GithubRelease hoster and what GitHub returns via HTML now.  Couple notes:

- I verified that autobump is correctly trying to use GithubRelease
- I verified that it's fetching the correct URL in this instance
- When I checked the downloaded HTML, I noticed that the links required (of this form `https://github.com/PacificBiosciences/HiPhase/releases/download/v{{ version }}/hiphase-v{{ version }}-x86_64-unknown-linux-gnu.tar.gz`) don't actually show up in the HTML.  
- When I reviewed it with web, it appears to be lazily loaded via a second URL of this pattern: https://github.com/PacificBiosciences/HiPhase/releases/expanded_assets/v0.10.0
- That pattern does show up in the autobump download

Below are some logs with things I added while trying to debug:
```
$ bioconda-utils autobump recipes/ config.yml --packages hiphase --loglevel debug
...
08:09:03 BIOCONDA DEBUG Matching url 'https://github.com/PacificBiosciences/HiPhase/releases/download/v0.8.1/hiphase-v0.8.1-x86_64-unknown-linux-gnu.tar.gz'                                   
08:09:03 BIOCONDA DEBUG GithubRelease matched https://github.com/PacificBiosciences/HiPhase/releases/download/v0.8.1/hiphase-v0.8.1-x86_64-unknown-linux-gnu.tar.gz with {'account': 'PacificBiosciences', 'project': 'HiPhase', 'prefix': '', 'version': '0.8.1', 'fname': 'hiphase-v0.8.1-x86_64-unknown-linux-gnu.tar.gz', 'ext': ''}
08:09:03 BIOCONDA DEBUG Scanning with GithubRelease                                                                                                                                            
08:09:03 BIOCONDA DEBUG before: <bioconda_utils.aiopipe.AsyncRequests object at 0x7f9b61f85eb0>, 0.8.1                                                                                         
08:09:03 BIOCONDA DEBUG OHH start: <bioconda_utils.aiopipe.AsyncRequests object at 0x7f9b61f85eb0>, 0.8.1                                                                                      
08:09:03 BIOCONDA DEBUG HH exclude: {'version', 'fname'}                                                                                                                                       
[...]

PR #896, opened 2023-05-25
fix: adds alternate lookup if base lookup fails

holtjma:
This is a first pass attempt at fixing #895.

Main changes are updating the GithubRelease hoster to look for alternate paths via the GitHub API.  The API sends a JSON response (example: https://api.github.com/repos/PacificBiosciences/HiPhase/releases) that contains each release along with a list of assets.  The assets are searched for files that match the provides pattern.  Currently, this first pass puts this as a backup to the main one (which I think is always failing now).  I also did not update the tests, I think someone may have a better idea of how to do that.  Happy to help push this further, but wanted to get the ball rolling.

### DIFF
--- a/bioconda_utils/hosters.py
+++ b/bioconda_utils/hosters.py
@@ -326,7 +326,61 @@ class GithubBase(OrderedHTMLHoster):
 class GithubRelease(GithubBase):
     """Matches release artifacts uploaded to Github"""
     link_pattern = r"/{account}/{project}/releases/download/{tag}/{fname}{ext}?"
+    alt_releases_formats = ["https://api.github.com/repos/{account}/{project}/releases"]
 
+    async def get_versions(self, req, orig_version):
+        # first, try the older version when HTML worked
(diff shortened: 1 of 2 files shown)


dpryan79:
Wow, yeah that DESeq2 failure is definitely not your fault, I suspect bioconductor has removed the file.

dpryan79:
@johanneskoester @mbargull Any thoughts on this PR?

holtjma:
@johanneskoester @mbargull @dpryan79 

Circling back on this, any thoughts on the functional changes and/or fixing the test?  Happy to look into making the changes, just need to know what's missing here.

johanneskoester (approved):
Looks good to me. I've updated with the master branch, which should fix the test issue with deseq if I remember correctly.

holtjma:
@johanneskoester I fixed that previous error by just bumping the versions (assuming this is fine).  However, there's some new error in the OSX tests.  I took a quick look, but it's not obvious to me what's happening there.

PR #908, opened 2023-08-09
feat: lint missing run exports

johanneskoester:


johanneskoester:
Comment by @mbargull:

What shall we do in case of non-standard versioning, e.g. calendar versioning?

bioconda_utils/lint/check_build_help.py
```diff
@@ -136,3 +136,51 @@ class cython_needs_compiler(LintCheck):
...
+    This ensures that the package is automatically pinned to a compatible version if
+    it is used as a dependency in another recipe.
+    This is a conservative strategy to avoid breakage. We came to the 
+    conclusion that it is better to require this little overhead instead
+    of trying to fix things when they break later on.
+    This holds for compiled packages (in particular those with shared
+    libraries) but also for e.g. Python packages, as those might also
+    introduce breaking changes in their APIs or command line interfaces.
+
+    Add run_exports to the recipe like this::
```
johanneskoester, line 153:
```suggestion
    We distinguish between three cases. If the software follows semantic versioning (or it has at least a normal version string (like 1.2.3) and the actual strategy of the devs is unknown), add run_exports to the recipe like this::
```

johanneskoester:
> No constraint for max_pin in such a case?

Decided in our meeting, and now reflected in the lint documentation.

bgruening (approved):
Looks good to me!

PR #914, opened 2023-08-25
docs: try to clarify some confusion with run_exports.

johanneskoester:


jmarshall:
If you want bioconda contributors to buy into this, you should also add an explanation of how doing this will make bioconda better.

johanneskoester:
> If you want bioconda contributors to buy into this, you should also add an explanation of how doing this will make bioconda better.

Indeed. We have a statement about that in the PR template but maybe it should be rephrased in case it is not clear enough. Also here it makes sense of course. I will try to come to that myself as well, but any PRs that improve the docs are of course always welcome.

PR #915, opened 2023-08-27
fix: Update htslib pinning to 1.18 (released 2023-07-25)

jmarshall:
Samtools/bcftools/htslib 1.18 were released over a month ago.

To make samtools (bioconda/bioconda-recipes#42168) and bcftools (bioconda/bioconda-recipes#42167) 1.18 packages built against htslib 1.18, this pinning must be updated first.


jmarshall:
Instead of updating this pinning, I would now recommend removing it (#917) and patching htslib dependency repodata (bioconda/bioconda-recipes#42895).

PR #923, opened 2023-10-09
feat: add support to run build for recipes with linux-aarch64 additional-platforms

Yikun:
This patches add support in bioconda-utils to run on linux-arm if the following is added to the recipe:
```yaml
extra:
  additional-platforms:
    - linux-aarch64
```

- Add a `check_native_platform_skippable` to checkout should we skip the recipes in current nativei platform
- Add `extra_additional_platforms` property to recipe object
- Add UT to test above function
```
py.test --durations=0 test/test_utils.py::test_native_platform_skipping
py.test --durations=0 test/test_recipe.py::test_recipe_extra_additional_platforms
```
- E2E test
```
bioconda-utils build recipes config.yml --git-range origin/master HEAD
bioconda-utils build recipes config.yml --git-range origin/master HEAD --force
```
- Fix test trigger by adding `fetch-depth`

related: https://github.com/bioconda/bioconda-recipes/pull/40550

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -260,23 +260,22 @@ def get_subdags(dag, n_workers, worker_offset):
     return subdags
 
 
-def check_native_platform_skippable(recipe_folder: str, recipe: str):
+def do_not_consider_for_additional_platform(recipe_folder: str, recipe: str, platform: str):
     """
     Given a recipe, check this recipe should skip in current platform or not.
 
     Arguments:
       recipe_folder: Directory containing possibly many, and possibly nested, recipes.
-      recipe: recipe name
+      recipe: Relative path to recipe
+      platform: current native platform
 
     Returns:
       Return True if current native platform are not included in recipe's additional platforms (no need to build).
     """
     recipe_obj = _recipe.Recipe.from_file(recipe_folder, recipe)
-    native_platform = utils.RepoData().native_platform()
     # On linux-aarch64 env, only build recipe with linux-aarch64 extra_additional_platforms
-    if native_platform == "linux-aarch64":
+    if platform == "linux-aarch64":
         if "linux-aarch64" not in recipe_obj.extra_additional_platforms:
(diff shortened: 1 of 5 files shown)


martin-g (approved):
LGTM

johanneskoester (changes requested):
Great, thanks! Would you mind adding a test case?

bioconda_utils/build.py
```diff
@@ -259,6 +260,27 @@ def get_subdags(dag, n_workers, worker_offset):
...
+
+    Returns:
+      Return True if current native platform are not included in recipe's additional platforms (no need to build).
+    """
+    recipe_obj = _recipe.Recipe.from_file(recipe_folder, recipe)
+    native_platform = utils.RepoData().native_platform()
+    # On linux-aarch64 env, only build recipe with linux-aarch64 extra_additional_platforms
+    if native_platform == "linux-aarch64":
+        if "linux-aarch64" not in recipe_obj.extra_additional_platforms:
+            logger.info("BUILD SKIP: skipping %s for %s platform", recipe, native_platform)
```
johanneskoester, line 279: I would move the log message into the if statement where this function is used below.

bioconda_utils/build.py
```diff
@@ -259,6 +260,27 @@ def get_subdags(dag, n_workers, worker_offset):
     return subdags
 
 
+def check_native_platform_skippable(recipe_folder: str, recipe: str):
```
johanneskoester, line 263:
```suggestion
def do_not_consider_for_additional_platform(recipe_folder: str, recipe: str):
```

johanneskoester (changes requested):

bioconda_utils/build.py
Yikun: Yes, I also considered before. So the `native_platform` should be moved out of function to help log print, the testcase can be simpler

johanneskoester (approved):

PR #930, opened 2023-10-27
feat: Enable Live logs and add option to disable

aliciaaevans:
- Enable live logging except when a new `--disable-live-logs` option is enabled.
- This will improve troubleshooting of CI builds while having the option to avoid excessive log messages.


johanneskoester (approved):

aliciaaevans:
Per core meeting today, I'm going to change this to default to live logging.

bgruening (approved):
Great stuff!

PR #939, opened 2023-11-24
feat: live logs for mulled build

aliciaaevans:
Enable live logs for mulled-test.

The `--disable-live-logs` option was added in  https://github.com/bioconda/bioconda-utils/pull/930 but was only for the build. This will enable live logs for the mulled test.

Draft because it still needs testing.


daler:
Tested over in https://github.com/bioconda/bioconda-recipes/pull/43995, looks good.

daler (approved):

PR #944, opened 2023-12-02
feat: bioconductor improvements

daler:
Some fixes identified when preparing for the Bioconductor 3.18 updates:

**run_exports.** `bioconductor-skeleton` now adds run_exports to the recipes it builds. Since it's unclear to what extent all bioconductor developers adhere to semantic versioning, I opted for `max_pin="x.x"`, even in the cases where versions start with `0.`.

**Better build numbers.** `bioconductor-skeleton update-all-packages` will now only modify the build number if it's needed. It checks what's available in the channel. So running the command multiple times (sometimes needed if the update process flakes out partway through) will not keep bumping the build number. This also uses the same check as the linting, which will hopefully reduce the frequency of build number adjustments in bulk.

**Fix manual build failure annotation.** Allow category & reason to show up in manual build failure yaml via `bioconda-utils annotate-build-failures` CLI

**More flexible dag.** `bioconda-utils dag` now supports globs. So now with `bioconda-utils dag --packages bioconductor-* | grep -v INFO > dag.gml` and then loading `dag.gml` into [Gephi](https://gephi.org/), you can visually identify which packages should be prioritized for building first. This example sizes nodes by out-degree and labels only those with >125 outgoing edges.

image

**And minor stuff.** The pinning for bioconductor-data-packages needed a space; there was a build-failure yaml that was empty which crashed osx builds.

### DIFF
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -1249,7 +1249,19 @@ def write_recipe(package, recipe_dir, config, bioc_data_packages=None, force=Fal
             (updated_version == current_version) and
             (updated_meta != current_meta)
         ):
-            proj.build_number = int(current_build_number) + 1
+            # Sometimes when updating all packages, the updating process fails
+            # partway. Re-running the updating process should not bump the
+            # build number if no builds for this version exist yet in the repo.
+            existing_bldnos = utils.RepoData().get_package_data(
+                key="build_number",
+                name="bioconductor-" + proj.package.lower(),
+                version=updated_version
+            )
+            if not existing_bldnos:
+                proj.build_number = 0
+            else:
+                proj.build_number = sorted([int(i) for i in existing_bldnos]) [-1] + 1
+
         if 'extra' in current_meta:
             exclude = set(['final', 'copy_test_source_files'])
             proj.extra = {x: y for x, y in current_meta['extra'].items() if x not in exclude}
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -959,7 +959,7 @@ class BioCProjectPage(object):
                 'build', OrderedDict((
                     ('number', self.build_number),
                     ('rpaths', ['lib/R/lib/', 'lib/']),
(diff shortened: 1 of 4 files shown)


bioconda_utils/bioconductor_skeleton.py
```diff
@@ -959,6 +959,7 @@ def sub_placeholders(x):
                 'build', OrderedDict((
                     ('number', self.build_number),
                     ('rpaths', ['lib/R/lib/', 'lib/']),
+                    ('run_exports', f'{{{{ pin_subpackage("bioconductor-{self.package_lower}", max_pin="x") }}}}'),
```
bgruening, line 962: Isn't it x.x here?

bgruening (approved):

bioconda_utils/bioconductor_skeleton.py
bgruening: I don't know. My gut feeling is that we with every release of BioC package devs have all freedom to deprecate functions etc ... so if I depend on deseq2 1.40 today I don't want to have deseq2 1.80 in a few years?

In other words, I do assume breakages in x.x releases with every BioC release.

bgruening: Keep in mind that I have no clue. And whatever you have done until now is better than what we have done so far.

dlaehnemann: Thanks to @aliciaaevans for pointing me here. I just quickly want to chime in and second (or third) the decision to `max_pin="x.x"`. I previously went through the bioconductor developer docs and talked to a bioconductor package maintainer who basically confirmed what @bgruening is saying above. Namely, that every minor version bump can break anything, but that package versions from the same bioconductor release are assumed to be compatible with each other. For cross-reference, the original discussion is here:
https://github.com/bioconda/bioconda-recipes/issues/43905#issuecomment-1783290046

dlaehnemann:
Also, just to document this, there was a suggestion by @mfansler to run unit tests whenever possible when including `run_exports`. See here for the reasoning:
https://github.com/bioconda/bioconda-recipes/issues/43905#issuecomment-1781885663

bioconda_utils/bioconductor_skeleton.py
daler: `x.x` it is!

PR #945, opened 2023-12-08
fix: add local channel for docker builds

aliciaaevans:
The local folder channel was not included in the command, causing newly built packages on the `bulk` branch to not be usable to their descendants on the same worker because of strict channel priority.


aliciaaevans:
tested this on bulk branch and it looks to have successfully fixed the issue.

PR #946, opened 2023-12-12
fix: Fix version number check in repodata_patches_no_version_bump()

jmarshall:
PR #855 was merged without review or, apparently, any testing :smile:


aliciaaevans (changes requested):
Actually per the docs, at least one commit message needs to have the `fix:` prefix.

jmarshall:
@aliciaaevans: I have adjusted the commit message. Until the linter enforces conventional commit formatting on the right piece of text, IMHO it would be reasonable for maintainers to simply make commit message adjustments themselves when merging.

aliciaaevans (approved):

PR #948, opened 2023-12-14
feat: Bulk build failure wiki

aliciaaevans:
In order to have a `bulk` branch-specific Build Failures wiki page, need to allow for a git range to be specified for which recipes will be included. This uses the same method of determining whether a recipe is included as the build command. (i.e., it's not diffing the build failure files between branches.)

Also added handling for recipes that exist but have no corresponding package in the channel. That avoids errors from new Bioconductor recipes during a bulk update.

### DIFF
--- a/bioconda_utils/build_failure.py
+++ b/bioconda_utils/build_failure.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import time
 from typing import Optional, Union
 import subprocess as sp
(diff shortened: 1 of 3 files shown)


aliciaaevans:
Implemented this on the bulk branch: https://github.com/bioconda/bioconda-recipes/wiki/build-failures-bulk

daler (approved):

PR #954, opened 2024-02-17
DRAFT: unify containers testing ground

daler:
Using this to test some aspects of #953, specifically, that we have a new bioconda-utils version but need to use existing base containers.


daler:
Looking good -- merging this into `unify-containers` and closing.

PR #959, opened 2024-02-27
feat: simplify unify containers

daler:
This takes the ideas from the code review on #953 and simplifies by refactoring how everything is configured. See https://github.com/bioconda/bioconda-utils/pull/953#issuecomment-1958104290 for the specific proposed items, which are addressed here.

The only major difference from those proposed items is storing the conda/mamba versions, which are not clearly defined in requirements. For example, I think mamba comes in from the base image (quay.io/condaforge/linux-anvil-cos7-x86_64). So I kept the same strategy as before, where the create-env depends on the build-env and inspects the conda/mamba versions used there.

### Edits 2024-08-10

Here's an overview of the changes...

This is strictly for the *base* images. They are used for creating images for packages, and once this is merged they will will be pushed to the same quay.io tags as existing base images. But this PR does not deal with creating *package* images.

(if the nomenclature of images is confusing, see [dockerfile inventory](https://bioconda.github.io/developer/dockerfile-inventory.html) for a guide)

The package building has been unified to use a single `build.sh` script which takes an image directory as its only argument. That image directory should have `prepare.sh`, `Dockerfile`, and `Dockerfile.test`. Each `<IMAGE DIR>/prepare.sh` sources `versions.sh` which is the common place to set versions and tags. It also provides some helper functions. You can read about all of this in `images/README.md`.

The new `build-images.yml` GitHub workflow handles the interdependencies of the image building and testing. It uses GitHub Container Registry as a temporary storage space for images. After testing, tagged images are pushed to quay.io.

The main advantage of this overhaul is to factor out all the commonalities across building images, which otherwise have grown in different directions over the years. This provides just one unified system that maintainers need to understand. The reason it is in this repo rather than `bioconda-containers` is that we have strong dependencies on the version of bioconda-utils, so ideally changes to bioconda-utils should also test that corresponding base images work correctly.

**NOTE: this does NOT yet replace the bioconda bot image** so bioconda-containers is still needed for that.


daler:
Some notes on the current status of this:

I would like to be able to test the just-built images (in the case that they were rebuilt in a PR) but also test the existing images on quay.io (in the case that only bioconda-utils has been changed).

This requires a clean way of specifying which images to use, but [this is currently complicated](https://bioconda.github.io/developer/dockerfile-inventory.html#details-of-containers-using-in-building). So the next stage here is cleaning up how images are specified throughout the code base. To avoid needing to simultaneously update mulled-build source, the strategy here should be to match what mulled-build looks for and uses.

Specifically in [these lines of mulled_build.py](https://github.com/galaxyproject/galaxy/blob/01c4de53162f4e4ee306ebdd008199a897222dc3/lib/galaxy/tool_util/deps/mulled/mulled_build.py#L61-L72), we have:

- `DEFAULT_BASE_IMAGE` which is the busybox base image
- `DEFAULT_EXTENDED_BASE_IMAGE` which is the debian base image
- `DEST_BASE_IMAGE`, also the busybox base image
- `CONDA_IMAGE` which is the "create-env" image

daler:
I noticed that the `build-env` container is installing a version of conda that is not specified in bioconda-utils requirements. Can't quite figure out why. Despite requirements saying `conda=24.5.*", it's installing `23.7`...and it's taking almost an hour. Going to revert [f311612](https://github.com/bioconda/bioconda-utils/pull/959/commits/f311612d41f530b69fa9220f42e3d6db5c925840) which removed mamba...

daler:
Current status: the version failure is highlighting the fact that we're not testing the very same image that was just built. This is because 1) the upstream quay.io image is hard-coded, and 2) bioconda-utils uses `docker` but the images were built with `podman`.

Trying to get docker to work with the local podman registry has been tricky, at least on a local test instance. Next attempt was to try not using docker at all and switch completely to `podman`, messing around with this in https://github.com/bioconda/bioconda-utils/pull/1045.

daler:
Closing in favor of #1049, which avoids using ghcr.io as a temporary cache and which also does include the bot image.

PR #962, opened 2024-03-12
feat: add support for excluding otherwise-selected recipes

daler:
The nightly rebuild job fails on bioconda-repodata-patches -- which is built by the same workflow immediately before the rebuild job. This allows us to exclude individual recipes from the CLI.

### DIFF
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -301,7 +301,7 @@ def build_recipes(recipe_folder: str, config_path: str, recipes: List[str],
                   record_build_failures: bool = False,
                   skiplist_leafs: bool = False,
                   live_logs: bool = True,
-                  exclude = List[str] = None,
+                  exclude: List[str] = None,
                   ):
     """
     Build one or many bioconda packages.
--- a/bioconda_utils/build.py
+++ b/bioconda_utils/build.py
@@ -358,7 +358,7 @@ def build_recipes(recipe_folder: str, config_path: str, recipes: List[str],
         linter = lint.Linter(config, recipe_folder, lint_exclude)
     else:
         linter = None
-    if not exclude:
+    if exclude:
         if isinstance(exclude, str):
             exclude = [exclude]
         blacklist.global_list.update(exclude)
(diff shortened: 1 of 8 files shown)


bioconda_utils/build.py
```diff
@@ -300,7 +300,9 @@ def build_recipes(recipe_folder: str, config_path: str, recipes: List[str],
                   mulled_conda_image: str = pkg_test.MULLED_CONDA_IMAGE,
                   record_build_failures: bool = False,
                   skiplist_leafs: bool = False,
-                  live_logs: bool = True):
+                  live_logs: bool = True,
+                  exclude = List[str] = None,
```
martin-g, line 304:
```suggestion
                  exclude: List[str] = None,
```

test/test_utils.py
```diff
@@ -156,6 +156,37 @@ def multi_build(request, recipes_fixture, config_fixture):
             ensure_missing(pkg)
 
 
+@pytest.fixture(scope='module', params=PARAMS, ids=IDS)
```
aliciaaevans, line 159: Doesn't this have to get called from a test like this in order to actually test it? https://github.com/bioconda/bioconda-utils/blob/b007d34e6c723f7f9d6fcb5a6f58e072d4618cdf/test/test_utils.py#L228

test/test_utils.py
```diff
@@ -268,6 +312,7 @@ def test_docker_builder_build(recipes_fixture):
                                 build_args='', env={})
     for pkg in pkgs:
         assert os.path.exists(pkg)
+        ensure_missing(pkg)
```
aliciaaevans, line 315: Ah, ok so it was running into old builds hanging around.

bioconda_utils/build.py
```diff
@@ -358,6 +362,10 @@ def build_recipes(recipe_folder: str, config_path: str, recipes: List[str],
     failed = []
 
     dag, name2recipes = graph.build(recipes, config=config_path, blacklist=blacklist)
+    if exclude:
+        for name in exclude:
+            dag.remove_node(name)
```
aliciaaevans, line 367: 👍

aliciaaevans (approved):

PR #965, opened 2024-03-21
feat: add osx-arm64 to platform checks

aliciaaevans:
add osx-arm64 to platforms and only build if additional platform is configured


daler (approved):
Looks good as-is; do you think it is worth thinking about tests at this point, or just wait to see how things look over in `recipes`?

jmarshall:
It's great to see from this and the previous PR that osx-arm64 support is progressing. It would be useful for those looking forward to this functionality if the core cabal could mention this progress (and give a vague idea of what's involved and timelines) in the related conversations on bioconda/bioconda-recipes#33333 or bioconda/bioconda-recipes#37068 or reply to the person who asked about it on gitter.

aliciaaevans:
> Looks good as-is; do you think it is worth thinking about tests at this point, or just wait to see how things look over in `recipes`?

Good thinking. I'll update the ones for the platform check

PR #972, opened 2024-03-27
feat: Update dependencies to latest versions

mbargull:
Update dependencies to their latest versions as of 2024-04-15 (apart from `conda-forge-pinning` and `python`) and adjust code accordingly.

### DIFF
--- a/bioconda_utils/docker_utils.py
+++ b/bioconda_utils/docker_utils.py
@@ -94,10 +94,13 @@ set -eo pipefail
 # will exist in the container but will be empty.  Channels expect at least
 # a linux-64/linux-aarch64 and noarch directory within that directory, so we
 # make sure it exists before adding the channel.
-mkdir -p {self.container_staging}/linux-64
-mkdir -p {self.container_staging}/linux-aarch64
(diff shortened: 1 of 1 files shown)


bioconda_utils/bioconda_utils-requirements.txt
```diff
@@ -2,60 +2,72 @@
...
+pyaml=23.12.*                 # Faster YAML parser (deprecate?)
+networkx=3.2.*                #
+pandas=2.2.*                  #
+libblas=*=*openblas           # Avoid large mkl package (pulled in by pandas)
+boltons=23.*                  #
+jsonschema=4.21.*             # JSON schema verification
+jinja2=3.1.*                  #
+
+anaconda-client=1.12.*        # anaconda_upload
+galaxy-tool-util=23.*         # mulled test and container build
```
martin-g, line 31: Would it be OK to update `galaxy-tool-util` to 24.0.0 ?

mbargull: There are a few deps that got updated in the meantime; I'm looking at their changes right now.

PR #973, opened 2024-03-28
feat: fetch artifacts from GitHub Actions

aliciaaevans:
GitHub requires a token for artifact downloads from their API. (A logged-in user is required for the normal download link shown on the Actions page. This needs to be kept in mind when we update the download comment from the bot.)

I could not find any way to get the workflow run ID or artifacts from the check run object except parsing the URL.

### DIFF
--- a/bioconda_utils/artifacts.py
+++ b/bioconda_utils/artifacts.py
@@ -219,13 +219,16 @@ def get_circleci_artifacts(check_run, platform):
                             yield artifact_url
 
 def parse_gha_build_id(url: str) -> str:
+    # Get workflow run id from URL
     return re.search("runs/(\d+)/", url).group(1)
 
 def get_gha_artifacts(check_run, platform, repo):
     gha_workflow_id = parse_gha_build_id(check_run.details_url)
     if (gha_workflow_id) :
+        # The workflow run is different from the check run
         run = repo.get_workflow_run(int(gha_workflow_id))
         artifacts = run.get_artifacts()
         for artifact in artifacts:
+            # This URL is valid for 1 min and requires a token
             artifact_url = artifact.archive_download_url
             yield artifact_url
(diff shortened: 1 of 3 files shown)


bioconda_utils/artifacts.py
```diff
@@ -110,17 +116,24 @@ def upload_pr_artifacts(config, repo, git_sha, dryrun=False, mulled_upload_targe
...
     logger.info(f"Downloading artifact {url}.")
-    resp = requests.get(url, stream=True, allow_redirects=True)
+    headers = {}
+    if artifact_source == "github-actions":
+        token = os.environ.get("GITHUB_TOKEN")
+        if not token:
+            logger.critical("GITHUB_TOKEN required to download GitHub Actions artifacts")
+            exit(1)
+        headers = {"Authorization": f"token {token}"}
+    resp = requests.get(url, stream=True, allow_redirects=True, headers=headers)
```
daler, line 128: is there a way of guarding against possible leakage of token here?

aliciaaevans: Any suggestions? PyGithub only provides the download URL, it doesn't have a method for downloading the artifact. I can make a request through that (though it's kinda hacky `artifact._requester`) which returns the temporary signed download URL and then download that.

aliciaaevans: Another option could be to do something with https://github.com/actions/download-artifact

It's frustrating that GHA requires some sort of credentials to get artifacts.

daler: Hmm, I feel like the temporary signed download URL would be a more secure option. But IIRC GitHub will obfuscate secrets in logs even if inadvertently leaked (like a malicious PR with `print(token)`), and looking more closely, the token is used in a similar manner elsewhere in the code. In other words, this is not changing the current security stance, so I think it's fine.

daler (approved):

bioconda_utils/artifacts.py
aliciaaevans: Thanks, @daler I think you are right about the secrets, as the docs just say to make sure to declare any that have been modified (e.g. base64 encoded) later on, etc.

For clarity/posterity, the GH API generates and redirects to that temporary signed URL from the download URL I'm using here. So, there's no way around sending the auth, it's just that using PyGitHub is already in use at this point in the code so if it was used there would be fewer places where we have to pass the token in Bioconda code. I think it's ultimately doing the same thing I'm doing here in sending a request to the API, just a bit more buried.

PR #976, opened 2024-04-03
test: run tests on osx-arm64

aliciaaevans:
See if all tests pass on osx-arm and resolve.

### DIFF
--- a/.github/workflows/GithubActionTests.yml
+++ b/.github/workflows/GithubActionTests.yml
@@ -51,7 +51,38 @@ jobs:
           fi
   test-macosx:
     name: OSX tests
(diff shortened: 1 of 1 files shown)


aliciaaevans:
Actually, these might not be passing simply because the test recipes don't have osx-arm64 configured as an additional platform.

milot-mirdita:
If you want something to test, all my MMseqs2 based recipes (`grep -R HAVE_ARM8 recipes`) work (should work :D) on macos arm64.

aliciaaevans:
@milot-mirdita Thanks! For this one, it's just unit tests mock recipes.

PR #985, opened 2024-05-15
chore!: Remove defaults channel dependency

mbargull:
Formally remove usage/dependency on the `defaults` channel.
This has been a ToDo on the backlog for a long time and will result in more consistent and more efficient environment resolutions.


mbargull:
I've tagged this with a `!` to indicate the breaking change and trigger a `3.0` release.
Adjustments for common, docs, recipes will be pushed once we have it released.

PR #986, opened 2024-05-15
chore!: Separate docs-only requirements

mbargull:
As of now, additional requirements for building docs add about 60 packages (+44MB downloaded, 290MB installed) not used by the build env.


mbargull:
I'll adjust the docs repo accordingly once this is released.

PR #938, opened 2023-11-23
feat: Add Python 3.11 to the build matrix

jmarshall:
I have no idea what all is involved in updating this and perhaps doing initial package builds on the **bulk** branch, but this PR provides a starting point for this long-overdue update. Python 3.12 was released almost two months ago, so it is embarrassing that bioconda is still not building packages for the previous year's release, Python 3.11.

This is tracked as bioconda/bioconda-recipes#37805. That links to conversation on #878 which indicates that bioconda has been waiting for conda-forge to include 3.11. It appears that [conda-forge's migration was completed in September](https://conda-forge.org/docs/user/announcements.html#python-3-12-migration-and-python-3-11-by-default).

PR #988, opened 2024-05-21
chore: Bump bamtools to 2.5.2

martin-g:
2.5.2 is the only version at the moment that provides linux-aarch64 package: https://anaconda.org/bioconda/bamtools/files

I faced the following error
```
WARNING: failed to get package records, retrying.  exception was: Unsatisfiable dependencies for platform linux-aarch64: {MatchSpec("bamtools=2.5.1")}
Encountered problems while solving:
  - nothing provides requested bamtools 2.5.1.*
```
while adding linux-aarch64 to `scalpel`
  (https://github.com/bioconda/bioconda-recipes/pull/47964)

Full logs with the error:
- https://circleci.com/api/v1.1/project/github/bioconda/bioconda-recipes/186948/output/107/0?file=true&allocation-id=664c88870a2ca54ac5eb924f-0-build%2FABCDEFGH

Successful build after adding custom conda_build_config.yaml for scalpel:
- https://app.circleci.com/pipelines/github/bioconda/bioconda-recipes/116194/workflows/00d6dc8f-1d05-48a5-b15f-1ae3bff71433/jobs/186951

PR #991, opened 2024-05-29
feat: Update pinning 2024-05

mbargull:
Amongst others, this includes:
- Update default minimum system targets to
    - `glibc 2.17`/CentOS 7 ( \*needs recipes to include new `- {{ stdlib("c") }}` dep in `requirements/build`)
    - macOS 10.13
- Python 3.11 and 3.12

closes gh-938, gh-988

### DIFF
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -21,7 +21,7 @@ MACOSX_DEPLOYMENT_TARGET:      # [osx]
   - 10.13                      # [osx and x86_64]
 
 bamtools:
-  - "2.5.1"
+  - "2.5.2"
 
 # NOTE: Workaround https://github.com/conda/conda-build/issues/3974 we slightly alter the values
 #       from conda-forge-pinnings here (inserting '.*' or ' ' which should be ignored later on).
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -20,15 +20,6 @@ MACOSX_DEPLOYMENT_TARGET:      # [osx]
   - 11.0                       # [osx and arm64]
   - 10.13                      # [osx and x86_64]
 
-pin_run_as_build:
-  htslib:
-    max_pin: x.x
-  bamtools:
-    max_pin: x.x.x
-  r-base:
-    max_pin: x.x
-    min_pin: x.x
-
 bamtools:
   - "2.5.1"
 
--- a/bioconda_utils/bioconda_utils-conda_build_config.yaml
+++ b/bioconda_utils/bioconda_utils-conda_build_config.yaml
@@ -28,13 +28,6 @@ bamtools:
 r_base:
   - "4.3.*"
 
-# Note: this is a (hopefully) temporary fix for the r-base 4.3 migration. That package was built 
-# with libxml2 2.11, but that migration on conda-forge isn't yet complete. Consequently conda-forge-pinning
-# still has libxml2 2.10. That means that there are a number of R and Bioconductor packages we can't currently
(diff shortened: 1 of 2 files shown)


bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -9,50 +9,46 @@
...
-
-pin_run_as_build:
-  htslib:
-    max_pin: x.x
-  bamtools:
-    max_pin: x.x.x
-  r-base:
-    max_pin: x.x
-    min_pin: x.x
+  - 10.13                      # [osx and x86_64]
```
mbargull, line 21: In the future we should also be able to drop `MACOSX_DEPLOYMENT_TARGET` and rely on the more uniform definition via `{{ stdlib("c") }}`.
Until we update our recipes to make use of it, we keep this defined here.

aliciaaevans:
Here's the conda-forge announcement for information of anyone wondering about the stdlib stuff: https://conda-forge.org/news/2024/03/24/stdlib-migration/

bioconda_utils/bioconda_utils-conda_build_config.yaml
aliciaaevans: So, the existing recipes can still work as-is, right? There is just the added benefit to using the new stdlib function?

bioconda_utils/bioconda_utils-conda_build_config.yaml
```diff
@@ -9,50 +9,46 @@
...
 
 bamtools:
-  - "2.5.1"
+  - "2.5.2"
```
martin-g, line 24: I've created a PR exactly for this few days ago - https://github.com/bioconda/bioconda-utils/pull/988

aliciaaevans:
@mbargull is this one dependent on https://github.com/bioconda/bioconda-utils/pull/992?

bioconda_utils/bioconda_utils-conda_build_config.yaml
mbargull: Exactly.
Depending on the decisions being made in `conda-forge`, we may want to add the `stdlib("c")` parts though sooner (but not needed for the next bulk rebuild) to ensure we get the right `sysroot_linux*` versions later
.
As is, the choice of the sysroot on Linux is done very messily by weighing down unwanted (newer) versions with https://github.com/conda-forge/linux-sysroot-feedstock/blob/2b16024e8907a07dc9430c3b4a18884bd679b65d/recipe/meta.yaml#L138-L140 but in the future those hacks may be removed.

mbargull:
Sorry, wanted to be available Friday but got knocked out with headaches and stuff :/ .

> @mbargull is this one dependent on #992?

Strictly speaking, no.
But there were some improvements in those updates which should make things work more smoothly going forward.

Over the weekend, I smoothed out some more rough edges (see https://github.com/bioconda/bioconda-common/pull/54) and rebuilt our container images, so this should be ready to go.

One caveat: I mentioned before that we are ready to put out `.conda` packages -- and we are in principle and I already uploaded one such package: https://anaconda.org/bioconda/bioconda-utils/files/modal/info/665ce27336ee7a8e2befd3c1 .
But it turns out those packages aren't added to the CDN-backed channel for `bioconda`, so I backed out of that change until https://github.com/conda/infrastructure/issues/950 is resolved.

mbargull:
>     * Update default minimum system targets to
>       
>       * `glibc 2.17`/CentOS 7

to clarify: this only applies for recipes for which we add the `- {{ stdlib("c") }}` part.

bioconda_utils/bioconda_utils-conda_build_config.yaml
aliciaaevans: Thanks, we'll just roll this into the new update.

aliciaaevans (approved):

PR #994, opened 2024-06-01
fix: Pass pkg_format to container-based build

mbargull:
This way we should be able to switch to `pkg_format: 2` (i.e., build `.conda` packages) via config from `bioconda-common`.

### DIFF
--- a/bioconda_utils/docker_utils.py
+++ b/bioconda_utils/docker_utils.py
@@ -105,6 +105,9 @@ conda config --add channels file://{self.container_staging} 2> >(
     grep -vF "Warning: 'file://{self.container_staging}' already in 'channels' list, moving to the top" >&2
 )
 
+# Pass on conda_pkg_format ("2" for .conda instead of .tar.bz2) from host's conda-build config.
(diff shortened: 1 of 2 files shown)


bioconda_utils/docker_utils.py
```diff
@@ -155,18 +158,6 @@ class DockerBuildError(Exception):
...
-def get_host_conda_bld():
-    """
-    Identifies the conda-bld directory on the host.
-
-    Assumes that conda-build is installed.
-    """
-    # v0.16.2: this used to have a side effect, calling conda build purge
-    # hopefully, it's not actually needed.
-    build_conf = utils.load_conda_build_config()
-    return build_conf.build_folder
```
mbargull, line 167: `.build_folder` is actually the `.../conda-bld/pkg_....` path if the `Config` would be initialized for some package `pkg` -- it just happens to be `.../conda-bld/` since the `pkg_...` part is empty for a `Config` not created for a certain package/recipe.
The new code has the correct `config.croot` (which is always only the `.../conda-bld` part.

PR #999, opened 2024-06-06
fix: add missing Arm platforms to skip existing builds

aliciaaevans:
When running bulk for linux-aarch64 on CircleCI, all recipes started building, even ones that had not been bumped. The `FILTER: not building recipe X because the same number of builds are in channel(s) and it is not forced.` message was not present in the log. This is because one place in bioconda-utils was missed when adding`linux-aarch64`. 

Bioconda-utils gets a dataframe of packages from the bioconda and conda-forge channels and then compares with the list of recipes requested to be built to see if a build already exists for that subdir, version, build number, and build string. Only `linux-64`, `osx-64`, and `noarch` were being considered. (The platform here gets converted to a subdir string which has the `-64` suffix.)

After this is merged and released, I will re-enable bulk for CircleCI.

### DIFF
--- a/bioconda_utils/utils.py
+++ b/bioconda_utils/utils.py
@@ -1400,7 +1400,7 @@ class RepoData:
     #: Columns available in internal dataframe
     columns = _load_columns + ['channel', 'subdir', 'platform']
     #: Platforms loaded
-    platforms = ['linux', 'osx', 'noarch']
+    platforms = ['linux', 'linux-aarch64', 'osx', 'osx-arm64', 'noarch']
     # config object
     config = None
 


aliciaaevans:
Setting this to draft while I double-check if this will have unwanted side-effects.

aliciaaevans:
I think the other places this affects actually benefit or are unaffected by this change. This is adding rows to the dataframe, not changing the existing linux and osx ones.

aliciaaevans:
> I don't understand yet how `linux` gets turned into `linux-64` but I guess `linux-aarch64` doesn't. But it sounds like you confirmed it's only adding rows (and specifically only the ARM rows), in which case I'm not worried!

Here for information:
https://github.com/bioconda/bioconda-utils/blob/ed8607c90fb0814c738a156bcb98630f996a1e05/bioconda_utils/utils.py#L1532-L1545

PR #1001, opened 2024-06-17
fix: token for CircleCI artifacts

aliciaaevans:
Although the endpoints to list workflow jobs and artifacts used to be accessible to anyone, now there is a 401 response code.

Previously we worked around this by changing the API version back to v1.1, but now it's happening again for both versions. This apparently happens often. https://discuss.circleci.com/t/api-token-needed-to-fetch-artifacts/50881/3 So, it makes sense to just use a token. I've configured this in CircleCI already.

(Another PR will be coming soon for the comment responder to fetch the artifacts.)


martin-g (approved):
LGTM!

martin-g (approved):

PR #1022, opened 2024-12-01
Pylint alerts corrections as part of an intervention experiment 298

evidencebp:
Makes the interventions describe in [intervention issue](https://github.com/alejandroautalan/pygubu/issues/298).
The experiment is described [here](https://github.com/evidencebp/pylint-intervention/).

Each intervention was done in a dedicated commit with a message explaining it.

### DIFF
--- a/bioconda_utils/artifacts.py
+++ b/bioconda_utils/artifacts.py
@@ -234,7 +234,7 @@ def parse_gha_build_id(url: str) -> str:
 
 def get_gha_artifacts(check_run, platform, repo):
     gha_workflow_id = parse_gha_build_id(check_run.details_url)
-    if (gha_workflow_id) :
+    if gha_workflow_id:
         # The workflow run is different from the check run
         run = repo.get_workflow_run(int(gha_workflow_id))
         artifacts = run.get_artifacts()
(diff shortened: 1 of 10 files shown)


jmarshall:
* Please use slashes rather than backslashes to delimit paths in your commit messages.

* _versioneer.py_ is external code copied into bioconda-utils. It is a mistake to alter this upstream code locally.

* IMHO these too-many-branches and too-many-statements changes in fact make this code **less** understandable, by moving code further away from the code it is related to and by introducing badly-named functions that implement meaningless abstractions. So my opinion would be not to apply these particular changes.

evidencebp:
@jmarshall , I'll exclude versioneer.py

As for the too-many interventions, in this PR the code is not very-much-too-many.
Your claim about locality is reasonable and that is what makes these cases interesting for the experiment.
If you find the intervention good enough, merging them will allow us to investigate the change in tendency to bugs, ease of modifications, etc.
Are you OK with that?

jmarshall:
It's not my decision. If it was, I would have just closed WONTFIX your initial issue. Your experiment may be interesting to you, but it's mostly a waste of time for the projects you're experimenting on.

PR #1023, opened 2024-12-13
feat: report resource usage before and after builds

aliciaaevans:
To diagnose out of disk and out of memory errors, log free disk space and available memory before and after each recipe is built and each mulled test.


aliciaaevans:
Going to release this but only update bioconda-common to point to this version for the `bulk` branch.

daler (approved):

PR #1025, opened 2024-12-13
fix: prune stopped containers after build

aliciaaevans:


daler:
Initial failure looks to be temporary 502 from anaconda; added additional `docker ps -a` to be able to track containers and confirm they get removed

PR #1027, opened 2024-12-22
feat: Bioconductor skeleton updates

daler:
- When running `bioconda-utils bioconductor-skeleton` on Bioconductor data packages, I was getting self-signed SSL errors (@bgruening is this expected?). This change allow such errors to be ignored, returning None for the url.
- In this round of Bioconductor 3.20 updates, it seems like a lot of packages are now missing `zlib`. I still haven't pinpointed the reason, but I'm guessing it has something to do with the updates to pinnings. So for BioC packages that are detected to be compiled, this adds `zlib` to host dependencies.

### DIFF
--- a/bioconda_utils/bioconductor_skeleton.py
+++ b/bioconda_utils/bioconductor_skeleton.py
@@ -943,6 +943,11 @@ class BioCProjectPage(object):
             additional_host_deps.append('libblas')
             additional_host_deps.append('liblapack')
 
+            # During the BioC 3.20 builds, which also corresponded to updates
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks @daler! I don't think this is expected, strange.

bgruening (approved):

martin-g (approved):

aliciaaevans:
@daler for SSL errors, you may be hitting this: https://stackoverflow.com/questions/42098126/mac-osx-python-ssl-sslerror-ssl-certificate-verify-failed-certificate-verify/61142526#61142526

PR #1031, opened 2025-01-05
feat: improve build failure table in wiki

daler:
Include a column for the "reason" field of build failures, which hopefully helps point people in the right direction for fixing. After 80 chars, will dump the rest in a `<details>` tag to avoid table overflow.

Also list multiple rows (one per platform) instead of comma-separated strings, which do not work well in minimally-styled github wiki table.

Example of new table:

Screenshot 2025-01-05 at 5 04 54 PM


aliciaaevans (approved):
I like it. I'll be better about filling out the reason from now on. It will especially help with prioritizing similar errors to fix across several recipes.

PR #1038, opened 2025-03-19
fix: include pypi.org/packages in PyPi hoster_pattern

tedil:
Adding `pypi.org/packages` to the hoster_pattern for PyPi (which is a redirection to `files.pythonhosted.org/packages`), which I believe is the URL that e.g. grayskull uses.

### DIFF
--- a/bioconda_utils/hosters.py
+++ b/bioconda_utils/hosters.py
@@ -683,6 +683,7 @@ class PyPi(JSONHoster):
     hoster_pattern = (r"(?P<hoster>"
                       r"files.pythonhosted.org/packages|"
                       r"pypi.python.org/packages|"
+                      r"pypi.org/packages|"
                       r"pypi.io/packages)")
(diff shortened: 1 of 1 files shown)


bgruening (approved):

bgruening:
Thanks!

aliciaaevans:
This has been released and the autobump picked up a whole bunch of new recipes.

dlaehnemann:
Is that a good or a bad thing? And does this need any action, like quickly reviewing a bunch of update pull requests, before they get auto-merged?

aliciaaevans:
> Is that a good or a bad thing? And does this need any action, like quickly reviewing a bunch of update pull requests, before they get auto-merged?

Good thing. I just wanted to mention that it has now been implemented. There's no auto-merge, so no particular rush to review, but it would be nice to merge at least the easy ones to reduce clutter.

PR #1040, opened 2025-03-27
ci: update upload-artifact version in changevisibility since v3 was removed

aliciaaevans:
I believe this should fix https://github.com/bioconda/bioconda-recipes/issues/54470

### DIFF
--- a/.github/workflows/changevisibility.yml
+++ b/.github/workflows/changevisibility.yml
@@ -20,7 +20,7 @@ jobs:
           QUAY_OAUTH_TOKEN: ${{ secrets.QUAY_BIOCONTAINERS_TOKEN }}
 
       - name: Upload logs
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: logs
           path: biocontainers-*.txt


.github/workflows/changevisibility.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           QUAY_OAUTH_TOKEN: ${{ secrets.QUAY_BIOCONTAINERS_TOKEN }}
 
       - name: Upload logs
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
bgruening, line 23: maybe we can just remove the version? We trust those actions, do we?

Thanks a lot @aliciaaevans!

aliciaaevans: If there was a "latest" tag that could make sense, but I can't seem to find documentation on what we'd get if we remove the ref and there doesn't seem to be a "latest". I wouldn't want to be using an unreleased default branch head. I'd prefer to leave it with this major version tag. It looks like the other `uses`  line in this file was updated a while ago, but this one just got missed.

martin-g (approved):

.github/workflows/changevisibility.yml
daler: Agreed, makes sense to be more explicit if the behavior is not currently well-defined.

daler (approved):

PR #1041, opened 2025-04-30
fix: keep existing patches during bioconductor updates

aliciaaevans:
Currently when doing a bulk Bioconductor upgrade, the patches section has been lost. That leads to errors and manually digging through commits to see what was lost. If we keep the patches, even if they need updating, at least it will be obvious.

Also set `sort_keys=False` so that top-level yaml items are written in the traditional order instead of being alphabetized. (Diffs will be annoying for the next upgrade but, I think future readability is worth it.)

I confirmed that the `additional-platforms` section in `extra` _is_ getting kept through upgrades, so no changes there.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build:
     name: Build image
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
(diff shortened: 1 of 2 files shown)


bgruening (approved):
Will be interesting to see how many of the old patches will still apply.

The PR LGTM!

PR #1042, opened 2025-04-30
ci: update image workflow to use ubuntu-24.04, remove deprecated set-output

aliciaaevans:
ubuntu-20.04 is no longer supported

set-output is deprecated: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -12,7 +12,7 @@ jobs:
       release_created: ${{ steps.release.outputs.release_created }}
       tag_name: ${{ steps.release.outputs.tag_name }}
     steps:
-      - uses: GoogleCloudPlatform/release-please-action@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/release-please.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           #tag=${{ github.event.release && github.event.release.tag_name || github.sha }}
           tag=${{ needs.release_please.outputs.tag_name }}
-          printf %s "::set-output name=tag::${tag#v}"
+          printf %s "tag=${tag#v}" >> $GITHUB_OUTPUT
```
bgruening, line 34: Oh that is interesting!

bgruening (approved):
Thanks @aliciaaevans

PR #1044, opened 2025-05-01
feat: allow and require {{ stdlib("c") }} in recipes

aliciaaevans:
- add stdlib function to `JINJA_VARS` to prevent the `version_constraints_missing_whitespace` lint error
- add new lint to require stdlib function if compilers are used
- update tests for the new lint (and to avoid tripping it when testing other lints)
- Add `{{ stdlib("c") }}` to bioconductor-skeleton

See https://github.com/bioconda/bioconda-recipes/issues/51185

### DIFF
--- a/bioconda_utils/lint/check_build_help.py
+++ b/bioconda_utils/lint/check_build_help.py
@@ -62,6 +62,24 @@ class compilers_must_be_in_build(LintCheck):
                     if "run" in location or "host" in location:
                         self.message(section=location)
 
+class compiler_needs_stdlib_c(LintCheck):
+    """The recipe requests a compiler in the build section, but does not have stdlib.
+
+    Please add the ``{{ stdlib('c') }}`` line to the
+    ``requirements: build:`` section.
+    """
+
+    def check_deps(self, deps):
+        compiler = False
+        stdlib = False
+        for dep, locations in deps.items():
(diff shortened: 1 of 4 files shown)


bioconda_utils/lint/check_build_help.py
```diff
@@ -62,6 +62,24 @@ def check_deps(self, deps):
...
                         self.message(section=location)
 
+class compiler_needs_stdlib_c(LintCheck):
+    """The recipe requests a compiler in the build section, but does not have stdlib.
+
+    Please add the ``{{ stdlib('c') }}`` line to the
+    ``requirements: build:`` section.
+    """
+
+    def check_deps(self, deps):
```
bgruening, line 72: Do we need a check for having stdlib but no compiler?

bgruening (commented):
Just one small question.

bioconda_utils/lint/check_build_help.py
aliciaaevans: Good question. It doesn't seem like [conda-forge](https://github.com/conda-forge/conda-smithy/blob/main/conda_smithy/linter/lints.py) has that check. Although maybe someone might get confused and think they could use it in place of a compiler.

bgruening (approved):

PR #1046, opened 2025-05-21
feat: Simplify unify containers, but with artifacts [WIP]

daler:
https://github.com/bioconda/bioconda-utils/pull/959 uses ghcr.io as a cache mechanism. However, images stored there are not straightforward to clean up. Artifacts, on the other hand, expire after 90 days. This PR is an experiment to see if it will simplify things further to use artifacts rather than ghcr.io.


daler:
In good shape; closing in favor of the cleaned-up #1049.

PR #1054, opened 2025-07-17
feat: Update dependencies to latest versions

mencian:


aliciaaevans (commented):
This looks good to me, but can you also update aiohttp to 3.12.14 at least? There was a security patch.

aliciaaevans (approved):
Thanks!

PR #1057, opened 2025-07-31
fix: also find tests under outputs, ensure all outputs have tests

dlaehnemann:
This pull request is motivated by trying to add a subpackage to an existing package and trying to get the syntax right:
https://github.com/bioconda/bioconda-recipes/pull/58141

I think I have to add both the main and the subpackage under an `outputs:` section, and the linter then complains that it doesn't find any global `test:` section. So this PR makes the linter go through each subpackage if an `outputs:` section exists, and checks whether all of them have a valid `test:` section specified.


dlaehnemann:
I can't ask for Reviewers on issues in this org, it seems. And I'm not sure whom to ping here for a review. So I'll just ping the top two contributors, @daler and @epruesse . Fell free to pass this on or let me know whom else to ping (or where).

PR #1063, opened 2025-08-13
Fix wrong jinja function definitions

jpfeuffer:


jpfeuffer:
Thank you for the quick answer. Interestingly it works with pin_subpackage although not registered here:
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/openms-meta/meta.yaml#L135

I am not sure if it just ignores this line in that case. Therefore I am a bit hesistant of adding more functions before knowing exactly what the code in recipe.py actually does. Maybe let's start with fixing the existing one first.

PR #1086, opened 2026-03-30
feat!: adjust linting to new outputs structure

dlaehnemann:
This pull request is basically `git cherry-picking` all the previous work in pull request #1073 , as that one's diff has gotten too confusing with a failed `rebase --onto master` attempt (I realized too late, that I am working on a fork, here...). In addition, this pull request is a bit more generic, trying to get a consistent linting and comprehensive linting tests for recipes with multiple outputs in general. The aim is to work towards supporting the new `rattler-build` driven `recipe.yaml` format that has been accepted as Conda Enhancement Proposals (CEPs) 13 and 14, by implementing [the `outputs:` specification](https://conda.org/learn/ceps/cep-0014/#outputs-section) in the linting over here.

So, this is a fresh start after the overhaul of the linting test suite setup in #1085 has been merged. And to also link more previous work, this is the follow-up to pull request #1057, which did a first step towards better linting of multiple outputs.

Please note, that this will require updating all of the 7 recipes with multiple outputs so that they pass the new lints, also see here:
https://github.com/bioconda/bioconda-utils/pull/1057#issuecomment-3159446608
The plan is to open a separate pull request for each of them on `bioconda-recipes`.


epruesse (approved):
original PR already approved

epruesse:
@dlaehnemann The PRs will be squashed into one commit to the main branch anyway. If you think this should stay as multiple pieces, you'd have to do separate PRs or we'd have to somehow to the merge manually. 

Also, I'm not sure about the classification as "fix". It might be a "feat!" (new feature with breaking change requiring major version bump). Thoughts?

dlaehnemann:
Basically, this has grown in scope since Johannes approved the original PR. and I did the overhaul of the linting test suite, which has been merged. So I cherry-picked all the relevant stuff here, so that we would have a clean diff in the pull request, mainly for the review of this pull request.
Without the cherry-picking, the resolution of the merge conflicts would have been harder and I also messed up the rebase, because I forgot that I was working on a fork...

But this is not yet done. I'm still working through all the cases with the new `outputs:` policy and trying to get the linting right (including dedicated test cases).

dlaehnemann:
OK, I think this is good to be reviewed and possibly merged.

This is another step towards proper linting of recipes with multiple outputs, also working towards supporting the new `recipe.yaml` format in that regard:
https://rattler-build.prefix.dev/v0.65.0/reference/multi_output/

However, I have restricted it to linting the stuff that [the existing multi-output recipes in bioconda](https://github.com/search?q=repo%3Abioconda%2Fbioconda-recipes+%22outputs%3A%22&type=code) require. This means that the linting currently does not check all of the [inheritance behavior of the `v1` recipe format](https://rattler-build.prefix.dev/v0.58.4/reference/multi_output/#inheritance-behavior).
For example, the build number is only checked (and possibly fixed) on the top-level of the recipe, not per-output, as all the existing recipes just have a single build number and especially updating the automatic fixing code for the build numbers would be more involved.
Also, it might make sense to eventually adopt the lint (and build) infrastructure of `conda-forge` at some point, which is already implementing the new recipe format in a backwards compatible manner. So duplicating all that work over here might be a waste of time...

Finally, the failing `ty` qc check on the typing seems unrelated:
```
error[unresolved-attribute]: Attribute `keys` is not defined on `bool` in union `dict[Any, Any] | bool`
   --> bioconda_utils/circleci.py:262:48
    |
262 |             f"success={x['success']} {':'.join(x['urls'].keys())}" for x in self.parsed
    |                                                ^^^^^^^^^^^^^^
    |

Found 1 diagnostic
```
And I don't understand why the `urls` entry in `self.parsed` should possibly be `bool` instead of a `dict`, and where this implicit typing would happen. [The respective code filling that key ](https://github.com/bioconda/bioconda-utils/blob/c42a28e687b36b22bdd11a22a4fc2bef00b97a72/bioconda_utils/circleci.py#L250-L252) should always return a `dict`, even if it is empty because there's no match (and never a `bool`):
```
            urls = {
                key: url for url, key in re.findall(r"<(http[^|>]+)\|([^>]+)>", text)
            }
```

@bioconda/core , anybody up for a review?

johanneskoester:
> OK, I think this is good to be reviewed and possibly merged.
> 
> This is another step towards proper linting of recipes with multiple outputs, also working towards supporting the new `recipe.yaml` format in that regard: https://rattler-build.prefix.dev/v0.65.0/reference/multi_output/
> 
> However, I have restricted it to linting the stuff that [the existing multi-output recipes in bioconda](https://github.com/search?q=repo%3Abioconda%2Fbioconda-recipes+%22outputs%3A%22&type=code) require. This means that the linting currently does not check all of the [inheritance behavior of the `v1` recipe format](https://rattler-build.prefix.dev/v0.58.4/reference/multi_output/#inheritance-behavior). For example, the build number is only checked (and possibly fixed) on the top-level of the recipe, not per-output, as all the existing recipes just have a single build number and especially updating the automatic fixing code for the build numbers would be more involved. Also, it might make sense to eventually adopt the lint (and build) infrastructure of `conda-forge` at some point, which is already implementing the new recipe format in a backwards compatible manner. So duplicating all that work over here might be a waste of time...
> 
> Finally, the failing `ty` qc check on the typing seems unrelated:
> 
> ```
> error[unresolved-attribute]: Attribute `keys` is not defined on `bool` in union `dict[Any, Any] | bool`
>    --> bioconda_utils/circleci.py:262:48
>     |
> 262 |             f"success={x['success']} {':'.join(x['urls'].keys())}" for x in self.parsed
>     |                                                ^^^^^^^^^^^^^^
>     |
> 
> Found 1 diagnostic
> ```
> 
> And I don't understand why the `urls` entry in `self.parsed` should possibly be `bool` instead of a `dict`, and where this implicit typing would happen. [The respective code filling that key ](https://github.com/bioconda/bioconda-utils/blob/c42a28e687b36b22bdd11a22a4fc2bef00b97a72/bioconda_utils/circleci.py#L250-L252) should always return a `dict`, even if it is empty because there's no match (and never a `bool`):
> 
> ```
>             urls = {
>                 key: url for url, key in re.findall(r"<(http[^|>]+)\|([^>]+)>", text)
>             }
> ```
> 
> @bioconda/core , anybody up for a review?

ty is not smart enough here, but the code is also quite old-fashioned I would say. Here is a PR that fixes this issue by using a dataclass instead: https://github.com/bioconda/bioconda-utils/pull/1114

bioconda_utils/lint/check_repo.py
```diff
@@ -63,11 +78,24 @@ class build_number_needs_reset(LintCheck):
...
+            for i in range(len(outputs)):
+                version = recipe.get_inherited_value(i, "package/version")
+                build_number = int(recipe.get_inherited_value(i, "build/number"))
+                bldnos = utils.RepoData().get_package_data(
+                    key="build_number",
+                    name=recipe.get(f"outputs/{i}/name"),
+                    version=version,
+                )
+                if not bldnos and build_number > 0:
+                    self.message(f"outputs/{i}/build/number", data=0)
```
johanneskoester: these if/else constructs lead to code duplication.
I would say there should be a common base class like `OutputAwareLintCheck` that contains the output iteration logic and calls an abstract helper method that is implemented in the subclasses which contains the actual check logic.

johanneskoester (changes requested):

bioconda_utils/lint/check_repo.py
dlaehnemann: Actually, I think I should roll this back to only handling the `build/number` section in the main / top-level recipe section. A build number output does not really make sense, unless the CI has a mechanism to selectively build only parts of a recipe. So the easiest fix for now would be to undo the respective lint changes for `build/number` here in `check_repo.py` and instead amend the linting message to notify users, that only top-level build number specification is currently supported. What do you think?

johanneskoester: I agree!

johanneskoester (approved):

Issue #1095, opened 2026-04-08
Hash discrepancy with stdlib('c') use


mbhall88:
I've run into a strange issue in https://github.com/bioconda/bioconda-recipes/pull/64176 where adding `stdlib('c')` to the build section causes the package build to fail due to a difference in the hashes, as seen in this section of a failing log

```
2026-04-07T14:17:57.0723124Z 14:17:44 [32mBIOCONDA INFO[0m (OUT) [0m
2026-04-07T14:17:57.0723558Z 14:17:44 [32mBIOCONDA INFO[0m (OUT) Resource usage statistics from testing nohuman:[0m
2026-04-07T14:17:57.0724032Z 14:17:44 [32mBIOCONDA INFO[0m (OUT)    Process count: 2[0m
2026-04-07T14:17:57.0724490Z 14:17:44 [32mBIOCONDA INFO[0m (OUT)    CPU time: Sys=-, User=0:00:00.0[0m
2026-04-07T14:17:57.0724929Z 14:17:44 [32mBIOCONDA INFO[0m (OUT)    Memory: 14.1M[0m
2026-04-07T14:17:57.0725300Z 14:17:44 [32mBIOCONDA INFO[0m (OUT)    Disk usage: 16B[0m
2026-04-07T14:17:57.0725703Z 14:17:44 [32mBIOCONDA INFO[0m (OUT)    Time elapsed: 0:00:02.0[0m
2026-04-07T14:17:57.0726078Z 14:17:44 [32mBIOCONDA INFO[0m (OUT) [0m
2026-04-07T14:17:57.0726367Z 14:17:44 [32mBIOCONDA INFO[0m (OUT) [0m
2026-04-07T14:17:57.0726895Z 14:17:44 [32mBIOCONDA INFO[0m (OUT) TEST END: /opt/conda/conda-bld/linux-64/nohuman-0.5.1-h4bf21ff_0.conda[0m
2026-04-07T14:17:57.0728057Z 14:17:44 [32mBIOCONDA INFO[0m (OUT) Renaming work directory '/opt/conda/conda-bld/nohuman_1775571179039/work' to '/opt/conda/conda-bld/nohuman_1775571179039/work_moved_nohuman-0.5.1-hee99bd0_0_linux-64_main_build_loop'[0m
[...]

PR #1096, opened 2026-04-08
fix(utils): patch conda-build to include __glibc in hashes when using stdlib('c')

mbhall88:
This relates to #1095. Is there a way I can test this out on the package that caused me to identify this issue in https://github.com/bioconda/bioconda-recipes/pull/64176?

**Full disclosure**: this was done in consultation with Gemini. There is very little chance I would have gotten to the bottom of this on my own in my lifetime. But please, someone with more intimate knowledge of this repo tell me where we have gone wrong.

# Fix hash discrepancy for `stdlib('c')` on Linux

## Summary
This PR fixes a discrepancy in how package hashes are calculated when a recipe uses the new `{{ stdlib('c') }}` syntax. Before this fix, `bioconda-utils` would predict a package hash that didn't match the one generated by `conda-build` during the actual build process on Linux, leading to CI failures.

## Root Cause
`bioconda-utils` frequently calls `conda-build` APIs (like `api.get_output_file_paths`) with `bypass_env_check=True` to avoid the overhead of a full environment solve when simply predicting package names or checking for existing builds.

On Linux, `{{ stdlib('c') }}` typically resolves to a `sysroot` package (e.g., `sysroot_linux-64`). This package contains `run_exports` that pin the `__glibc` virtual package. 
- During a **real build**, `conda-build` performs a full solve, discovers the `__glibc` pin, and includes `__glibc` in the package hash.
- In **bypass mode**, `conda-build` skips the solve and therefore never "sees" the `__glibc` pin. Consequently, it omits `__glibc` from the hash.

This resulted in a mismatch where `bioconda-utils` expected one hash (e.g., `h4bf21ff`) but the actual built package had another (e.g., `hee99bd0`).

## Implementation
The fix involves monkeypatching `conda_build.metadata.find_used_variables_in_text` within `bioconda_utils/__init__.py`. 

The patched version explicitly adds `__glibc` to the set of "used variables" whenever the `{{ stdlib('c') }}` syntax is detected in the recipe text and `__glibc` is present in the variant. This forces `conda-build` to include `__glibc` in its hash calculation even when the environment solve is bypassed. Checking for `__glibc` in the variant before performing the regex search ensures minimal performance impact.

## Verification
I verified the fix using a reproduction script that rendered a test recipe in both bypass and finalized modes:
- **Before Fix:** Hash in bypass mode (`h41f06ac`) != Hash in finalized mode (`h1f94ec8`).
- **After Fix:** Hash in bypass mode (`hd8ae301`) matches the logic including `__glibc`, ensuring consistency with real builds.

### DIFF
--- a/bioconda_utils/__init__.py
+++ b/bioconda_utils/__init__.py
@@ -48,11 +48,10 @@ _old_find_used_variables_in_text = conda_build.metadata.find_used_variables_in_t
 
 def _patched_find_used_variables_in_text(variant, recipe_text, selectors_only=False):
     used = _old_find_used_variables_in_text(variant, recipe_text, selectors_only)
-    if not selectors_only:
+    if not selectors_only and "__glibc" in variant:
         # Match stdlib('c')
         if re.search(r"\{\{\s*stdlib\(\s*[\'\"]c[\'\"]\s*\)", recipe_text):
-            if "__glibc" in variant:
-                used.add("__glibc")
+            used.add("__glibc")
     return used
 
 
--- a/bioconda_utils/__init__.py
+++ b/bioconda_utils/__init__.py
@@ -33,7 +33,30 @@ Bioconda Utilities Package
    utils
 """
 
+import re
+
+import conda_build.metadata
+
 from ._version import get_versions
+
+# Monkeypatch conda_build.metadata.find_used_variables_in_text to include __glibc
+# in the hash when stdlib('c') is used on Linux. This ensures that hashes
+# calculated with bypass_env_check=True (as done in built_package_paths)
+# match those from real builds.
+_old_find_used_variables_in_text = conda_build.metadata.find_used_variables_in_text
+
+
+def _patched_find_used_variables_in_text(variant, recipe_text, selectors_only=False):
+    used = _old_find_used_variables_in_text(variant, recipe_text, selectors_only)
+    if not selectors_only:
+        # Match stdlib('c')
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks @mbhall88. Linting is failing.

bgruening (approved):

epruesse:
I'm not sure whether the location for this monkey patching is good. Yes, everything _currently_ imports `bioconda_utils.utils` I believe, but that's not guaranteed. It might be better to do this in `__init__.py`? Just to be sure it's done always and exactly once?

epruesse:
@mbhall88 I can do a fresh PR, or commit this and do another PR to make it more clean. Easier for me with multiple files. Preferences?

bioconda_utils/utils.py
```diff
@@ -55,13 +55,33 @@
...
+# match those from real builds.
+_old_find_used_variables_in_text = conda_build.metadata.find_used_variables_in_text
+
+
+def _patched_find_used_variables_in_text(variant, recipe_text, selectors_only=False):
+    used = _old_find_used_variables_in_text(variant, recipe_text, selectors_only)
+    if not selectors_only:
+        # Match stdlib('c')
+        if re.search(r"\{\{\s*stdlib\(\s*[\'\"]c[\'\"]\s*\)", recipe_text):
+            if "__glibc" in variant:
```
epruesse: This bit isn't explained in the PR header, also could be checked at top, before doing the regexp search.

mbhall88: Done, and I have updated the PR description accordingly.

PR #1103, opened 2026-04-27
feat: add support for local (file://) channels

dpryan79:
I suspect I'm the only one who does this, but perhaps someone else will find this useful.

Sometimes it's useful to have a local channel of packages defined in a config.yml, for example for packages that can't be publicly distributed. If you want to mix that with bioconda-utils you currently get an error when it tries to load the repodata.json, since it always looks for remote files. This change explicitly looks for channels starting with `file://` and handles them with aiofiles, which is already a dependency. I've tested this locally with some private channels and it seems to work nicely.

This is probably nicer than my currently very-ugly-hack of just copying the channel contents into the conda-bld/ directory within my build container and then cleaning things up afterward.

### DIFF
--- a/bioconda_utils/utils.py
+++ b/bioconda_utils/utils.py
@@ -1432,7 +1432,7 @@ class AsyncRequests:
         result = []
         if url.startswith("file://"):
             if os.path.exists(url[7:]):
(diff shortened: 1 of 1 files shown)


epruesse (approved):
Makes sense to me! :-)

PR #1105, opened 2026-04-29
refactor: add type annotations

paperbenni:
- add type annotations to most of the codebase
- fix bugs surfaced by type annotations
    - outdated method name still in use, in code path not caught by unit tests
    - account for values possibly being none in places which did not expect it
- add CI job for checking for type regressions
- used pyupgrade to switch from `Dict` to built-in `dict` and remove other deprecated syntax

changes in .github/quay-namespace-info.py are line endings, hence the massive diff


.github/workflows/GithubActionTests.yml
```diff
@@ -15,6 +15,22 @@ jobs:
         name: ruff format
         with:
           args: "format --check --diff"
+      - name: set path
```
bgruening, line 18: I would be fine to decouple this from the conda env. But this can happen in a follow-up.

bioconda_utils/artifacts.py
```diff
@@ -244,10 +248,13 @@ def get_circleci_artifacts(check_run, platform):
 
 def parse_gha_build_id(url: str) -> str:
     # Get workflow run id from URL
-    return re.search("runs/(\d+)/", url).group(1)
+    match = re.search(r"runs/(\d+)/", url)
+    if match is None:
+        raise ValueError(f"Could not parse GitHub Actions run ID from {url}")
```
bgruening, line 253: Nice.

bioconda_utils/autobump.py
```diff
@@ -220,23 +225,23 @@ class Scanner(AsyncPipeline[Recipe]):
...
-        self.status: List[Tuple[str, str]] = []
+        self.status: list[tuple[str, EndProcessingItem]] = []
         #: filename to write statuses to
-        self.status_fn: str = status_fn
+        self.status_fn = status_fn
         #: async requests helper
         self.req = AsyncRequests(cache_fn)
 
-    def run(self) -> bool:
+    def run(self) -> None:
```
bgruening, line 250: this function seems to have a return value

bioconda_utils/autobump.py
```diff
@@ -494,12 +503,12 @@ def find_reason(cls, recipe, metas):
             avail_vers = RepoData().get_package_data("version", name=var)
             if avail_vers:
                 resolved_vers[var] = list(
-                    set(
+                    {
```
bgruening, line 516: why that? that is not  a set anymore, maybe also Type this function?

bioconda_utils/bioconductor_skeleton.py
```diff
@@ -55,29 +56,27 @@
...
-        "clustersim",
-        "hiver",
-        "lidr",
-        "mixomics",
-        "snpls",
-        "matlib",
-        "qpcr",
-    ]
-)
+CRAN_X_PACKAGES = {
```
bgruening, line 59: why are you changing set() to dicts?

bioconda_utils/build.py
```diff
@@ -422,18 +449,18 @@ def build_recipes(
 
     # setup linting
     if do_lint:
-        always_exclude = ("build_number_needs_bump",)
+        always_exclude = ["build_number_needs_bump"]
         if not lint_exclude:
             lint_exclude = always_exclude
         else:
-            lint_exclude = tuple(set(lint_exclude) | set(always_exclude))
```
bgruening, line 429: why are those changes needed?

bioconda_utils/hosters.py
```diff
@@ -82,18 +86,25 @@ class HosterMeta(abc.ABCMeta):
...
     ) -> type:
         """Creates Hoster classes
 
         - expands references among ``{var}_pattern`` attributes
         - compiles ``{var}_pattern`` attributes to ``{var}_re``
         - registers complete classes
         """
-        typ = super().__new__(cls, name, bases, namespace, **kwargs)
+        typ = cast(
+            type["Hoster"],
```
bgruening: "Hoster" or Hoster?

bioconda_utils/hosters.py
```diff
@@ -842,9 +856,9 @@ async def get_versions_from_json(self, data, req, orig_version):
 class CRAN(JSONHoster):
     """R packages hosted on r-project.org (CRAN)"""
 
-    async def get_versions_from_json(self, data, _, orig_version):
+    async def get_versions_from_json(self, data, req, orig_version):
         res = []
-        versions = list(set((str(data["latest"]), self.vals["version"], orig_version)))
+        versions = list({str(data["latest"]), self.vals["version"], orig_version})
```
bgruening, line 850: why not set?

bioconda_utils/utils.py
```diff
@@ -63,7 +64,7 @@
 from colorlog import ColoredFormatter
 from boltons.funcutils import FunctionBuilder
 
-conda.gateways.logging.initialize_logging = lambda: None
+cast(Any, conda.gateways.logging).initialize_logging = lambda: None
```
bgruening, line 67: Hu, that looks wired. maybe some typing magic? Can we remove that?

test/test_hosters.py
```diff
@@ -118,7 +121,7 @@ def test_get_version(self, event_loop):
             pytest.xfail("No release_links or release_json in test case")
 
         versions_data = event_loop.run_until_complete(
-            self.instance.get_versions(self, self.case["version"])
+            self.instance.get_versions(cast(Any, self), self.case["version"])
```
bgruening: that looks also pretty useless and degrades readability.

test/test_utils.py
```diff
@@ -108,7 +110,7 @@ def config_fixture():
     yield config
 
 
-@pytest.fixture(scope="module", params=PARAMS, ids=IDS)
+@pytest.fixture(scope="function", params=PARAMS, ids=IDS)
```
bgruening, line 113: this needs to be double checked

justfile
bgruening, line 1: Maybe a Makefile would also be useful for Humans :)

paperbenni:
ty is now newer on CI than on my testing machine, hence uncaught type errors

test/test_utils.py
paperbenni: image
I did, this fixes a test failure where one test breaks the other when both are running in the same VM

bioconda_utils/autobump.py
paperbenni: It returns the return value of a function which also returns none :raised_eyebrow:

bioconda_utils/bioconductor_skeleton.py
paperbenni: pyupgrade. This is the new set syntax. There are no keys, so this results in a set, not a dict

bioconda_utils/artifacts.py
paperbenni: :smile:

bioconda_utils/autobump.py
paperbenni: image
new syntax

bioconda_utils/bioconductor_skeleton.py
paperbenni: image

it is a set

justfile
paperbenni: we'd have to keep both in sync. `just` is not meant for agents, it's just nicer in general and actually meant as a taskrunner, not a build system :)

https://just.systems/man/en/what-are-the-idiosyncrasies-of-make-that-just-avoids.html

image

env variables in particular are a lot nicer

.github/workflows/GithubActionTests.yml
paperbenni: ty needs dependencies to be installed to check types of values returned from library functions. It is quite a lot of stuff to process for a single CI check though, maybe pixi would be nice here :)

bioconda_utils/build.py
paperbenni:
```
    exclude: list[str] | None = None,
```

the original cast was wrong, this is only ever used in places where list is expected (annotations predating my changes, just never checked), Sequence would probably be most accurate

bioconda_utils/hosters.py
paperbenni: Hoster, fixed

bioconda_utils/utils.py
paperbenni: This is a workaround to make ty happy, can be fixed through proper dependency inversion, that might be another big change though

test/test_hosters.py
paperbenni: this was a workaround to make ty happy, fixed through dependency inversion

PR #1109, opened 2026-05-07
feat: migrate images from bioconda-containers

paperbenni:
Migrate images from bioconda-containers to bioconda-utils to bring new bioconda-utils master more in line with #1049  

#1049  migrates images from bioconda/bioconda-containers to bioconda/bioconda-utils. Containers have been updated in bioconda-containers since, so this contains the new versions of bioconda-containers. After this is merged, bioconda-containers should be archived

the env vars whose names #1049 changed were not used in any repo in bioconda

since the containers and #1049 contain a lot of untested bash code, I added bash linting and formatting and fixed some minor issues the linter found

images from bioconda-containers have not been updated or built in a long time, hence some fixes to things which broke in the meantime
- mamba errs out, switched to conda
- busybox cgit went down, switched to official GH mirror linked on busybox.net
- switched to skopeo from quay REST api to circumvent strict quay REST rate limits
- fix broken version parsing regex not liking new format

PR #1115, opened 2026-06-12
feat: multi platform biocontainers

paperbenni:
# Platform argument

This allows specifying which target platform a biocontainer or package should have should have from the CLI.
This simultaneously removes previous X86-only assumptions and opens the door for cross-compilation by simply specifying a foreign platform (docker can run foreign architecture containers using Qemu binfmt. By building inside such a container we can build for foreign architectures without dedicated hardware or a full VM). 

Both mulled-build and involucro support specifying a target platform now, so this part of the PR mostly just threads that argument through to them. 

# Multi-arch manifests

Since native ARM runners are already in use in bioconda/bioconda-recipes, I aim to keep the package builds on separate runners (as Qemu comes with a performance penalty). X86 and ARM will build and publish their own container images, with arm appending a suffix to its image name (involucro limitation). each image build job also emits metadata about its image as a JSONL artifact. 

A subsequent job then collects all JSONL records, builds a single manifest out of all of these, and then publishes that manifest. This means a consumer can irrespective of architecture pull the image with the same name, and the manfiest redirects to the image for the correct architecture.  

# Blocked by 

https://github.com/bioconda/bioconda-utils/pull/1128

# Renames

Some terminology was outdated or misleading. Anything part of that which was touched by this PR has been renamed more fittingly. Noteworthy examples

- `release_conda` is a cleanup function, but sounds like a publishing function
- `test_package` builds production biocontainer artifacts
- same for --mulled-test, this builds production artifacts

# Types

Conda, OCI and Docker use different notations for platforms. When all of them were typed as `str`, it was quite easy to mix them up. This PR strengthens type safety for platform arguments. 

I found readability with some functions to be quite poor due to `str` arguments with non-descript names. These were tightened to Path or suffixed with `_url` where appropriate. The same goes for config handling. Some functions were taking Path | dict configuration, scattering config loading and normalization, and making it possible to use unverified, unnormalized config in places on accident. 

# Refactors

Some functions were already quite long and the extra functionality from this PR has blown then past ~250 lines. Due to readability suffering, these were split into smaller, more readable functions

# Bug fixes

A new ruff version discovered some minor bugs which were fixed

# Follow-up

update callsites in bioconda-recipes to actually make use of the new features

https://github.com/bioconda/bioconda-recipes/pull/67786


aliciaaevans:
I'm going to try to review this over the next few days, so I'll probably have some questions. Right now just getting back up to speed with the details of the current state of containers, etc.

Looks like the mulled changes are set for Galaxy 26.2 which isn't release yet, so I won't merge it yet.

If you have looked at https://github.com/bioconda/bioconda-utils/pull/1049 and can quickly summarize similarities/differences in approach that might help. (If not, that's OK.) I see the container images were in a previous PR (https://github.com/bioconda/bioconda-utils/pull/1109), were those based on the most recent state of bioconda-containers or #1049? (I'm not seeing a lot of the bash scripts from #1049, so I'm guessing the former.)

paperbenni:
#1049 has a different goal than this. There are dependencies between bioconda-utils and the base images. This means if multiple ones are updated at the same time, they could be using stale versions of the others in various places (meaning an image from quay instead of the same commit). Actions doesn't allow modelling this cleanly (or didn't at the time #1049 and its predecessors were written), so a lot of the bash in #1049 is reinventing some GH actions to allow for this. 

Given that base images update so infrequently that nobody noticed way more pressing problems (excerpt from thesis) with the image builds, I am not entirely sure if we want to maintain thousands of lines of bash for this.  

image


The important functionality in #1049 for ARM biocontainers is multi-arch manifests for the base images, so that an x86 base image doesn't have a different name than the ARM one. That has been carried over to bioconda-utils already though, separately from the manual build system. 
(though at the time CI had invalid credentials so there'll have to be another image bump to actually get it published)

Apart from that, being architecture agnostic and modelling inter-image dependencies are separate problems from what I can see.

paperbenni:
The mulled changes were backported to 26.1 which should hopefully be out today. And yes, the base images are from bioconda-containers which as far as I remember had more recent changes than #1049 did

aliciaaevans:
@paperbenni very helpful, thanks! From my discussions with @daler, the original intention was to simplify things, but the dependency issues made the complexity grow.

I should have time to look at this more later today.

bioconda_utils/cli.py
```diff
@@ -145,16 +209,29 @@ def _validate_git_range(value: str | None) -> str | None:
...
+    typer.Option(
+        "--use-existing-auth",
+        help="Use existing Docker or skopeo registry authentication when Quay credentials are unset.",
+    ),
+]
+MulledUploadRecordsOpt = Annotated[
+    Path | None,
+    typer.Option(
+        "--mulled-upload-records",
+        help="Append uploaded mulled image records as JSONL for manifest publication.",
```
aliciaaevans: This help text is a little confusing. It reads like a boolean but it's really expecting a path.

bioconda_utils/lint/__init__.py
aliciaaevans:
```suggestion
          recipe_name: Name of recipe to lint
```
just noticed this

bioconda_utils/build.py
```diff
@@ -350,42 +427,40 @@ def get_subdags(
...
     recipe_obj = _recipe.Recipe.from_file(recipe_folder, recipe)
-    # On linux-aarch64 or osx-arm64 env, only build recipe with matching extra_additional_platforms
-    if platform == "linux-aarch64":
-        if "linux-aarch64" not in recipe_obj.extra_additional_platforms:
-            return True
-    if platform == "osx-arm64":
-        if "osx-arm64" not in recipe_obj.extra_additional_platforms:
-            return True
-    return False
+    primary_platforms = {PackageSubdir.LINUX_64, PackageSubdir.OSX_64}
```
aliciaaevans: I like this change, but may want to make this more easily configurable rather than hard-coded. In case we want to make osx-64 still available as an additional platform. I see there's some discussion on #1121

bioconda_utils/container_manifests.py
```diff
@@ -0,0 +1,517 @@
...
+    m = _BUILD_STRING_RE.match(build_string)
+    if m and m.group(1):
+        return m.group(1)
+    return ""
+
+
+def multiarch_ref(canonical_ref: str) -> str:
+    """Return the platform-independent multi-arch tag for a canonical ref.
+
+    Strips the conda build hash from the tag, preserving version and any
```
aliciaaevans, line 103: Thinking through the results of doing this. If we have a recipe build number bump, are we potentially going to end up with different build numbers under one multi-arch at different times? We have some cases where one arch builds and the other needs a rebuild, etc. Will the underlying images still have the full build string tag? and then do they get replaced in the manifest if a new build comes in? 

It may be fine, but just trying to think if there are any implications that need to be considered.

aliciaaevans (commented):
Here's my first batch of comments. I am still looking through a couple of the larger files, but I haven't found any big concerns.

What kind of testing have you been able to do? At some point we need to just get it out there and monitor it, of course.

I did try to use the platform setting to build linux-aarch64 on my local (I built the linux-64 and linux-arm64 build-env images locally first, since they don't exist on quay.io at the moment). But I've been struggling a bit with errors relating to the package not found due to the expected path and build string not matching. Not sure if there's something up with my environment. I can provide more information, but I just want to check that that's something I should be able to do with this change.

paperbenni:
> Here's my first batch of comments. I am still looking through a couple of the larger files, but I haven't found any big concerns.
> 
> What kind of testing have you been able to do? At some point we need to just get it out there and monitor it, of course.
> 
> I did try to use the platform setting to build linux-aarch64 on my local (I built the linux-64 and linux-arm64 build-env images locally first, since they don't exist on quay.io at the moment). But I've been struggling a bit with errors relating to the package not found due to the expected path and build string not matching. Not sure if there's something up with my environment. I can provide more information, but I just want to check that that's something I should be able to do with this change.

this needs to be merged first

https://github.com/bioconda/bioconda-utils/pull/1128

edit: resolved, perms waiting for actual base image release

bioconda_utils/container_manifests.py
paperbenni: I am taking a similar approach to many images on docker hub

image


With this new approach, packages get multiple identifiers with varying specificity, the multi arch tag is the less specific one. This one is mutable and always points to the latest image providing that version of the package (retaining intentional variant matrices as separate names). 

The fully qualified names are referring to a specific build of the package, as the their names include a hash of the metadata which will differ between architectures. Due to having completely different names, as far as quay is concerned, fully qualified architecture variants of the same package have nothing to do with each other.

paperbenni: This system is easily extensible, for example some docker images also publish tags with the base image name in the name, so that users can get the same application code with different base distros, or we could publish image:major image:major.minor and image:major.minor.patch at the same time

aliciaaevans: Ok, great, that lines up with what I'd expect.

bioconda_utils/upload.py
```diff
@@ -2,14 +2,95 @@
...
+        response.raise_for_status()
+    else:
+        data = response.json()
+        if data.get("is_public") is False:
+            logger.warning(
+                "Repository %s/%s is private; changing visibility to public",
+                namespace,
+                repository,
+            )
+            response = requests.post(
```
aliciaaevans: It may be that the token used by bioconda-recipes doesn't actually have the organization permissions to do this. I think that might be why it was done by a separate job in bioconda-utils.

bioconda_utils/docker_utils.py
```diff
@@ -460,12 +510,18 @@ def build_recipe(
         build_dir = os.path.realpath(tempfile.mkdtemp())
         # conda_exports.subdir is {platform}-{arch} like: 'linux-64' 'linux-aarch64'
         script = self.build_script_template.format_map(
-            {"self": self, "arch": "noarch" if noarch else conda_exports.subdir}
+            {
+                "self": self,
+                "arch": "noarch" if noarch else conda_exports.subdir,
```
aliciaaevans: I think this is where my local testing is going wrong.  conda_exports.subdir is "linux-64" and the resulting linux-aarch64 conda package file is getting copied to that host directory by the build script. Then bioconda-utils is failing to find the package in "conda-bld/linux-aarch64".

bgruening:
I have triggered the CI and fixed the permissions.

bioconda_utils/docker_utils.py
aliciaaevans: Just to expand on this comment, is `conda_exports.subdir` expected to honor the platform setting at this point? Since it's outside the container, I don't think it's getting the right subdir here.

bioconda_utils/upload.py
aliciaaevans: (I'm referring to the biocontainers org permissions here.)

aliciaaevans:
Overall, good stuff! Thanks for taking this on, @paperbenni.

I got through all the files (finally), so what I've commented already are my only questions/concerns that might hold up merging this. I'll take a look at the corresponding bioconda-recipes PR next, hopefully can make time tomorrow.

bioconda_utils/build.py
paperbenni: I made it configurable through the config file

bioconda_utils/docker_utils.py
paperbenni: This did lose platform information, I fixed it

bioconda_utils/cli.py
paperbenni: I renamed the argument. I find the general direction of naming user facing arguments after the fact they're using mulled under the hood questionable.

bioconda_utils/upload.py
paperbenni: I'm not sure, but given how fun quay permission management always is, I removed it :sweat_smile: 
It's just an optional nicety anyway

aliciaaevans: Sounds good

bioconda_utils/docker_utils.py
aliciaaevans: Yup, that fixed it. :smiley:

aliciaaevans (approved):
@paperbenni everything looks good to me. I'm going to approve, but looks like you were still making a few tweaks. Are you ready for this to be merged?

paperbenni:
I'm ready for it to be merged :+1:

bgruening:
Lovely, thanks a lot!

paperbenni:
> I have triggered the CI and fixed the permissions.

there still are no base images for 4.5.0 and the older ones dont have multi arch manifests. @bgruening if you dont want to give the CI base image bump permissions, then maybe could you manually create and upload a manifest of the old images?

aliciaaevans:
Are 4.5.0 images needed or can we just move forward with releasing 5.0.0?  I think it's been too long to rerun the 4.5.0 actions.

bgruening:
Can we go with 5.0.0 instead as @aliciaaevans suggested? Then its also a clear cut :)

paperbenni:
> Can we go with 5.0.0 instead as @aliciaaevans suggested? Then its also a clear cut :)

That works as well

bgruening:
@aliciaaevans do you want to merge the 5.0 PR?

aliciaaevans:
@paperbenni @bgruening The 5.0.0 images are up now.

bgruening:
:tada: number one! 

What should we do next, make it work on bioconda-recipes master, or first one bulk and test with 1-2 recipes?

aliciaaevans:
I like the idea of trying it on bulk. I can work on it later today.

PR #1116, opened 2026-06-14
fix: platform detection

paperbenni:
When `_load_platform_metas()` returns `"linux"` on ARM64 hardware, `config.platform` is set to `"linux"`, which causes conda-build to resolve x86_64 variants. 
This fixes that by doing proper platform detection

### DIFF
--- a/bioconda_utils/cli.py
+++ b/bioconda_utils/cli.py
@@ -1509,6 +1509,7 @@ def autobump(
     help="Platforms to annotate",
     nargs="+",
     type=str,
+    choices=sorted(conda.base.constants.PLATFORM_DIRECTORIES),
 )
 @arg(
     "--existing-only",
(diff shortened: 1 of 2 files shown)


bioconda_utils/cli.py
```diff
@@ -1509,7 +1509,6 @@ def autobump(
     help="Platforms to annotate",
     nargs="+",
     type=str,
-    default=["linux-64", "osx-64"],
```
bgruening, line 1512: Can we add here a list of supported platforms/strings please?

bgruening (approved):

bioconda_utils/cli.py
paperbenni: https://github.com/bioconda/bioconda-utils/pull/1116/changes#diff-510c03d8200a68840aa42ef413cf2630ce781badcb54433dcc750a8d016d2d51R1512

done.

PR #1118, opened 2026-06-17
chore: change bioconda-utils to build using pixi

paperbenni:
image
brings down solving time for bioconda-utlis by a lot 

pixi.toml is now the single source of truth

the dependencies are also parsed back into the legacy format because they get shipped as data (also an odd practice, something to revisit someday). Drift gets prevented through a CI check

CI boilerplate is reduced due to using the existing pixi action

justfile was updated to utilize pixi

IDEs pick up the pixi env automatically like they did with the conda env

### DIFF
--- a/bioconda_utils/bioconda_utils-requirements.txt
+++ b/bioconda_utils/bioconda_utils-requirements.txt
@@ -2,11 +2,7 @@
 # runtime [dependencies] only; feature/dev/test dependencies are excluded
 # See https://github.com/bioconda/bioconda-utils/blob/main/pixi.toml
 
-
-# pinnings
 conda-forge-pinning==2026.01.07.13.16.30
-
-# basics
 python=3.13.*
 setuptools=81.*
 setuptools-scm>=8
@@ -34,11 +30,7 @@ involucro=1.1.*
 skopeo=1.22.*
 findutils
 git=2.*
-
-# hosters - special regex not supported by RE
 regex=2026.*
-
-# asyncio
 aiohttp=3.13.*
 aiohttp-jinja2
 aiohttp-session
@@ -47,25 +39,13 @@ aiofiles=25.*
 aioftp=0.27.*
 backoff=2.2.*
 cachetools=7.0.*
-
-# client API wrappers
 gitpython=3.1.*
 gidgethub=5.*
 pyjwt>=2.4.0
-
-# bioconductor-skeleton
 beautifulsoup4=4.*
 requests=2.33.*
-
-# merge handling
 pygithub
-
-# caching
 diskcache=5.*
 platformdirs=4.*
-
-# build failure output
 tabulate=0.10.*
-
-# resource reporting for builds
 psutil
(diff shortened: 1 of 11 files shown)


bioconda_utils/bioconda_utils-requirements.txt
```diff
@@ -11,60 +12,40 @@ conda-build=26.3.*
...
-# caching
-diskcache=5.*                 #
-platformdirs=4.*              #
-
-# build failure output
-tabulate=0.10.*               #
-
-# resource reporting for builds
-psutil                        #
+argh=0.31.*
```
bgruening, line 15: Can you keep the grouping and comments when the still apply?

paperbenni: The comments and dependencies were moved to pixi.toml, the txt file is only generated from the toml for backwards compatibility. Are there benefits to having them in both? I can expand the conversion script, but that will make it even bigger

paperbenni: tomllib discards comments, so it would probably be uglier too. Still very much doable, but not pretty

bgruening: I see. No this is fine. Please update the documentation as well.

justfile
```diff
@@ -5,22 +5,32 @@
...
+
+check:
+    pixi run -e dev ruff check .
+    pixi run -e dev ty check .
+    pixi run -e dev shellcheck {{shell_scripts}}
 
 # this takes a very long time to execute, use check if not finished with your work yet
 test: install
-    pytest
+    pixi run -e dev pytest
```
johanneskoester: I wonder if the justfile is still needed. By properly crafting the rules here as pixi tasks, they are similarly simple.

```
pixi run test
pixi run check
pixi install
pixi run format
pixi run global-install # implementing the commands that are shown above under install
```

johanneskoester (changes requested):
Very cool! For a second, I thought that this PR is about using pixi to build the recipes :-D.

bioconda_utils/bioconda_utils-requirements.txt
johanneskoester: Can't we get rid of the old requirements.txt overall while migrating to pixi? We can deploy bioconda-utils via pixi global install anyway. For what do we still need the requirements.txt?

paperbenni: The txt file is shipped as part of the actual package, so that would be a breaking change. I think long term we should stop doing that, but maybe that's a separate PR

paperbenni: bioconda-utils also parses its own dependencies txt at runtime, that part would have to be rewritten.  Not sure whether or not that is out of scope for this PR

justfile
paperbenni: I moved most tasks to pixi now. 
I like the Shell and IDE integration of just, and it is a nice way to prevent agents from running off in random directions. I kept the justfile around as a thin wrapper and to distinguish frequently needed dev tasks.

bgruening (approved):

PR #1121, opened 2026-06-22
chore: Drop osx intel

paperbenni:
The Next MacOS version is dropping X86 support, even Rosetta is EOL. This removes all MacOS X86 references, and quite conveniently saves us a lot of CI time.

Also removes some dead Travis CI references

### DIFF
--- a/.github/workflows/GithubActionTests.yml
+++ b/.github/workflows/GithubActionTests.yml
@@ -69,7 +69,7 @@ jobs:
           fi
   test-macosx:
     name: OSX ARM tests
-    runs-on: macos-latest
+    runs-on: macos-26
     steps:
       - uses: actions/checkout@v6
         with:
--- a/.github/workflows/GithubActionTests.yml
+++ b/.github/workflows/GithubActionTests.yml
@@ -82,8 +82,8 @@ jobs:
               echo "Skipping pytest - only docs modified"
           fi
   test-macosx:
-    name: OSX tests
-    runs-on: macos-15-intel
+    name: OSX ARM tests
+    runs-on: macos-latest
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/GithubActionTests.yml
```diff
@@ -68,8 +68,8 @@ jobs:
               echo "Skipping pytest - only docs modified"
           fi
   test-macosx:
-    name: OSX tests
-    runs-on: macos-15-intel
+    name: OSX ARM tests
+    runs-on: macos-latest
```
bgruening: Should we pin to a specific release? I fear we are getting here incompatibilities

bgruening (approved):

.github/workflows/GithubActionTests.yml
paperbenni: I pinned to 26 :)

jmarshall:
Dropping macOS-intel support is surely a policy decision that requires a formal decision.

Many projects are continuing to build for Intel macOS until GitHub decommissions its Intel macOS runners, which is [scheduled for August 2027](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/#notice-of-macos-x86_64-intel-architecture-deprecation) at the earliest, though this may have been extended with the [deployment of the `macos-26-intel` image](https://github.com/actions/runner-images/issues/13739#issuecomment-3971668146).

bgruening:
@paperbenni how hard would it be to make osx and opt-in, just like arm-osx currently?

bgruening:
@jmarshall its a capacity problem, too many architectures. I will try to create an issue to get more feedback. Would an opt-in work for you for the time being?

paperbenni:
> @paperbenni how hard would it be to make osx and opt-in, just like arm-osx currently?

I am having my last exam tomorrow, so I can't look into it right now, but from what I remember it would be easier to just keep osx-64 around for the time being

PR #1122, opened 2026-06-25
chore: remove unused code

paperbenni:
Remove unused code

I used vulture to build an initial list, then excluded false-positives (auto-discovered code etc) as well as anything imported by external scripts in bioconda-recipes. 

Lots of html templates were unused since 2019, these are removed as well


bioconda_utils/autobump.py
aliciaaevans:
```suggestion
```
Since we are getting rid of check_version_pin_conflict, this should be removed.

bioconda_utils/autobump.py
```diff
@@ -64,7 +64,6 @@
 import conda_build.variants
```
aliciaaevans:
```suggestion
```
Looks like this import also unused after removing check_version_pin_conflict.

test/test_utils.py
```diff
@@ -982,48 +982,6 @@ def test_build_container_no_default_gcc(tmpdir):
     )
     assert build_result.success
 
-    for k, v in r.recipe_dirs.items():
```
aliciaaevans, line 985: Why is this block being removed? Just checking if it's intentional or only the test after this.

bioconda_utils/githandler.py
```diff
@@ -712,7 +700,3 @@ def close(self) -> None:
 
 class BiocondaRepo(GitHandler, BiocondaRepoMixin):
     pass
-
-
-class TempBiocondaRepo(TempGitHandler, BiocondaRepoMixin):
```
aliciaaevans, line 717: If this isn't needed, then I think TempGitHandler isn't needed either.

bioconda_utils/cli.py
```diff
@@ -390,30 +390,6 @@ def remove_package(spec):
...
-     --pull-request). Also needs --user and --repo to be set. Requires the env
-     var GITHUB_TOKEN to be set.""",
-)
-@arg(
-    "--pull-request",
-    type=int,
-    help="""Pull request id on github on which to
-     post a comment.""",
-)
 @arg("--user", help="Github user")
```
aliciaaevans: Looks like `user` and `repo` aren't used either.

bioconda_utils/cli.py
```diff
@@ -424,13 +400,6 @@ def remove_package(spec):
      include uncommitted changes). All recipes modified within this range will
      be built if not present in the channel.""",
 )
-@arg(
-    "--full-report",
```
aliciaaevans, line 428: Looks like some of the workflows in bioconda-recipes are trying to use `--full-report`. That should be updated before this version of bioconda-utils is used for builds.

aliciaaevans (commented):
Thanks, some of this hasn't been needed in years. 

I still want to take a look at githubhandler.py tomorrow (ran out of time), but here are my comments so far. I think I found a couple more things that can be removed, unless I'm missing some place they are used.

bioconda_utils/autobump.py
aliciaaevans:
```suggestion
```
This comment can be removed since you're removing check_version_pin_conflict.

aliciaaevans (changes requested):
I think that one for loop in test_utils.py is the only thing concerning to me as it's not clear why it was removed.

bioconda_utils/cli.py
paperbenni: removed :)

bioconda_utils/autobump.py
paperbenni: removed

bioconda_utils/githandler.py
paperbenni: removed those :)

test/test_utils.py
paperbenni: I selected too many lines when deleting the test and since it's a cleanup on an xfail which normally does not run it did not break anything. I restored it

bioconda_utils/cli.py
paperbenni: https://github.com/bioconda/bioconda-recipes/pull/66744

I did a PR for that

aliciaaevans (approved):
Looks good. I think the osx tests timed out, so rerunning.

bgruening:
Wonderful!

PR #1124, opened 2026-07-02
docs: showcase more pixi usage details

dlaehnemann:
@paperbenni, as this goes along your recent pixi changes, you're probably best suited to review this!

I was just trying out the latest development version to check my linting updates against all of the recipes in `bioconda-recipes` and got thrown by the new `Just` wrapper recommendation. Having another level of wrappers hiding the rather new `pixi` setup, confused me a bit. And I think explaining a bit what's where and how things work, with linkouts to pixi docs, reduces the barrier of contributing a tiny bit. Feel free to edit further for even more clarity...

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,17 @@ bioconda-utils, you can use the respective Pixi tasks defined in
 [`pixi.toml`](pixi.toml). You will need
 [an installation of `pixi`](https://pixi.prefix.dev/latest/installation/).
 Then, you can install the current `bioconda-utils` version in your local folder
-into the `dev` environment, by running the `install` task:
+into the `dev` environment, by
+[running](https://pixi.prefix.dev/latest/reference/cli/pixi/run/) the `install`
+task:
 
 ```bash
 pixi run install
 ```
 
-To then run `bioconda-utils` from anywhere, start a shell with that environment
-activated:
+To then run `bioconda-utils` from anywhere, start a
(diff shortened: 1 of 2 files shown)


paperbenni:
I like the IDE and shell integration of `just`.
Spending lots of time with Rust has given `just` a firm place in my muscle memory 😅️

I also wanted to do some progressive disclosure for agents, an agent shouldn't have the dependency list forced into its context whenever it wants to verify it hasn't produced invalid syntax. 
I am not hugely attached to `just` though, `pixi task list` seems like it also does the job. 

The shellcheck script discovery would also be a lot more awkward in pixi.toml, `just` being [language agnostic](https://just.systems/man/en/shebang-recipes.html) is nice there

dlaehnemann:
I think it's fine to keep both. I only wanted to also properly document the pixi background and workflow, and point out that the just installation wrapper does a global install.

Ah, and I guess for the formal review requirement we'll have to find someone with the appropriate permissions, right?

paperbenni:
> I think it's fine to keep both. I only wanted to also properly document the pixi background and workflow, and point out that the just installation wrapper does a global install.
> 
> Ah, and I guess for the formal review requirement we'll have to find someone with the appropriate permissions, right?

image

best I can do is this

dlaehnemann:
At least that's loud and clear.

aliciaaevans (approved):
Thanks, I haven't been exposed to just and was hesitant to install another tool. I like that this shows both approaches.

PR #1125, opened 2026-07-02
chore!: migrate CLI dispatch from argh to typer

paperbenni:
# What & Why

Argh caused some type errors to not be picked up by `ty`. Typer enables analyzing the entire path from CLI to leaf functions better, preventing drift. It is also much better maintained (last argh update ~2 years ago), and allows generating shell completions etc. 

# breaking CLI changes

- `--testonly` -> `--test-only`
- `--pkg_dir`  -> `--package-dir`
- `--skiplist-leafs` -> `--skiplist-leaves`
- `--no-presolved-mulled-test` → `--presolved-mulled-test`/`--no-presolved-mulled-test`
- `--prelint` removed (was alias for `--lint`)

Git-range now needs a repeated flag for providing multiple values

- `--git-range a b` (space-separated, one flag) -> `--git-range a --git-range b` (repeated flag)

Corresponding recipes PR accounts for these

https://github.com/bioconda/bioconda-recipes/pull/66851

backwards compatibile aliases were not added, because typer does not properly support the git-range syntax, which means a breaking change is needed either way, and one of the argument names is fixing an actual typo


bgruening (approved):
@paperbenni thanks for implementing this!

Looks good for me! @aliciaaevans do you want to have a look here?

aliciaaevans:
Cool, I'm taking a look in detail now. Probably should add `!` to the PR title/commit for the breaking changes. Also, can you make a PR to update bioconda-docs with the `--git-range` change (and any others, I just did a quick search)?

We'll have to make sure to coordinate the bioconda-recipes PR with updating bioconda-common.

aliciaaevans:
Found another breaking change:

`bioconda-utils bioconductor-skeleton recipes config.yml DESeq2` example from the docs no longer works. With this has to be `bioconda-utils bioconductor-skeleton DESeq2 recipes config.yml`. Although, it might make more sense to use `--packages` (or some other name since it omits `bioconductor-`) like the other commands if it's going to break anyway.

jmarshall:
If `--git-range` means what the name suggests, then the two-argument form could be usefully expressed as `--git-range a..b` and be expressible in the new parser.

paperbenni:
> If `--git-range` means what the name suggests, then the two-argument form could be usefully expressed as `--git-range a..b` and be expressible in the new parser.

I'll bring it in line with the official `git diff A...B` [syntax](https://git-scm.com/docs/git-diff). `A..B` already has another meaning in Git, that could be confusing.

jmarshall:
Whichever. The syntax one usually wants is [`git log a..b` or `git diff a...b`](https://matthew-brett.github.io/pydagogue/pain_in_dots.html). Whether this usage is more like log or more like diff is 🤷

paperbenni:
> Found another breaking change:
> 
> `bioconda-utils bioconductor-skeleton recipes config.yml DESeq2` example from the docs no longer works. With this has to be `bioconda-utils bioconductor-skeleton DESeq2 recipes config.yml`. Although, it might make more sense to use `--packages` (or some other name since it omits `bioconductor-`) like the other commands if it's going to break anyway.

I changed it to `--package` now

paperbenni:
> Cool, I'm taking a look in detail now. Probably should add `!` to the PR title/commit for the breaking changes. Also, can you make a PR to update bioconda-docs with the `--git-range` change (and any others, I just did a quick search)?
> 
> We'll have to make sure to coordinate the bioconda-recipes PR with updating bioconda-common.

https://github.com/bioconda/bioconda-docs/pull/53

This updates the docs. 
The docs referenced the legacy deps txt file from this repo, so I changed the build pipeline (and theme while I was at it)

aliciaaevans (approved):
Looks good to me.

I like the `--git-range` update, so thanks for the suggestion, @jmarshall.

It would be nice if there's a way to not have to parse that string ourselves but pass it directly to the git package, but I'm OK with this.

I'll take a look at the docs PR. Thanks!

bgruening:
Cool! This is shaping up nicely!

PR #1128, opened 2026-07-31
refactor: unify container building/testing

paperbenni:
# Why

All container build workflows were nearly identical and included some drift.
Some of the tests also accidentally tested a local temporary image instead of the claimed remote image. 
The cos7 image being inconsistent with the rest of the building workflows also caused it to attempt to push to a private repo. 

Also only one of the Dockerfile.test files was actually repeatedly tested, with the other ones testing the incorrect image. 

It was possible a bug would pass PR tests but fail release CI (or get released)

# What

Extract duplicated build logic into a shared action and a shared test script.  

This PR also removes the superseded responder image, (bot image is the one used nowadays)

All images now use multi-arch manifests instead of architecture specific suffixes.  
This means imagename-aarch64 is no longer valid, and instead all architectures simply use imagename

This will enable (and require) simplifying the bioconda-recipes workflows which use conditional image names

Fixes the bug testing the wrong image so that now all Dockerfile.test file are actually used with their correct base images


paperbenni:
new ruff is way more strict, this has been fixed in #1115

bgruening:
This looks mostly good to me, ruff is unhappy, can you make it happy.

bgruening:
@aliciaaevans do you want to have a look, or merge after fixing ruff?

paperbenni:
image

I can, but it would basically be cherry picking thousands of lines from #1115

paperbenni:
@aliciaaevans can this be merged? I'd rather not do all the ruff work again if it's being overwritten with  #1115 anyway

aliciaaevans (approved):
Makes sense.

aliciaaevans:
@bgruening it seems there is still an issue with the quay credentials? I will check, but I don't think I have access to that.

aliciaaevans:
Hmm, create-env was pushed successfully.

paperbenni:
I think some of the base images have separate more strict upload policies since changing them could break more

aliciaaevans:
Yeah, hopefully someone with that access can take a look. I put a message on the core team chat.

PR #1133, opened 2026-09-11
fix: fix pixi env mixup

paperbenni:
addresses 
https://github.com/bioconda/bioconda-recipes/issues/68981

also unifies installation methods across all CI jobs and providers so regressions are caught earlier


bgruening (approved):
Danke, das war fix

PR #1134, opened 2026-09-14
refactor: split up massive file

paperbenni:
utils.py was mixing unrelated things, and 1.7k lines was not great for maintenance

- split file into modules, so things are easier to find
- remove two unused helpers
- update pixi lock

There were some hacky lookup paths, which relied on `__file__` or $PATH instead of the conda env and could silently break in weird ways. Splitting up the file made them less silent, so these were fixed properly. 

I also opened an upstream fix PR which when merged will allow us to get rid of some of the lookup stuff entirely
https://github.com/conda-forge/skopeo-feedstock/pull/62


bgruening:
Thanks a lot @paperbenni

